PR #90, opened 2018-05-09
Fixed auth and tests

### LINKED ISSUE #89: Anonymous users show as Authenticated
When switching ultimate tic tac toe to use the service, it ends up treating users that aren't logged in as authenticated users.

davidfowl:
- Made it so tests don't depend on timeouts
- Fixed the auth bug and added unit tests

Fixes #89 

PS: I branched off the tag, but we should probably get a fix out to nuget ASAP once this is fixed.

### DIFF
--- a/src/Microsoft.Azure.SignalR/HubHost/IClientConnectionManager.cs
+++ b/src/Microsoft.Azure.SignalR/HubHost/IClientConnectionManager.cs
@@ -9,6 +9,8 @@ namespace Microsoft.Azure.SignalR
     {
         void AddClientConnection(ServiceConnectionContext clientConnection);
 
+        void RemoveClientConnection(string connectionId);
+
         ConcurrentDictionary<string, ServiceConnectionContext> ClientConnections { get; }
     }
 }
--- a/test/Microsoft.Azure.SignalR.Tests/Infrastructure/ServiceConnectionProxy.cs
+++ b/test/Microsoft.Azure.SignalR.Tests/Infrastructure/ServiceConnectionProxy.cs
@@ -4,56 +4,57 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.IO;
+using System.IO.Pipelines;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Connections;
 using Microsoft.AspNetCore.SignalR;
 using Microsoft.AspNetCore.SignalR.Protocol;
 using Microsoft.Azure.SignalR.Protocol;
 using Microsoft.Extensions.Logging.Abstractions;
+using HandshakeRequestMessage = Microsoft.Azure.SignalR.Protocol.HandshakeRequestMessage;
+using HandshakeResponseMessage = Microsoft.Azure.SignalR.Protocol.HandshakeResponseMessage;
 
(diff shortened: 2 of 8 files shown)


test/Microsoft.Azure.SignalR.Tests/Infrastructure/ServiceConnectionProxy.cs
```diff
@@ -4,127 +4,169 @@
...
+        public ConcurrentDictionary<string, ServiceConnectionContext> ClientConnections => ClientConnectionManager.ClientConnections;
+
+        private readonly ConcurrentDictionary<string, TaskCompletionSource<ConnectionContext>> _waitForConnectionOpen = new ConcurrentDictionary<string, TaskCompletionSource<ConnectionContext>>();
+        private readonly ConcurrentDictionary<string, TaskCompletionSource<object>> _waitForConnectionClose = new ConcurrentDictionary<string, TaskCompletionSource<object>>();
+
+        public ServiceConnectionProxy()
         {
             ConnectionContext = new TestConnection();
             ConnectionFactory = new TestConnectionFactory(ConnectionContext);
             ClientConnectionManager = new ClientConnectionManager();
```
xscript: This can be removed?

src/Microsoft.Azure.SignalR/HubHost/IClientConnectionManager.cs
```diff
@@ -9,6 +9,8 @@ internal interface IClientConnectionManager
     {
         void AddClientConnection(ServiceConnectionContext clientConnection);
 
+        void RemoveClientConnection(string connectionId);
+
         ConcurrentDictionary<string, ServiceConnectionContext> ClientConnections { get; }
```
JamesNK: Is ClientConnections still used externally?

test/Microsoft.Azure.SignalR.Tests/Infrastructure/ServiceConnectionProxy.cs
```diff
@@ -4,127 +4,169 @@
...
         }
 
-        public void Start()
+        public Task StartAsync()
         {
             _ = ServiceConnection.StartAsync();
+            return HandshakeAsync();
         }
 
         public void Stop()
```
JamesNK: Not saying it should change but it is kind of weird that Start is now async and Stop isn't.

test/Microsoft.Azure.SignalR.Tests/Infrastructure/ServiceConnectionProxy.cs
```diff
@@ -4,127 +4,169 @@
...
+            connection.Transport.Input.OnWriterCompleted((ex, state) =>
             {
-                while (true)
-                {
-                    var result = await connection.Transport.Input.ReadAsync();
+                tcs.TrySetResult(null);
+            },
+            null);
+
+            return tcs.Task; ;
```
ccic: nit: double ";"

src/Microsoft.Azure.SignalR/HubHost/IClientConnectionManager.cs
davidfowl: Yes, it's part of the interface and is used to dispatch messages to specific connections.

PR #93, opened 2018-05-09
Update external files and remove unused

JamesNK:
MemoryBufferWriter has been updated with latest changes from SignalR.
Unused files have been removed.


src/Common/MemoryBufferWriter.cs
```diff
@@ -285,7 +293,7 @@ public override void Write(byte[] buffer, int offset, int count)
             }
         }
 
-#if NETCOREAPP2_1
+#if NETCOREAPP2_2
```
JamesNK: Azure SignalR SDK is netstandard2.0 only so this changing doesn't effect you.

PR #94, opened 2018-05-10
Make AuthenticationHelper internal

### LINKED ISSUE #91: Should AuthenticationHelper be public?
https://github.com/Azure/azure-signalr/blob/dev/src/Microsoft.Azure.SignalR/Utilities/AuthenticationHelper.cs

Should this be public?

Either make internal or add documentation (and consider moving out of the Utilities directory).

davidfowl:
Fixes #91


src/Microsoft.Azure.SignalR/Utilities/AuthenticationHelper.cs
```diff
@@ -10,7 +10,7 @@
 
 namespace Microsoft.Azure.SignalR
 {
-    public class AuthenticationHelper
+    internal class AuthenticationHelper
```
JamesNK: While you're here, make it static too 😸

Issue #99, opened 2018-05-13
Visual Studio build error -  current .NET SDK does not support targeting .NET Core 2.1


felipecruz91:
Hi there!

I'd like to start contributing to this project so I have followed the [Developer Getting Started](https://github.com/Azure/azure-signalr#developer-getting-started) section.

The project seems to have been built correctly via the `build.cmd` command (click [here ](https://gist.github.com/felipecruz91/366407c7f64b7c415946b16538c64525) to see the output for more information.)

However, when building the `AzureSignalR` solution in Visual Studio version 15.7.1, I get the following error:

`The current .NET SDK does not support targeting .NET Core 2.1.  Either target .NET Core 2.0 or lower, or use a version of the .NET SDK that supports .NET Core 2.1.`

The current dotnet SDK version installed in my computer is 2.1.200.

Thank you!

xscript:
Thanks @felipecruz91 for reporting the issue.
`build.cmd` will install the required SDK for you. You need to add the path of that SDK to the `PATH` environment variable so that Visual Studio can use it.
By default, it is installed at `c:\Users\<username>\.dotnet\x64`

davidfowl:
We should be able to switch to the official 2.1 builds now.

felipecruz91:
Thanks @xscript - After having added the env. variable to the `PATH`, the project now builds successfully.

However, when launching the `ChatSample` project locally, the following runtime exception is thrown:

`An unhandled exception of type 'System.IO.FileLoadException' occurred in System.Private.CoreLib.dll
Could not load file or assembly 'System.Memory, Version=4.1.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)`

PR #100, opened 2018-05-15
fix issue #99

ccic:
Korebuild's config for dotnet SDK needs to be updated. Otherwise, it reports exception:
"Could not load file or assembly 'System.Memory, Version=4.1.0.0, Culture=neutral,
PublicKeyToken=cc7b13ffcd2ddd51'"

dotnet 2.1.300-rc1-008673 works fine. But lower version has this issue.

### DIFF
--- a/korebuild-lock.txt
+++ b/korebuild-lock.txt
@@ -1,2 +1,2 @@
-version:2.1.0-rc1-15775
-commithash:f967cebaec4c8f46d7cb3434bc9bc1a4a59b4095
+version:2.1.0-rtm-15783
+commithash:5fc2b2f607f542a2ffde11c19825e786fc1a3774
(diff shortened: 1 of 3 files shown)


korebuild-lock.txt
```diff
@@ -1,2 +1,2 @@
-version:2.1.0-rc1-15775
-commithash:f967cebaec4c8f46d7cb3434bc9bc1a4a59b4095
+version:2.1.0-rtm-15783
```
xscript: We are still supposed to use RC1?
Add @dviadfowl to confirm on this.

davidfowl:
I think we need to drop KoreBuild eventually. There's no plan AFAIK to lock to an RTM version of the SDK once it's released.

ccic:
Let us first fix the blocking issue. In next step replace KoreBuild with script which is responsible to download/install dotnet core SDK, build, pack, and run test.

PR #102, opened 2018-05-21
send PingMessage to service to keep alive

xscript:
Send `PingMessage` from server to service every 5 seconds, so that service can quickly detect an unresponsive service connection.

### DIFF
--- a/src/Microsoft.Azure.SignalR/HubHost/ServiceConnection.cs
+++ b/src/Microsoft.Azure.SignalR/HubHost/ServiceConnection.cs
@@ -249,14 +249,14 @@ namespace Microsoft.Azure.SignalR
             }
         }
 
-        private Timer StartKeepAliveTimer()
+        private TimerAwaitable StartKeepAliveTimer()
         {
             Log.StartingKeepAliveTimer(_logger, DefaultKeepAliveInterval);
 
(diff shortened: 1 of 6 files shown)


src/Microsoft.Azure.SignalR/HubHost/ServiceConnection.cs
```diff
@@ -241,20 +249,39 @@ internal Task StopAsync()
             }
         }
 
-        private Timer StartTimeoutTimer()
+        private Timer StartKeepAliveTimer()
```
davidfowl: Since we're moving to a single timer model here I would change this logic to use the TimerAwaitable.

xscript: Sure. Let me update it.

src/Microsoft.Azure.SignalR/HubHost/TimerAwaitable.cs
```diff
@@ -1,109 +1,109 @@
-// Copyright (c) .NET Foundation. All rights reserved.
```
davidfowl: What changed here? Just the license?

xscript: Just the license. No idea why the whole file is marked as changed.

PR #105, opened 2018-05-23
Fix MultiUserDataMessage type

### LINKED ISSUE #104: MultiUserDataMessage has incorrect message type, causing exception when ServiceProtocol parsing it
https://github.com/Azure/azure-signalr/blob/dev/src/Microsoft.Azure.SignalR.Protocols/ServiceProtocol.cs#L275

xscript:
- Fix #104 
- Add test cases for `ServiceProtocol`

### DIFF
--- a/test/Microsoft.Azure.SignalR.Protocols.Tests/ServiceProtocolFacts.cs
+++ b/test/Microsoft.Azure.SignalR.Protocols.Tests/ServiceProtocolFacts.cs
@@ -190,8 +190,9 @@ Please verify the MsgPack output and update the baseline");
 
             // Parse the input fully now.
             bytes = Frame(bytes);
-            var message = ParseServiceMessage<OpenConnectionMessage>(bytes);
-            Assert.Equal(expectedMessage, message, ServiceMessageEqualityComparer.Instance);
+            var message = ParseServiceMessage(bytes);
+            var openConnectionMessage = Assert.IsType<OpenConnectionMessage>(message);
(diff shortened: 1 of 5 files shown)


test/Microsoft.Azure.SignalR.Protocols.Tests/ServiceProtocolFacts.cs
```diff
@@ -68,5 +223,35 @@ private static byte[] Frame(byte[] input)
                 MemoryBufferWriter.Return(stream);
             }
         }
+
+        private static ServiceMessage ParseServiceMessage(byte[] bytes)
```
davidfowl: Why do we need this method?

xscript: Used [here](https://github.com/Azure/azure-signalr/pull/105/files#diff-6dee76e558c2ed86e6135a0ebd3ece52R157) because we don;t have the specific type.
Making it a method to look nicer.

davidfowl: Can't you call the generic one with ServiceMessage as the type?

xscript: Yeah, i tried but failed with error. Looks like `IsType` will check the exact type.

PR #106, opened 2018-05-25
Use condition PackageReference in MicroBuild

zackliu:


src/Directory.Build.props
```diff
@@ -6,5 +6,10 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <PackageTags>aspnetcore;signalr</PackageTags>
     <EnableApiCheck>false</EnableApiCheck>
+    <EnableSigning>false</EnableSigning>
   </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="MicroBuild.Core" Version="$(MicroBuildCorePackageVersion)" Condition="'$(EnableSigning)' == 'true'" />
```
xscript: So we pass the `EnableSigning` flag from build definition?

zackliu: Yes

PR #107, opened 2018-05-25
filter claims in OpenConnectionMessage for unauthenticated user

xscript:


.gitignore
```diff
@@ -287,4 +287,7 @@ __pycache__/
 *.odx.cs
 *.xsd.cs
 
-.publish/
\ No newline at end of file
+.publish/
+
+# VSCode
+.vscode
```
davidfowl: This should actually be checked in

xscript: I didn't actually use VSCode to debug. So not sure my configuration is valid. Let me simply revert it.

PR #110, opened 2018-05-26
add JWT authentication in ChatSample

xscript:


samples/ChatSample/Controllers/TokenController.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Text;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.IdentityModel.Tokens;
+
+namespace ChatSample
+{
+    [Route("token")]
+    public class TokenController : Controller
+    {
+        private const string SigningKey = "123456789012345678901234567890";
```
chenkennt: Move this to a configuration entry.

xscript: Right!

PR #114, opened 2018-05-30
update SignalR dependencies to 1.0.0

xscript:


samples/ChatSample/package.json
```diff
@@ -2,7 +2,7 @@
     "name": "chatsample",
     "private": true,
     "dependencies": {
-        "@aspnet/signalr": "^1.0.0-rc1-final",
+        "@aspnet/signalr": "^1.0.0",
         "@aspnet/signalr-protocol-msgpack": "^1.0.0-rc1-final"
```
davidfowl: Update this

xscript: Why is thing happening... 😕 
I will fix it.

PR #120, opened 2018-06-07
Use cdn version scripts for chat sample

vicancy:


davidfowl:
While I like this, it means the sample won't work offline.

vicancy:
@davidfowl that's true... if offline debugging is a requirement...

PR #129, opened 2018-06-19
load connection string from ConnectionStrings section if default key not exist

### LINKED ISSUE #126: Consider adding `ConnectionStrings` in the fallback path when loading default connection string from environment
Azure Web App has a "Connection Strings" section specifically for secret connection strings. We should consider adding it into the fallback path when loading default connection string from environment, to optimize the experiences in Azure Web App.

xscript:
- fix #126

### DIFF
--- a/src/Microsoft.Azure.SignalR/HubHost/ServiceOptionsSetup.cs
+++ b/src/Microsoft.Azure.SignalR/HubHost/ServiceOptionsSetup.cs
@@ -8,11 +8,22 @@ namespace Microsoft.Azure.SignalR
 {
     internal class ServiceOptionsSetup : IConfigureOptions<ServiceOptions>
     {
+        private static readonly string ConnectionStringSecondaryKey =
+            $"ConnectionStrings:{ServiceOptions.ConnectionStringDefaultKey}";
+
         private readonly string _connectionString;
 
         public ServiceOptionsSetup(IConfiguration configuration)
         {
-            _connectionString = configuration.GetSection(ServiceOptions.ConnectionStringDefaultKey).Value;
+            var connectionString = configuration.GetSection(ServiceOptions.ConnectionStringDefaultKey).Value;
+
+            // Load connection string from "ConnectionStrings" section when default key doesn't exist or holds an empty value.
+            if (string.IsNullOrEmpty(connectionString))
+            {
+                connectionString = configuration.GetSection(ConnectionStringSecondaryKey).Value;
+            }
+
+            _connectionString = connectionString;
         }
 
         public void Configure(ServiceOptions options)
--- a/test/Microsoft.Azure.SignalR.Tests/AddAzureSignalRFacts.cs
+++ b/test/Microsoft.Azure.SignalR.Tests/AddAzureSignalRFacts.cs
@@ -21,16 +21,16 @@ namespace Microsoft.Azure.SignalR.Tests
         {
             var services = new ServiceCollection();
             var config = new ConfigurationBuilder()
-                            .AddInMemoryCollection(new Dictionary<string, string>
-                            {
-                                {"Azure:SignalR:ConnectionString", DefaultValue}
-                            })
-                            .Build();
+                .AddInMemoryCollection(new Dictionary<string, string>
+                {
+                    {"Azure:SignalR:ConnectionString", DefaultValue}
+                })
+                .Build();
             var serviceProvider = services.AddSignalR()
-                                          .AddAzureSignalR()
-                                          .Services
-                                          .AddSingleton<IConfiguration>(config)
-                                          .BuildServiceProvider();
+                .AddAzureSignalR()
+                .Services
+                .AddSingleton<IConfiguration>(config)
+                .BuildServiceProvider();
 
             var options = serviceProvider.GetRequiredService<IOptions<ServiceOptions>>().Value;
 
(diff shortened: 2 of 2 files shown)


src/Microsoft.Azure.SignalR/HubHost/ServiceOptionsSetup.cs
```diff
@@ -8,11 +8,22 @@ namespace Microsoft.Azure.SignalR
 {
     internal class ServiceOptionsSetup : IConfigureOptions<ServiceOptions>
     {
+        private static readonly string ConnectionStringSecondaryKey =
```
davidfowl: Did you test it on azure web apps?

test/Microsoft.Azure.SignalR.Tests/AddAzureSignalRFacts.cs
```diff
@@ -85,8 +89,41 @@ public void AddAzureReadsConnectionStringFirst()
...
+        {
+            var services = new ServiceCollection();
+            var config = new ConfigurationBuilder()
+                            .AddInMemoryCollection(new Dictionary<string, string>
+                            {
+                                {"Azure:SignalR:ConnectionString", defaultValue},
+                                {"ConnectionStrings:Azure:SignalR:ConnectionString", secondaryValue}
+                            })
+                            .Build();
+            var serviceProvider = services.AddSignalR()
```
davidfowl: nit: This formatting looks a bit indented, can we reduce it a little

test/Microsoft.Azure.SignalR.Tests/AddAzureSignalRFacts.cs
```diff
@@ -85,8 +89,41 @@ public void AddAzureReadsConnectionStringFirst()
...
+        [InlineData(null, null, SecondaryValue, SecondaryValue)]
+        public void AddAzureSignalRLoadConnectionStringOrder(string customValue, string defaultValue,
+            string secondaryValue, string expected)
+        {
+            var services = new ServiceCollection();
+            var config = new ConfigurationBuilder()
+                            .AddInMemoryCollection(new Dictionary<string, string>
+                            {
+                                {"Azure:SignalR:ConnectionString", defaultValue},
+                                {"ConnectionStrings:Azure:SignalR:ConnectionString", secondaryValue}
```
davidfowl: This looks so strange 😄. `ConnectionStrings:Azure:SignalR:ConnectionString`

Should it be `ConnectionStrings:Azure:SignalR` instead?

chenkennt: Are ConnectionStrings section mainly for SQL databases? When I create a connection string in Azure portal, I need to choose from MySQL, SQLServer, SQLAzure, and Custom. What's the difference? Should user choose Custom when using SignalR?

davidfowl: Usually it's a database connection string. For example, I don't think you specify your service bus connection string in there.

src/Microsoft.Azure.SignalR/HubHost/ServiceOptionsSetup.cs
xscript: Yeah, tested. Works well.

test/Microsoft.Azure.SignalR.Tests/AddAzureSignalRFacts.cs
xscript: Sure.

test/Microsoft.Azure.SignalR.Tests/AddAzureSignalRFacts.cs
xscript: Agree that the key is repetitive and doesn't sound natural.
However this fix is specifically for Azure Web Apps. When user "accidentally" put the connection string in the `Connection Strings` section, we can still pick it up. So if we change the key name, it means we provide another key and need to document it.
IMO, this is not a recommended approach to set the connection string in environment. But if user prefer to put it in `Connection Strings` section, the Service SDK can still work.

test/Microsoft.Azure.SignalR.Tests/AddAzureSignalRFacts.cs
```diff
@@ -68,25 +69,55 @@ public void AddAzureReadsConnectionStringFirst()
...
+        [InlineData(null, null, SecondaryValue, SecondaryValue)]
+        public void AddAzureSignalRLoadConnectionStringOrder(string customValue, string defaultValue,
+            string secondaryValue, string expected)
+        {
+            var services = new ServiceCollection();
+            var config = new ConfigurationBuilder()
+                .AddInMemoryCollection(new Dictionary<string, string>
+                {
+                    {"Azure:SignalR:ConnectionString", defaultValue},
+                    {"ConnectionStrings:Azure:SignalR:ConnectionString", secondaryValue}
```
davidfowl: This key isn't correct right?

src/Microsoft.Azure.SignalR/HubHost/ServiceOptionsSetup.cs
davidfowl: We should change this to exclude the trailing ConnectionString right?

Issue #108, opened 2018-05-25
Unable to pass information from the client to the hub


yigitgursoy:
Context.GetHttpContext() returns the HttpContext in SignalR mode, but returns null in AzureSignalR mode. Therefore I can't read any QueryString or custom Http Header values in the OnConnectedAsync() method of the Hub. I'm using the following code to create the connection.

```csharp
var connection = new HubConnectionBuilder().WithUrl("http://localhost:5000/message?t=TestValue",
                    options => options.Headers = new Dictionary<string, string>
                    {
                        {
                            "X-CustomHeader", "CustomValue"
                        }
                    })
                .Build();
await connection.StartAsync();
```

Is there any way to pass information from the client to the hub using the AzureSignalR service?

xscript:
Yes, the headers are not passed through from client to server. We will need some discussion to decide whether to support it.
A work around is to create a separate hub method and call this method to send information once the client is connected.

xscript:
The fix is ready and on the way to be deployed to production.
A workaround is you can add the room id in a custom claim. We have an option called [ClaimsProvider](https://github.com/Azure/azure-signalr/blob/dev/src/Microsoft.Azure.SignalR/ServiceOptions.cs#L38). The default behavior is we will get `HttpContext.User.Claims` to generate the access token, which will be used to authenticate with Azure SignalR Service.
You can customize the claims by adding your custom claim like below. And in your hub, you can get the custom claim from `Hub.Context.User`.
```csharp
services.AddSignalR()
        .AddAzureSignalR(options =>
        {
            options.ClaimsProvider = context =>
            {
                var claims = context.User.Claims.ToList();
                claims.Add(new Claim("<your-namespace>.room.id", "1"));
                return claims;
            };
        });
```

maxiptah:
Do you have an estimation as to when this fix is going to be deployed?
I'm trying to pass an auth0 token to the hub method. This is normally passed as a bearer token in authorization header. However as stated in this [post](https://stackoverflow.com/questions/15528221/passing-token-through-http-headers-signalr), custom header is not possible because of limitations of JS websocket client which does not allow for it. So I wanted to pass my token as query parameter, but it never reaches my hub. 
Where should I read the query parameters anyway? From Context.Items?

xscript:
I can't give you a concrete time when the deployment is finished. Deployment is not a one-day job. We have to update multiple regions and also thousands of existing instances.
[...]

PR #131, opened 2018-06-21
Add headers and query string in OpenConnectionMessage

xscript:
- SDK change for #108


test/Microsoft.Azure.SignalR.Protocols.Tests/ServiceMessageEqualityComparer.cs
```diff
@@ -176,6 +180,30 @@ private static bool ClaimsEqual(Claim[] x, Claim[] y)
             return true;
         }
 
+        private static bool HeadersEqual(HeaderDictionary x, HeaderDictionary y)
+        {
+            if (x == null && y == null)
```
vicancy: check when `x` or `y` is `null`

xscript: fixed.

build/dependencies.props
```diff
@@ -6,6 +6,7 @@
     <!-- Azure SignalR -->
     <MessagePackPackageVersion>1.7.3.4</MessagePackPackageVersion>
     <MicrosoftAspNetCoreHttpConnectionsClientPackageVersion>1.0.0</MicrosoftAspNetCoreHttpConnectionsClientPackageVersion>
+    <MicrosoftAspNetCoreHttpPackageVersion>2.1.0</MicrosoftAspNetCoreHttpPackageVersion>
```
davidfowl: I think we should consider dropping this dependency and just use a dictionary instead.

src/Microsoft.Azure.SignalR.Protocols/ConnectionMessage.cs
```diff
@@ -33,15 +34,40 @@ public class OpenConnectionMessage : ConnectionMessage
...
 
         /// <summary>
         /// Gets or sets the associated claims.
         /// </summary>
         public Claim[] Claims { get; set; }
+
+        /// <summary>
+        /// Gets or sets the associated headers.
+        /// </summary>
+        public HeaderDictionary Headers { get; set; }
```
davidfowl: This should be `Dictionary<string, string>`

build/dependencies.props
xscript: Yeah, we can do this.

test/Microsoft.Azure.SignalR.Protocols.Tests/ServiceMessageEqualityComparer.cs
chenkennt: if `x == null && y == []` the code below will throw

src/Microsoft.Azure.SignalR.Protocols/ServiceProtocol.cs
```diff
@@ -401,12 +423,23 @@ private static HandshakeResponseMessage CreateHandshakeResponseMessage(byte[] in
...
 
-        private static OpenConnectionMessage CreateOpenConnectionMessage(byte[] input, ref int offset)
+        private static OpenConnectionMessage CreateOpenConnectionMessage(int arrayLength, byte[] input, ref int offset)
         {
             var connectionId = ReadString(input, ref offset, "connectionId");
             var claims = ReadClaims(input, ref offset);
 
-            return new OpenConnectionMessage(connectionId, claims);
+            // Backward compatible with old versions
+            if (arrayLength > 3)
```
chenkennt: This is still not compatible if service SDK is old version and runtime is new version?

test/Microsoft.Azure.SignalR.Protocols.Tests/ServiceMessageEqualityComparer.cs
xscript: Fixed. `x?.Count != y?.Count` won't throw.

src/Microsoft.Azure.SignalR.Protocols/ServiceProtocol.cs
xscript: Old SDK will only read the first 3 fields. Extra data will be ignored.

src/Microsoft.Azure.SignalR.Protocols/ConnectionMessage.cs
```diff
@@ -33,15 +34,40 @@ public class OpenConnectionMessage : ConnectionMessage
...
 
         /// <summary>
         /// Gets or sets the associated claims.
         /// </summary>
         public Claim[] Claims { get; set; }
+
+        /// <summary>
+        /// Gets or sets the associated headers.
+        /// </summary>
+        public IHeaderDictionary Headers { get; set; }
```
davidfowl: OK giving it some more thought, I think this should be `IDictionary<string, StringValues>` and we should change this dependency `Microsoft.AspNetCore.Http` to `Microsoft.Extensions.Primitives`

xscript: Any reason other than minimum dependency?

davidfowl: Nope It’s about minimizing dependencies

src/Microsoft.Azure.SignalR.Protocols/ServiceProtocol.cs
```diff
@@ -529,6 +566,30 @@ private static Claim[] ReadClaims(byte[] input, ref int offset)
             return null;
         }
 
+        private static Dictionary<string, StringValues> ReadHeaders(byte[] input, ref int offset)
+        {
+            var headerCount = ReadMapLength(input, ref offset, "headers");
+            if (headerCount > 0)
+            {
+                var headers = new Dictionary<string, StringValues>((int)headerCount);
```
davidfowl: Shouldn't this be case insensitive? Do we have tests for that case?

xscript: Yes, it should be case insensitive.
I have a question though. Since we just forward the headers from `HttpContext`, it really doesn't breka anything here, right? I assume headers with the same key have been aggregated by ASP.NET Core?

xscript: Yes, it should be case insensitive.
I have a question though. Since we just forward the headers from `HttpContext`, it really doesn't break anything here, right? I assume headers with the same key have been aggregated by ASP.NET Core?

src/Microsoft.Azure.SignalR/HubHost/ServiceConnectionContext.cs
```diff
@@ -128,5 +130,33 @@ private static bool IsAuthenticatedUser(Claim[] claims)
...
+            features.Set<IHttpContextFeature>(this);
+            return features;
+        }
+
+        private HttpContext BuildHttpContext(OpenConnectionMessage message)
+        {
+            var httpContextFeatures = new FeatureCollection();
+            httpContextFeatures.Set<IHttpRequestFeature>(new HttpRequestFeature
+            {
+                Headers = new HeaderDictionary((Dictionary<string, StringValues>) message.Headers),
```
davidfowl: Why do you have to cast here?

src/Microsoft.Azure.SignalR.Protocols/ServiceProtocol.cs
davidfowl: It does, this code is in control of the creation of the `HttpContext`, there is no ASP.NET Core Server involved in the pipeline here. This basically replaces the server. So no, this logic is responsible for header aggregation.

src/Microsoft.Azure.SignalR/HubHost/ServiceConnectionContext.cs
xscript: Because `HeaderDictionary` only accepts a `Dictionary<string, StringValues>`.
https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.headerdictionary?view=aspnetcore-2.0#constructors

src/Microsoft.Azure.SignalR.Protocols/ServiceProtocol.cs
xscript: Are you talking about when service protocol is used out of Azure SignalR Service?
My understanding is when user's request is received by Azure SignalR Service runtime, ASP.NET Core has already handled the headers? Then headers are passed from SignalR Service to Server through `OpenConnectionMessage`. Am I missing anything?

davidfowl: When using the SDK the http request is shuttled over the websocket connection via the service protocol. This logic needs to create a case insensitive dictionary.

PR #136, opened 2018-07-02
Throw more friendly exception when calling AddAzureSignalR without UseAzureSignalR

### LINKED ISSUE #121: More friendly error message when `UseAzureSignalR` is not called with `AddAzureSignalR`
More context is at #119 .
After calling `AddAzureSignalR`, if user calls `UseSignalR` instead of `UseAzureSignalR`, we should throw exception with better error message to tell user what is wrong.

zackliu:
Fix #121


src/Microsoft.Azure.SignalR/ApplicationBuilderExtensions.cs
```diff
@@ -20,6 +20,8 @@ public static class AzureSignalRApplicationBuilderExtensions
         /// <returns>The same instance of the <see cref="IApplicationBuilder"/> for chaining.</returns>
         public static IApplicationBuilder UseAzureSignalR(this IApplicationBuilder app, Action<ServiceRouteBuilder> configure)
         {
+            var marker = app.ApplicationServices.GetService(typeof(AzureSignalRMarkerService));
```
xscript: Use `GetService<>` then you don't have to cast.

src/Microsoft.Azure.SignalR/Internal/AzureSignalRMarkerService.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.SignalR
+{
+    internal class AzureSignalRMarkerService
+    {
+        internal bool UseAzureSignalRFlag { set; get; }
```
vicancy: naming: `UseAzureSignalR` is enough?

src/Microsoft.Azure.SignalR/DependencyInjectionExtensions.cs
```diff
@@ -63,6 +63,7 @@ private static ISignalRServerBuilder AddAzureSignalRCore(this ISignalRServerBuil
                 .AddSingleton(typeof(IServiceConnectionManager<>), typeof(ServiceConnectionManager<>))
                 .AddSingleton(typeof(IServiceEndpointUtility), typeof(ServiceEndpointUtility))
                 .AddSingleton(typeof(ServiceHubDispatcher<>))
+                .AddSingleton(typeof(AzureSignalRMarkerService))
```
chenkennt: What will happen if `AddAzureSignalR()` is not called?

zackliu: In the present version, will get exception of 'No service for type xxx has been registered.' 
If add a marker check, will get exception like "Please add all the required services by calling           'IServiceCollection.AddAzureSignalR'"

src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManager.cs
```diff
@@ -23,8 +23,14 @@ internal class ServiceLifetimeManager<THub> : HubLifetimeManager<THub> where THu
 
         public ServiceLifetimeManager(IServiceConnectionManager<THub> serviceConnectionManager,
             IClientConnectionManager clientConnectionManager,
-            IHubProtocolResolver protocolResolver, ILogger<ServiceLifetimeManager<THub>> logger)
+            IHubProtocolResolver protocolResolver, ILogger<ServiceLifetimeManager<THub>> logger,
+            AzureSignalRMarkerService azureSignalRMarkerService)
```
xscript: `marker` would be clear enough for the parameter name.

src/Microsoft.Azure.SignalR/DependencyInjectionExtensions.cs
davidfowl: Actually as written it'll null ref since you're calling `GetService` instead of `GetRequiredService`

src/Microsoft.Azure.SignalR/Internal/AzureSignalRMarkerService.cs
zackliu: Or `IsUseAzureSignalR`? Do we have some naming convention of adding `Is` prefix in bool property

xscript: How about `IsConfigured` ?

PR #139, opened 2018-07-06
Fix wrong options in dotnet nuget push

zackliu:


zackliu:
Tested a lot and find a strange behavior. `dotnet nuget push` seems can only support the first option on Travis.
When I use `dotnet nuget push ./artifacts/build/*.nupkg -s $NUGET_SOURCE -k $API_KEY`, I will get `error: Missing value for option 'api-key'` 
And if I use `dotnet nuget push ./artifacts/build/*.nupkg -k $API_KEY -s $NUGET_SOURCE`, I will get `error: Source parameter was not specified.`
Don't know why, please have a look

Issue #141, opened 2018-07-17
Validate endpoint from connection string


xscript:
From https://github.com/MicrosoftDocs/azure-docs/issues/11680 .
It is possible that the connection string is messed up when it is copied around. SDK can validate the endpoint to be a valid URI and display error message to help troubleshoot.

xscript:
Fixed by #142

PR #142, opened 2018-07-17
Add Endpoint validation

zackliu:
#141

### DIFF
--- a/src/Microsoft.Azure.SignalR/EndpointUtility/ServiceEndpointUtility.cs
+++ b/src/Microsoft.Azure.SignalR/EndpointUtility/ServiceEndpointUtility.cs
@@ -147,7 +147,7 @@ namespace Microsoft.Azure.SignalR
                 {
                     if (!ValidateEndpoint(dict[EndpointProperty]))
                     {
-                        throw new ArgumentException($"Endpoint is not valid: {dict[EndpointProperty]}");
+                        throw new ArgumentException($"Endpoint is invalid: {dict[EndpointProperty]}");
                     }
 
                     return (dict[EndpointProperty].TrimEnd('/'), dict[AccessKeyProperty]);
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.SignalR/EndpointUtility/ServiceEndpointUtility.cs
```diff
@@ -145,11 +145,22 @@ internal static (string, string) ParseConnectionString(string connectionString)
 
                 if (dict.ContainsKey(EndpointProperty) && dict.ContainsKey(AccessKeyProperty))
                 {
+                    if (!ValidateEndpoint(dict[EndpointProperty]))
+                    {
+                        throw new ArgumentException($"Endpoint is invalid: {dict[EndpointProperty]}");
```
xscript: -> `Endpoint property in connection string is not a valid URI: {dict[EndpointProperty]}`.

src/Microsoft.Azure.SignalR/EndpointUtility/ServiceEndpointUtility.cs
```diff
@@ -145,11 +145,22 @@ internal static (string, string) ParseConnectionString(string connectionString)
...
             throw new ArgumentException(MissingRequiredProperty);
         }
+
+        internal static bool ValidateEndpoint(string endpoint)
+        {
+            return Uri.TryCreate(endpoint, UriKind.Absolute, out var uriResult) &&
+                   (uriResult.Scheme == Uri.UriSchemeHttp || uriResult.Scheme == Uri.UriSchemeHttps);
+        }
     }
 }
```
xscript: Let's add a new line here.

test/Microsoft.Azure.SignalR.Tests/ConnectionStringParsingFacts.cs
```diff
@@ -9,13 +9,13 @@ namespace Microsoft.Azure.SignalR.Tests
     public class ConnectionStringParsingFacts
     {
         [Theory]
-        [InlineData("Endpoint=aaa;AccessKey=bbb;")]
-        [InlineData("ENDPOINT=aaa/;ACCESSKEY=bbb;")]
+        [InlineData("endpoint=https://aaa;AccessKey=bbb;")]
+        [InlineData("ENDPOINT=https://aaa/;ACCESSKEY=bbb;")]
```
xscript: Since we also support `http`, let's add a test case for it.

PR #145, opened 2018-07-20
Fulfill `RunAzureSignalRCore` with necessary DI injections 

vicancy:
Fulfill `RunAzureSignalRCore` with necessary DI injections, add necessary interfaces and classes, with implementation to-do


src/Microsoft.Azure.SignalR.AspNet/OwinExtensions.cs
```diff
@@ -30,7 +39,62 @@ public static void RunAzureSignalR(this IAppBuilder builder)
...
+            else
+            {
+                // TODO: log something
+            }
+        }
+
+        private static void RegisterServiceObjects(HubConfiguration configuration)
+        {
+            // share the same object all through
+            var serviceOptions = new OptionsWrapper<ServiceOptions>(new ServiceOptions
```
davidfowl: Options.Create(new ServiceOptions(...))

davidfowl: Wait, do you even need the Options at all?

vicancy: I need an interface wrapper for `ServiceOptions` to DI, as I don't want to create an interface for `ServiceOptions`, (because Ninject expects interfaces...) so I reused the `IOptions`.

Issue #87, opened 2018-05-08
Group add and removal should complete the task when the user is in the group


davidfowl:
Today we're not waiting for the connection to be added to the group before returning from the method. That's one of the SignalR contracts that isn't being respected right now by the service and sdk. It might require a protocol change to send ACKs back (like we do in redis).

davidfowl:
@xscript @chenkennt we need to make sure this is the case before GA. No acks actually need to come back to the application. It can be done all on the service side itself.

xscript:
Yeah, this is the next issue I plan to fix. But my fix requires a new message type to send ACK back to application. Let's sync offline about your idea.

xscript:
Fixed in [the latest Service SDK](https://www.nuget.org/packages/Microsoft.Azure.SignalR/1.0.0-preview1-10200).

PR #147, opened 2018-07-25
send group message from a fixed service message

xscript:
Part of the fix to #87 

cc: @vwxyzh

### DIFF
--- a/src/Microsoft.Azure.SignalR/HubHost/ServiceConnectionManager.cs
+++ b/src/Microsoft.Azure.SignalR/HubHost/ServiceConnectionManager.cs
@@ -12,8 +12,6 @@ namespace Microsoft.Azure.SignalR
 {
     internal class ServiceConnectionManager<THub> : IServiceConnectionManager<THub> where THub : Hub
     {
-        private static readonly int DefaultHashCode = string.Empty.GetHashCode();
-
         private readonly List<ServiceConnection> _serviceConnections = new List<ServiceConnection>();
 
         public void AddServiceConnection(ServiceConnection serviceConnection)
@@ -37,8 +35,13 @@ namespace Microsoft.Azure.SignalR
         {
             if (_serviceConnections.Count == 0) return;
 
-            var hash = partitionKey?.GetHashCode() ?? DefaultHashCode;
-            var index = Math.Abs(hash % _serviceConnections.Count);
+            // If we hit this check, it is a code bug.
+            if (string.IsNullOrEmpty(partitionKey))
+            {
+                throw new ArgumentNullException(nameof(partitionKey));
+            }
+
+            var index = Math.Abs(partitionKey.GetHashCode() % _serviceConnections.Count);
             await _serviceConnections[index].WriteAsync(serviceMessage);
         }
     }
--- a/src/Microsoft.Azure.SignalR/HubHost/ServiceConnectionManager.cs
+++ b/src/Microsoft.Azure.SignalR/HubHost/ServiceConnectionManager.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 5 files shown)


src/Microsoft.Azure.SignalR/HubHost/ServiceConnectionManager.cs
```diff
@@ -29,5 +30,13 @@ public void AddServiceConnection(ServiceConnection serviceConnection)
             var index = StaticRandom.Next(_serviceConnections.Count);
             await _serviceConnections[index].WriteAsync(serviceMessage);
         }
+
+        public async Task WriteAsync(ServiceMessage serviceMessage, int index)
```
vicancy: >WriteAsync [](start = 26, length = 10)

how about also hiding the hashing details and expose a `WriteWithGroup`, so that the manager can decide what algorithm to use, e,g, `name.GetHashCode % 5`, or maybe other partitioning algorithms?

src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManager.cs
```diff
@@ -81,111 +90,126 @@ public override Task SendConnectionAsync(string connectionId, string methodName,
...
-            return true;
+            return string.IsNullOrEmpty(value);
         }
 
-        private bool IsEmptyString(string value, string name)
+        private static bool IsInvalidArgument(IReadOnlyList<object> list)
         {
-            if (!string.IsNullOrEmpty(value)) return false;
-            return true;
+            return list == null || !list.Any();
```
vicancy: >Any [](start = 41, length = 3)

use `.Count` for List

src/Microsoft.Azure.SignalR/HubHost/ServiceConnectionManager.cs
xscript: Make sense. Will be consistent with the original `WriteAsync`.
I plan to update it to `public async Task WriteAsync(string partitionKey, ServiceMessage serviceMessage)`.

vicancy: I am not sure. So the caller decides what is the `partitionKey`, is it too much for the caller?

xscript: Essentially the same as `WriteWithGroup`, but more general. `partitionKey` just means that the same key will go through the same service connection. Caller doesn't have to care the details of mapping partition keys.
In this case, when the same group name is passed in as partition key, the service message will be sent out via the same service connection.

src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManager.cs
xscript: Sure.

src/Microsoft.Azure.SignalR/HubHost/ServiceConnectionManager.cs
davidfowl: I thin partition key is fine.

src/Microsoft.Azure.SignalR/HubHost/ServiceConnectionManager.cs
```diff
@@ -29,5 +32,14 @@ public void AddServiceConnection(ServiceConnection serviceConnection)
             var index = StaticRandom.Next(_serviceConnections.Count);
             await _serviceConnections[index].WriteAsync(serviceMessage);
         }
+
+        public async Task WriteAsync(string partitionKey, ServiceMessage serviceMessage)
+        {
+            if (_serviceConnections.Count == 0) return;
+
+            var hash = partitionKey?.GetHashCode() ?? DefaultHashCode;
```
vicancy: if `partitionKey` is `null`, shall we call `WriteAsync(ServiceMessage)` instead? otherwise all the traffic with `null` `partitionKey` goes into one service, is that intentional? Or shall we prevent `null` case?

xscript: This is just defensive code. Right now the partition key won't be null.

xscript: Also I don't think we should fall back to the random `WriteAsync`. Cause once this method is called, the intention is to send  message from a fix service connection, right? So it will probably break functionality if a random service connection is used.

vicancy: Then throw instead?

xscript: Fixed.

PR #148, opened 2018-07-25
Add custom UserAgent for general telemetry tracking

JialinXin:
sample UA: Microsoft.Azure.SignalR/1.0.0-preview1-t000 (.NET Core 4.6.26426.02; Microsoft Windows 10.0.17134; X64)

### DIFF
--- a/src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
+++ b/src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
@@ -27,7 +27,7 @@ namespace Microsoft.Azure.SignalR
         private readonly IClientConnectionFactory _clientConnectionFactory;
         private readonly string _userId;
         private static readonly string Name = $"ServiceHubDispatcher<{typeof(THub).FullName}>";
-        private static Dictionary<string, string> customHeader = ProductInfo.ToHeader();
+        private static Dictionary<string, string> CustomHeader = ProductInfo.ToHeader();
 
         public ServiceHubDispatcher(IServiceProtocol serviceProtocol,
             IServiceConnectionManager<THub> serviceConnectionManager,
@@ -91,7 +91,7 @@ namespace Microsoft.Azure.SignalR
                 AccessTokenProvider = () => Task.FromResult(_serviceEndpointUtility.GenerateServerAccessToken<THub>(_userId)),
                 Transports = HttpTransportType.WebSockets,
                 SkipNegotiation = true,
-                Headers = customHeader
+                Headers = CustomHeader
             };
             var httpConnection = new HttpConnection(httpConnectionOptions, _loggerFactory);
             try
--- a/src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
+++ b/src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
@@ -27,7 +27,7 @@ namespace Microsoft.Azure.SignalR
         private readonly IClientConnectionFactory _clientConnectionFactory;
         private readonly string _userId;
         private static readonly string Name = $"ServiceHubDispatcher<{typeof(THub).FullName}>";
-        private static Dictionary<string, string> CustomHeader = ProductInfo.ToHeader();
+        private static Dictionary<string, string> CustomHeader = new Dictionary<string, string> { { "User-Agent", ProductInfo.Get() } };
 
         public ServiceHubDispatcher(IServiceProtocol serviceProtocol,
             IServiceConnectionManager<THub> serviceConnectionManager,
--- a/src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
+++ b/src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Connections;
@@ -26,6 +27,7 @@ namespace Microsoft.Azure.SignalR
         private readonly IClientConnectionFactory _clientConnectionFactory;
         private readonly string _userId;
         private static readonly string Name = $"ServiceHubDispatcher<{typeof(THub).FullName}>";
+        private static ProductInfo _productInfo;
 
         public ServiceHubDispatcher(IServiceProtocol serviceProtocol,
             IServiceConnectionManager<THub> serviceConnectionManager,
@@ -45,6 +47,7 @@ namespace Microsoft.Azure.SignalR
             _logger = loggerFactory.CreateLogger<ServiceHubDispatcher<THub>>();
             _clientConnectionFactory = clientConnectionFactory;
             _userId = GenerateServerName();
+            _productInfo = new ProductInfo();
         }
 
         private static string GenerateServerName()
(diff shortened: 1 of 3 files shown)


src/Microsoft.Azure.SignalR/Internal/ProductInfo.cs
```diff
@@ -0,0 +1,32 @@
...
+
+using System;
+using System.Reflection;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Azure.SignalR
+{
+    internal class ProductInfo
+    {
+        public string Extra { get; set; } = "";
```
xscript: Let 's remove it if it is not used.

src/Microsoft.Azure.SignalR/Internal/ProductInfo.cs
```diff
@@ -0,0 +1,32 @@
...
+            var processorArchitecture = RuntimeInformation.ProcessArchitecture.ToString().Trim();
+
+            var userAgent = $"{packageId}/{version.InformationalVersion} ({runtime}; {operatingSystem}; {processorArchitecture})";
+
+            if (!String.IsNullOrWhiteSpace(Extra))
+            {
+                userAgent += $" {Extra.Trim()}";
+            }
+
+            return userAgent;
```
xscript: We can cache it, right?

src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
```diff
@@ -83,12 +86,17 @@ private Uri GetServiceUrl(string connectionId)
 
         public async Task<ConnectionContext> ConnectAsync(TransferFormat transferFormat, string connectionId, CancellationToken cancellationToken = default)
         {
+            var customHeaders = new Dictionary<string, string>
```
xscript: Or we can cache the headers instead of creating it every time.

src/Microsoft.Azure.SignalR/Internal/ProductInfo.cs
JialinXin: OK.

src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
JialinXin: Agree. Will update.

src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
```diff
@@ -26,6 +27,7 @@ internal class ServiceHubDispatcher<THub> : IConnectionFactory where THub : Hub
         private readonly IClientConnectionFactory _clientConnectionFactory;
         private readonly string _userId;
         private static readonly string Name = $"ServiceHubDispatcher<{typeof(THub).FullName}>";
+        private static ProductInfo _productInfo;
```
vicancy: >static [](start = 16, length = 6)

instantiate it here instead of in the .ctor as it is a static object?

src/Microsoft.Azure.SignalR/Internal/ProductInfo.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        public string Extra { get; set; } = "";
+
+        public override string ToString()
+        {
+            var packageId = Assembly.GetExecutingAssembly().GetName().Name;
+            var version = Assembly.GetExecutingAssembly().GetCustomAttribute(typeof(AssemblyInformationalVersionAttribute)) as AssemblyInformationalVersionAttribute;
+            var runtime = RuntimeInformation.FrameworkDescription.Trim();
+            var operatingSystem = RuntimeInformation.OSDescription.Trim();
+            var processorArchitecture = RuntimeInformation.ProcessArchitecture.ToString().Trim();
```
vicancy: >RuntimeInformation [](start = 40, length = 18)

No need to create an object, a static method is enough?

JialinXin: Correct. Fix in latest version.

src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
```diff
@@ -26,6 +27,7 @@ internal class ServiceHubDispatcher<THub> : IConnectionFactory where THub : Hub
         private readonly IClientConnectionFactory _clientConnectionFactory;
         private readonly string _userId;
         private static readonly string Name = $"ServiceHubDispatcher<{typeof(THub).FullName}>";
+        private static Dictionary<string, string> CustomHeader = ProductInfo.ToHeader();
```
vicancy: >ToHeader [](start = 77, length = 8)

It is possible that we might have other items in the header, so I think it is more appropriate for `ProductInfo` to just return UA information instead of the whole header

JialinXin: The header inits when we new HttpConnectionOptions(), so should be safe to new header here.

src/Microsoft.Azure.SignalR/Internal/ProductInfo.cs
```diff
@@ -0,0 +1,24 @@
...
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Azure.SignalR
+{
+    internal static class ProductInfo
+    {
+        public static string Get()
```
davidfowl: nit: GetProductInfo

src/Microsoft.Azure.SignalR/Internal/ProductInfo.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Azure.SignalR
+{
+    internal static class ProductInfo
+    {
+        public static string GetProductInfo()
+        {
+            var packageId = typeof(ProductInfo).GetTypeInfo().Assembly.GetName().Name;
+            var version = typeof(ProductInfo).GetTypeInfo().Assembly.GetCustomAttribute(typeof(AssemblyInformationalVersionAttribute)) as AssemblyInformationalVersionAttribute;
```
vicancy: >typeof(ProductInfo).GetTypeInfo().Assembly [](start = 26, length = 42)

extract to a local var as it is used twice?

vicancy: >GetCustomAttribute [](start = 69, length = 18)

can use the generic version `GetCustomAttribute<AssemblyInformationalVersionAttribute>()`

src/Microsoft.Azure.SignalR/Internal/ProductInfo.cs
```diff
@@ -0,0 +1,24 @@
...
+            var runtime = RuntimeInformation.FrameworkDescription.Trim();
```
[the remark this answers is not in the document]
JialinXin: No. They shouldn't be null unless runtime not working at all.

src/Microsoft.Azure.SignalR/Internal/ProductInfo.cs
JialinXin: OK.

JialinXin: cool.

src/Microsoft.Azure.SignalR/Internal/ProductInfo.cs
```diff
@@ -0,0 +1,25 @@
...
+namespace Microsoft.Azure.SignalR
+{
+    internal static class ProductInfo
+    {
+        public static string GetProductInfo()
+        {
+            var assembly = typeof(ProductInfo).GetTypeInfo().Assembly;
+            var packageId = assembly.GetName().Name;
+            var version = assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>().InformationalVersion;
+            var runtime = RuntimeInformation.FrameworkDescription.Trim();
```
davidfowl: I would make sure these never return null

davidfowl: Or add ?.Trim()

Issue #101, opened 2018-05-19
ASP.NET SignalR Support


davidfowl:
This is an issue to track supporting ASP.NET SignalR as part of the SignalR Azure Service.

vicancy:
- [ ] 1. Service side
  - [ ] Reconnect doesn’t work, we need to figure out what to do here. Do we indeed reconnect or do you return a non 200 status code so the client doesn’t reconnect.
  - [ ] We need to do message framing on the way in and on the way out (this one needs a bit more explanation)
  - [ ] Negotiate settings are hard coded, how do we want to handle the fact that settings set by the app isn’t propagated to the client (might require a SignalR change)
  - [ ] Do we copy all of the transport code from SignalR itself into the SDK to make this work? If the changes are in the core transport layer in SignalR. How is the service going to get these changes?

- [ ] 2. SDK side

phil0007:
How is this issue tracking. What is the ETA for Azure SignalR support on .NET Framework?

PinkPanda17:
Is there any update?

vicancy:
We will have this feature public preview in Azure Service late this month. The announcement https://blogs.msdn.microsoft.com/webdev/2018/09/17/the-future-of-asp-net-signalr/ have some more details of the timeline.

PinkPanda17:
Thanks For this :)

vicancy:
2.4.0 preview for ASP.NET SignalR is released, which contains the support for Azure SignalR Service, please have a try.
https://blogs.msdn.microsoft.com/webdev/2018/10/04/announcing-asp-net-signalr-2-4-0-preview-1/

PR #152, opened 2018-08-03
[ASP.NET SignalR]Implement ServiceMessageBus and ServiceConnectionManager

vicancy:
Continue with #101 
Implement a `ServiceMessageBus` to redirect `Message`s to service side.


src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ServiceMessageBus.cs
```diff
@@ -2,24 +2,175 @@
...
+            if (TryGetName(message.Key, PrefixHelper.ConnectionIdPrefix, out var connectonId))
+            {
+                return _serviceConnectionManager.WriteAsync(new ConnectionDataMessage(connectonId, segment));
+            }
+
+            // echo case
+            if (TryGetName(message.Key, PrefixHelper.HubConnectionIdPrefix, out var connectionWithHubPrefix))
+            {
+                // naming: hc-{HubName}.{ConnectionId}
+                var result = _serviceConnectionManager.GetPossibleConnections(connectionWithHubPrefix);
```
vwxyzh: naming: `result` -> `connections`

src/Microsoft.Azure.SignalR.AspNet/Internals/ListHelper.cs
```diff
@@ -0,0 +1,16 @@
...
+using System.Collections.ObjectModel;
+
+namespace Microsoft.AspNet.SignalR.Infrastructure
+{
+    /// <summary>
+    /// Copied from https://github.com/SignalR/SignalR/blob/dev/src/Microsoft.AspNet.SignalR.Core/Infrastructure/ListHelper.cs
+    /// </summary>
+    internal class ListHelper<T>
+    {
+        public static readonly IList<T> Empty = new ReadOnlyCollection<T>(new List<T>());
```
vwxyzh: Why not `Array.Empty<T>()`?

vicancy: It is copied from signalr repo and I tend to keep it as it is.

vicancy:
cc @davidfowl Updated the implementation to contain a per app connection, and pass the appName through claim to the client, please take a look. 😄

src/Microsoft.Azure.SignalR.AspNet/OwinExtensions.cs
```diff
@@ -18,58 +18,67 @@ namespace Owin
 {
     public static partial class OwinExtensions
     {
-        public static IAppBuilder MapAzureSignalR(this IAppBuilder builder)
+        public static IAppBuilder MapAzureSignalR(this IAppBuilder builder, string appName)
```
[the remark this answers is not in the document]
vicancy: Separate it into another PR https://github.com/Azure/azure-signalr/pull/190

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ServiceMessageBus.cs
```diff
@@ -2,24 +2,190 @@
...
         public ServiceMessageBus(IDependencyResolver resolver) : base(resolver)
         {
+            _serviceConnectionManager = resolver.Resolve<IServiceConnectionManager>() ?? throw new ArgumentNullException(nameof(IServiceConnectionManager));
+            _serializer = resolver.Resolve<JsonSerializer>() ?? throw new ArgumentNullException(nameof(JsonSerializer));
+            _ackHandler = resolver.Resolve<IAckHandler>() ?? throw new ArgumentNullException(nameof(IAckHandler));
+            _serviceProtocol = resolver.Resolve<IServiceProtocol>() ?? throw new ArgumentNullException(nameof(IServiceProtocol));
+            _pool = resolver.Resolve<IMemoryPool>() ?? throw new ArgumentNullException(nameof(IMemoryPool));
         }
 
         public override Task Publish(Message message)
```
davidfowl: Lets move all of the parsing to a separate component that returns "addresses".

vicancy: Move inside a [SignalRMessageParser](https://github.com/Azure/azure-signalr/pull/152/files#diff-88b66ec494a1d6fe9bb4b5c7d496f26f)

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/SignalRMessageParser.cs
```diff
@@ -0,0 +1,216 @@
...
+        private readonly IMemoryPool _pool;
+        private readonly IServiceProtocol _serviceProtocol;
+
+        public SignalRMessageParser(IReadOnlyList<string> hubs, IDependencyResolver resolver)
+        {
+            _serializer = resolver.Resolve<JsonSerializer>() ?? throw new ArgumentNullException(nameof(JsonSerializer));
+            _serviceProtocol = resolver.Resolve<IServiceProtocol>() ?? throw new ArgumentNullException(nameof(IServiceProtocol));
+            _pool = resolver.Resolve<IMemoryPool>() ?? throw new ArgumentNullException(nameof(IMemoryPool));
+
+            foreach(var hub in hubs)
```
davidfowl: nit: Spacing between foreach and (

davidfowl: Add a comment about what these hub names look like. Are these the full type name in the default case?

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/SignalRMessageParser.cs
```diff
@@ -0,0 +1,216 @@
...
+                // It is possible that the hub contains dot character, while the fully qualified name is formed as {HubName}.{Name} (Name can be connectionId or userId or groupId)
+                // So keep a copy of the hub names containing dots and return all the possible combinations when the fully qualified name is provided
+                if (hub.IndexOf(DotChar) > -1)
+                {
+                    _hubNameWithDots.Add(hub);
+                }
+            }
+        }
+
+        public IEnumerable<AppMessage> GetMessages(Message message)
```
davidfowl: I want to see this method super unit tested with all of the bad cases.

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/SignalRMessageParser.cs
```diff
@@ -0,0 +1,216 @@
...
+            }
+            else
+            {
+                throw new NotSupportedException($"Message {message.Key} is not supported.");
+            }
+        }
+
+        private ReadOnlyMemory<byte> GetPayload(Message message)
+        {
+            IJsonWritable value = new PersistentResponse(m => false, tw => tw.Write("Cursor"))
```
davidfowl: I think we should change "Cursor" to something more obscure so that nobody pays attention to it.

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/SignalRMessageParser.cs
vicancy: They are fetched from `IHubManager.GetHubs()`'s `Name` property, and by default it is the type `Name` instead of `FullName`

vicancy: oops

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/SignalRMessageParser.cs
vicancy: Is it that anything is ok? how about empty string?

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/SignalRMessageParser.cs
davidfowl: Can you add that comment?

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/SignalRMessageParser.cs
davidfowl: Maybe null?

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/SignalRMessageParser.cs
vicancy: sure

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ServiceMessageBus.cs
```diff
@@ -2,24 +2,74 @@
...
-    /// </summary>
     internal class ServiceMessageBus : MessageBus
     {
+        private readonly IMessageParser _parser;
+        private readonly IServiceConnectionManager _serviceConnectionManager;
+        private readonly IAckHandler _ackHandler;
+
         public ServiceMessageBus(IDependencyResolver resolver) : base(resolver)
         {
+            _serviceConnectionManager = resolver.Resolve<IServiceConnectionManager>() ?? throw new ArgumentNullException(nameof(IServiceConnectionManager));
```
davidfowl: This will be problematic but we'll see the reports of what doesn't work with custom DI containers.

davidfowl: There's some weirdness here around passing in the resolver to the constructor. It probably won't be the right one if you override the `IDependencyResolver`.

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ServiceMessageBus.cs
```diff
@@ -2,24 +2,74 @@
...
+
+        private async Task WriteMessage(IServiceConnectionContainer connection, AppMessage appMessage)
+        {
+            var message = appMessage.Message;
+            switch (message)
+            {
+                // For group related messages, make sure messages are written to the same partition
+                case JoinGroupMessage joinGroupMessage:
+                    await connection.WriteAsync(joinGroupMessage.GroupName, joinGroupMessage);
+                    _ackHandler.TriggerAck(appMessage.RawMessage.CommandId);
```
davidfowl: This needs to always fire in case of a group command. Can WriteAsync ever throw?

vicancy: It is possible, wrap into a `finally` block

test/Microsoft.Azure.SignalR.AspNet.Tests/SignalRMessageParserTest.cs
```diff
@@ -0,0 +1,310 @@
...
+        [InlineData("")]
+        [InlineData("key")]
+        [InlineData("pc-SignalRMessageParserTest")]
+        [InlineData("pcg-GroupSignalRMessageParserTest")]
+        [InlineData("c-GroupSignalRMessageParserTest")]
+        public void TestGetMessageWithUnknownKeyThrows(string key)
+        {
+            var hubs = new List<string> { };
+            var parser = new SignalRMessageParser(hubs, _resolver);
+            var raw = "<script type=\"\"></script>";
```
davidfowl: Why a script tag?

vicancy: no special reason, to test various chars

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ServiceMessageBus.cs
vicancy: Add a TODO for this

PR #153, opened 2018-08-07
Extract a base connection class to for service connection and utilities

vicancy:
1. For AspNet support, the underlying server-service connection is also leveraging the aspnetcore http transport logic. So a base connection class is extracted for feature sharing, including:
  a. KeepAlive
  b. Process handshake request and response
  c. Process and dispatch incoming messages
2. Move utility classes here, so that it can be shared by AspNet SDK
3. The newly added Common project is Marked as `IsPackable = false` so that it will not get published, instead, it will be copied to the sdk package when `dotnet pack`. (Is it meaningful to netcore projects?)


src/Microsoft.Azure.SignalR.Common/Microsoft.Azure.SignalR.Common.csproj
```diff
@@ -0,0 +1,17 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Microsoft.Azure.SignalR.Protocols\Microsoft.Azure.SignalR.Protocols.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="System.IdentityModel.Tokens.Jwt" Version="$(SystemIdentityModelTokensJwtPackageVersion)" />
+    <PackageReference Include="Microsoft.AspNetCore.Http.Connections.Client" Version="$(MicrosoftAspNetCoreHttpConnectionsClientPackageVersion)" />
+    <PackageReference Include="Microsoft.AspNetCore.SignalR" Version="$(MicrosoftAspNetCoreSignalRPackageVersion)" />
```
davidfowl: This can't be a dependency in the shared code.

vicancy: oops

src/Microsoft.Azure.SignalR.Common/StaticRandom.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace Microsoft.Azure.SignalR
 {
-    internal class StaticRandom
+    public class StaticRandom
```
davidfowl: I don't like that this is public.

src/Microsoft.Azure.SignalR.Common/TimerAwaitable.cs
```diff
@@ -1,109 +1,109 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Runtime.CompilerServices;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.SignalR
+{
+    public class TimerAwaitable : IDisposable, ICriticalNotifyCompletion
```
davidfowl: I don't think this should be public. Lets us InternalsVisibleTo

src/Microsoft.Azure.SignalR.Common/ServiceConnectionBase.cs
```diff
@@ -0,0 +1,641 @@
...
+
+        protected abstract Task<ConnectionContext> CreateConnection();
+
+        protected abstract Task DisposeConnection();
+
+        protected abstract Task CleanupConnections();
+
+        protected abstract Task OnConnectedAsync(OpenConnectionMessage openConnectionMessage);
+
+        protected abstract Task PerformDisconnectAsync(CloseConnectionMessage closeConnectionMessage);
```
davidfowl: nit: Rename to `OnDisconnectedAsync`

PR #159, opened 2018-08-13
Implement ServiceConnection for Azure AspNet Service

vicancy:
Work on #101 

1. Use `ClientConnectionManager` to create client connections, unlike the ASP.NET Core one, there is no need to maintain a globally available client connections here, because *heartbeat* is managed by Service side, so there is only *service<->server* and *service<->client* heartbeat, but no *server<->client* heartbeat.
2. Use `ServiceConnection` for the underlying *service<->server* connection, as similar to the ASP.NET Core one, it is based on `ServiceConnectionBase`
3. Add unit tests for them. The one for `ServiceConnection` is not completed, needs integration when `ServiceMessageBus` is implemented.


src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ClientConnectionManager.cs
```diff
@@ -0,0 +1,131 @@
...
+                    Debug.Fail("Response StatusCode is " + hostContext.Response.StatusCode);
+                    var errorResponse = GetMemoryString(responseStream);
+                    throw new InvalidOperationException(errorResponse);
+                }
+                return (AzureTransport)hostContext.Environment[Constants.Context.AzureSignalRTransportKey];
+            }
+
+            // This happens when hub is not found
+            Debug.Fail("Unauthorized");
+            throw new InvalidOperationException("Unable to authorize request");
```
davidfowl: Will this end up sending a close?

vicancy: Yes, in https://github.com/Azure/azure-signalr/pull/159/files/74cd48519753460f87f4e030b18068447ef66576#diff-5e3f883b4759d1fe0925a6e812dc026bR122

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ClientConnectionManager.cs
```diff
@@ -0,0 +1,131 @@
...
+            "exp", // Expiration time claims. A token is valid only before its expiration time.
+            "iat", // Issued At claim. Added by default. It is not validated by service.
+            "nbf"  // Not Before claim. Added by default. It is not validated by service.
+        };
+
+        private readonly PersistentConnection _dispatcher;
+
+        public ClientConnectionManager(IDependencyResolver resolver)
+        {
+            _dispatcher = resolver.Resolve<PersistentConnection>() ?? throw new ArgumentNullException(nameof(resolver));
```
davidfowl: This doesn't look right. Why are we resolving a PersistentConnection here?

vicancy: used to `_dispatcher.Authorize` and `_dispatcher.ProcessRequest` when `CreateConnection` https://github.com/Azure/azure-signalr/pull/159/files/74cd48519753460f87f4e030b18068447ef66576#diff-4a1af73f4f7a69cde0fb2ba90b135438R44

davidfowl: I think we need a dispatcher but this shouldn't be in the IDependencyResolver. It's the wrong type.

davidfowl: We should be creating one of these per connection, not resolving it from the container.

vicancy: Sure

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ClientConnectionManager.cs
```diff
@@ -0,0 +1,135 @@
...
+
+            response.Body = responseStream;
+
+            var user = request.User = GetUserPrincipal(message);
+
+            request.Path = new PathString("/");
+
+            var userToken = string.IsNullOrEmpty(user.Identity.Name) ? string.Empty : ":" + user.Identity.Name;
+
+            var queryCollection = HttpUtility.ParseQueryString(message.QueryString ?? string.Empty);
```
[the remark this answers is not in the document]
vicancy: We need override `connectionToken` with `{connectionId}{userToken}` format. There is check inside `ProcessRequest`, refer to https://github.com/SignalR/SignalR/blob/dev/src/Microsoft.AspNet.SignalR.Core/PersistentConnection.cs#L324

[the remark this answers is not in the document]
davidfowl: This is another thing we should fix.

TryGetConnectionId should just be made virtual.

src/Microsoft.Azure.SignalR.AspNet/Constants.cs
```diff
@@ -7,8 +7,14 @@ internal static class Constants
     {
         public static class ClaimType
```
xscript: Shall we move this class to the `Common` project? AspNetCore SignalR can use the same claim types.

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ClientConnectionManager.cs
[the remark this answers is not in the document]
vicancy: cc @anurse

src/Microsoft.Azure.SignalR.AspNet/Constants.cs
vicancy: Sure, please check out https://github.com/Azure/azure-signalr/pull/164

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ServiceConnection.cs
```diff
@@ -0,0 +1,151 @@
...
+            var connection = _connection;
+            _connection = null;
+            return _connectionFactory.DisposeAsync(connection);
+        }
+
+        protected override Task CleanupConnections()
+        {
+            try
+            {
+                foreach (var connectionId in _clientConnections.Keys)
```
davidfowl: Performance tip: It's more efficient to loop over the ConcurrentDictionary itself rather than keys or values. Each of those collections takes a lock on all of the buckets.

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ServiceConnection.cs
```diff
@@ -0,0 +1,151 @@
...
+        {
+            if (_clientConnections.TryRemove(connectionId, out var transport))
+            {
+                transport.OnDisconnected();
+            }
+
+            Log.ConnectedEnding(_logger, connectionId);
+        }
+
+        private void OnReceived(AzureTransport transport, ReadOnlySequence<byte> payload)
```
davidfowl: Why does this method exist? Just call it directly from the OnMessageAsync

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ClientConnectionManager.cs
```diff
@@ -0,0 +1,135 @@
...
+                }
+            }
+
+            return new ClaimsPrincipal(new ClaimsIdentity(claims, authenticationType));
+        }
+
+        internal static string GetMemoryString(MemoryStream stream)
+        {
+            stream.Seek(0, SeekOrigin.Begin);
+            using (var reader = new StreamReader(stream))
```
vwxyzh: using `StreamReader` will close the stream, but from the method name, I haven't realized it.

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ClientConnectionManager.cs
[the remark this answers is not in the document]
vicancy: Add a todo here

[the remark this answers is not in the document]
vicancy: Create https://github.com/SignalR/SignalR/issues/4175 to track work in signalr

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ClientConnectionManager.cs
vicancy: how about `GetContentAndDispose`?

PR #163, opened 2018-08-16
Send UpdateConnectionMessage when there is custom IUserIdProvider

### LINKED ISSUE #140: SignalR Azure does not seem to use provided id from custom IUserIdProvider
Hi

I first opened an [issue](https://github.com/aspnet/SignalR/issues/2602) in [aspnet/signalr](https://github.com/aspnet/SignalR) repository, but the problem seem to be related to the library `Microsoft.Azure.SignalR` on top of SignalR.

Experienced issue:
Using a custom user id provider implementing `IUserIdProvider` does not seem to work. I can see it hitting my breakpoint, but does not use my custom user id.

Without using the Azure Signal R service, it work as expected. But with Azure service it does not seem to have any effect what I return other than the JWT claim `NameIdentifier` used as default.

How to recreate:
1. Create a new web project `dotnet new mvc`
2. Add the package `Microsoft.Azure.SignalR`
3. Add the client library from npm
4. Implement a custom user id provider (`IUserIdProvider`) and register it in `Startup.cs`
5. Return some custom id from the class above.
6. Try to send a message from the client using the custom id specified.

- .Net Core 2.1.1
- Microsoft.Azure.SignalR 1.0.0-preview1-10015
- @aspnet/signalr 1.0.0
- Javascript client from Latest Chrome
- Kestrel
- Ubuntu 18.04 (Tried on Windows 10 as well)

xscript:
This PR is part of the fix to #140 .

### DIFF
--- a/src/Microsoft.Azure.SignalR.Protocols/ConnectionMessage.cs
+++ b/src/Microsoft.Azure.SignalR.Protocols/ConnectionMessage.cs
@@ -130,4 +130,25 @@ namespace Microsoft.Azure.SignalR.Protocol
         /// </summary>
         public ReadOnlySequence<byte> Payload { get; set; }
     }
+
+    /// <summary>
+    /// A update-connection message.
+    /// </summary>
+    public class UpdateConnectionMessage : ConnectionMessage
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="UpdateConnectionMessage"/> class.
+        /// </summary>
+        /// <param name="connectionId">The connection Id.</param>
+        /// <param name="userId">New user Id of the connection.</param>
+        public UpdateConnectionMessage(string connectionId, string userId) : base(connectionId)
+        {
+            UserId = userId;
+        }
+
+        /// <summary>
+        /// Gets or sets connection's new user Id.
+        /// </summary>
+        public string UserId { get; set; }
+    }
 }
--- a/src/Microsoft.Azure.SignalR/DependencyInjectionExtensions.cs
+++ b/src/Microsoft.Azure.SignalR/DependencyInjectionExtensions.cs
@@ -65,7 +65,8 @@ namespace Microsoft.Extensions.DependencyInjection
                 .AddSingleton(typeof(ServiceHubDispatcher<>))
                 .AddSingleton(typeof(AzureSignalRMarkerService))
                 .AddSingleton<IClientConnectionFactory, ClientConnectionFactory>()
-                .AddSingleton<IHostedService, HeartBeat>();
+                .AddSingleton<IHostedService, HeartBeat>()
+                .AddSingleton<NegotiateHandler>();
             return builder;
         }
     }
--- a/src/Microsoft.Azure.SignalR/NegotiateHandler.cs
+++ b/src/Microsoft.Azure.SignalR/NegotiateHandler.cs
@@ -1,0 +1,60 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.Security.Claims;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Connections;
+using Microsoft.AspNetCore.SignalR;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.SignalR
+{
+    internal class NegotiateHandler
+    {
+        private readonly IServiceEndpointProvider _endpointProvider;
+        private readonly bool _isDefaultUserIdProvider;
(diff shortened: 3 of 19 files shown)


src/Microsoft.Azure.SignalR/DependencyInjectionExtensions.cs
```diff
@@ -1,72 +1,73 @@
...
+                .AddSingleton(typeof(HubLifetimeManager<>), typeof(ServiceLifetimeManager<>))
+                .AddSingleton(typeof(IServiceProtocol), typeof(ServiceProtocol))
+                .AddSingleton(typeof(IClientConnectionManager), typeof(ClientConnectionManager))
+                .AddSingleton(typeof(IServiceConnectionManager<>), typeof(ServiceConnectionManager<>))
+                .AddSingleton(typeof(IServiceEndpointProvider), typeof(ServiceEndpointProvider))
+                .AddSingleton(typeof(ServiceHubDispatcher<>))
+                .AddSingleton(typeof(AzureSignalRMarkerService))
+                .AddSingleton<IClientConnectionFactory, ClientConnectionFactory>()
+                .AddSingleton<IHostedService, HeartBeat>()
+                .AddSingleton<NegotiateHandler>();
```
xscript: This is the updated line. Not sure why the whole file is marked as updated.

src/Microsoft.Azure.SignalR.Protocols/ConnectionMessage.cs
```diff
@@ -130,4 +130,25 @@ public ConnectionDataMessage(string connectionId, ReadOnlySequence<byte> payload
         /// </summary>
         public ReadOnlySequence<byte> Payload { get; set; }
     }
+
+    /// <summary>
+    /// A update-connection message.
+    /// </summary>
+    public class UpdateConnectionMessage : ConnectionMessage
```
davidfowl: Why is it called UpdateConnection?

src/Microsoft.Azure.SignalR/DependencyInjectionExtensions.cs
```diff
@@ -1,72 +1,73 @@
-// Copyright (c) Microsoft. All rights reserved.
```
davidfowl: What happened here?

src/Microsoft.Azure.SignalR.Protocols/ConnectionMessage.cs
xscript: Want to make it more general so we might add more fields later. Any suggestion?

davidfowl: So the intent is that we may add more fields in the future and only the ones that have values marked things that have changed?

src/Microsoft.Azure.SignalR/NegotiateHandler.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Collections.Generic;
+using System.Security.Claims;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Connections;
+using Microsoft.AspNetCore.SignalR;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.SignalR
+{
+    internal class NegotiateHandler
```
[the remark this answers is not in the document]
xscript: To align with `NegotiateProtocol`.

src/Microsoft.Azure.SignalR/NegotiateHandler.cs
```diff
@@ -0,0 +1,60 @@
...
+        private IEnumerable<Claim> BuildClaims(HttpContext context)
+        {
+            var claims = _claimsProvider?.Invoke(context) ?? context.User.Claims;
+
+            if (_isDefaultUserIdProvider)
+            {
+                return claims;
+            }
+
+            // Add an empty user Id claim to tell service that user has a custom IUserIdProvider.
```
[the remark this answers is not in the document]
xscript: To make service-side logic simpler. Will always use the value in CustomUserId claim if it is present.

src/Microsoft.Azure.SignalR/DependencyInjectionExtensions.cs
BrennanConroy: Are any of these services extension points for users? If so they should be `TryAddSingleton`

xscript: Good point. In our case, none of them are extensible for user. All of them are internal, except two of them are public class/interface.
- `class HubLifetimeManager` is `public` and `abstract`. The implementation of `ServiceLifetimeManager` is service specific. So we don't want it to be overridden.
- `interface IServiceProtocol` is `public`. We don't want it to be overridden either.

PR #165, opened 2018-08-16
Shorten AzureSignalRPrefix in claimType, as longer one will be result…

vicancy:
…ed in longer url

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/Constants.cs
+++ b/src/Microsoft.Azure.SignalR.Common/Constants.cs
@@ -10,8 +10,8 @@ namespace Microsoft.Azure.SignalR
         public static class ClaimType
         {
             public const string AzureSignalRSysPrefix = "asrs.s.";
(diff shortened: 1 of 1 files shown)


src/Microsoft.Azure.SignalR.Common/Constants.cs
```diff
@@ -9,9 +9,11 @@ internal static class Constants
     {
         public static class ClaimType
         {
-            public const string AzureSignalRSysPrefix = "azure.signalr.sys.";
-            public const string AuthenticationType = AzureSignalRSysPrefix + "authenticationtype";
-            public const string UserId = AzureSignalRSysPrefix + "userid";
+            public const string AzureSignalRSysPrefix = "asrs.s.";
+            public const string AuthenticationType = AzureSignalRSysPrefix + "au";
+            public const string UserId = AzureSignalRSysPrefix + "ur";
```
xscript: `uid` might be more clear.

PR #170, opened 2018-08-22
Use IUserIdProvider to generate user id and include it in access token

xscript:


davidfowl:
Is this a revert?

xscript:
@davidfowl  Yeah, And also add the implementation we talked about yesterday.

davidfowl:
Lets split those into different PRs

xscript:
Sure. Let me create another PR to do the revert.

src/Microsoft.Azure.SignalR/ServiceHubConnectionContext.cs
```diff
@@ -0,0 +1,73 @@
...
+        public const string ConnectionIdUnavailableError = nameof(ConnectionId) + UnavailableErrorTemplate;
+        public const string ItemsUnavailableError = nameof(Items) + UnavailableErrorTemplate;
+        public const string ProtocolUnavailableError = nameof(Protocol) + UnavailableErrorTemplate;
+
+        public ServiceHubConnectionContext(HttpContext context)
+            : base(new DummyConnectionContext(context), TimeSpan.MaxValue, NullLoggerFactory.Instance)
+        {
+        }
+
+        public override CancellationToken ConnectionAborted => throw new InvalidOperationException(ConnectionAbortedUnavailableError);
```
davidfowl: Why throw here? Most of these properties can just do nothing.

src/Microsoft.Azure.SignalR/ServiceHubConnectionContext.cs
```diff
@@ -0,0 +1,73 @@
...
+namespace Microsoft.Azure.SignalR
+{
+    /// <summary>
+    /// ServiceHubConnectionContext is only passed to IUserIdProvider.GetUserId as a paramter.
+    /// Only HttpContext and User properties are available for access.
+    /// Exception will be thrown when other properties are accessed.
+    /// </summary>
+    internal class ServiceHubConnectionContext : HubConnectionContext
+    {
+        public const string UnavailableErrorTemplate = " property is not available in current context of Azure SignalR Service.";
```
davidfowl: Remove "current"

src/Microsoft.Azure.SignalR/ServiceHubConnectionContext.cs
xscript: Per offline sync, we will leave these exceptions as they are now to prevent users using these properties.

src/Microsoft.Azure.SignalR/ServiceHubConnectionContext.cs
xscript: Sure.

PR #171, opened 2018-08-22
Revert "Send UpdateConnectionMessage when there is custom IUserIdProvider (#163)"

xscript:
Revert  #163


davidfowl:
This was a manual revert right? You didnt; run git revert? I still see the NegotiateHandler

xscript:
Ah, you want to a full revert. Yeah, this is a manual revert because I want to keep `NegotiateHandler`.
Let me update it now.

PR #174, opened 2018-08-23
Preserve query string in negotiate request

xscript:


src/Microsoft.Azure.SignalR/EndpointProvider/IServiceEndpointProvider.cs
```diff
@@ -1,34 +1,22 @@
...
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.
-
-using System;
-using System.Collections.Generic;
-using System.Security.Claims;
-using Microsoft.AspNetCore.SignalR;
-
-namespace Microsoft.Azure.SignalR
-{
-    internal interface IServiceEndpointProvider
```
xscript: Remove unused methods and properties.

src/Microsoft.Azure.SignalR/EndpointProvider/ServiceEndpointProvider.cs
```diff
@@ -1,229 +1,190 @@
...
+
+        public string GenerateServerAccessToken<THub>(string userId, TimeSpan? lifetime = null) where THub : Hub
+        {
+            var audience = _generator.GetServerAudience(typeof(THub).Name);
+            var claims = userId != null ? new[] {new Claim(ClaimTypes.NameIdentifier, userId)} : null;
+
+            return InternalGenerateAccessToken(audience, claims, lifetime ?? _accessTokenLifetime);
+        }
+
+        public string GetClientEndpoint(string hubName, QueryString queryString)
```
xscript: Changes are mainly in this method.

src/Microsoft.Azure.SignalR/EndpointProvider/ServiceEndpointProvider.cs
```diff
@@ -1,229 +1,107 @@
...
+        {
+            if (string.IsNullOrEmpty(hubName))
+            {
+                throw new ArgumentNullException(nameof(hubName));
+            }
+
+            var endpoint = _generator.GetClientEndpoint(hubName);
+
+            if (queryString == QueryString.Empty)
```
vicancy: Also check null?

src/Microsoft.Azure.SignalR/EndpointProvider/ServiceEndpointProvider.cs
```diff
@@ -1,229 +1,107 @@
...
+            {
+                return endpoint;
+            }
+
+            return endpoint.Contains("?")
```
vicancy: Is it possible that endpoint can contain `#a?b`? Or `_generator` supports QueryString and form the final URL directly instead?

test/Microsoft.Azure.SignalR.Tests/ConnectionStringParserFacts.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace Microsoft.Azure.SignalR.Tests
 {
-    public class ConnectionStringParsingFacts
+    public class ConnectionStringParserFacts
```
vicancy: extremely nit: why it is *Facts* when it contains *Theories*...

src/Microsoft.Azure.SignalR/EndpointProvider/ServiceEndpointProvider.cs
xscript: `QueryString` is a struct (ValueType). It can never be `null`.

test/Microsoft.Azure.SignalR.Tests/ConnectionStringParserFacts.cs
xscript: I don't think `Facts` in the class name is the same term of `[Facts]` attribute. In the class name, it is basically interchangeable with `Tests`, defining the behavior of `ConnectionStringParser` .

src/Microsoft.Azure.SignalR/EndpointProvider/ServiceEndpointProvider.cs
xscript: Not possible for now.

I think you are right. If we move `QueryString` to `_generator`, we don't have to do this check then.

xscript:
Per offline sync with @davidfowl , SignalR client should preserve the query strings. Open https://github.com/aspnet/SignalR/issues/2859 to track it.

PR #176, opened 2018-08-24
Refactor ServiceEndpointProvider

xscript:


src/Microsoft.Azure.SignalR/EndpointProvider/ConnectionStringParser.cs
```diff
@@ -0,0 +1,99 @@
...
+{
+    internal static class ConnectionStringParser
+    {
+        private const string EndpointProperty = "endpoint";
+        private const string AccessKeyProperty = "accesskey";
+        private const string VersionProperty = "version";
+        private const string PortProperty = "port";
+        // For SDK 1.x, only support Azure SignalR Service 1.x
+        private const string SupportedVersion = "1";
+        private const string ValidVersionRegex = "^" + SupportedVersion + @"\.\d+(?:[\w-.]+)?$";
```
davidfowl: Can't we just try parse the version and avoid using regex?

xscript: Just copied over here. And it is just a one-time thing, so it shouldn't be a big issue?

PR #178, opened 2018-08-27
Include CustomUserId claim as internal claims

xscript:


src/Microsoft.Azure.SignalR/HubHost/ServiceConnectionContext.cs
```diff
@@ -33,7 +33,8 @@ internal class ServiceConnectionContext : ConnectionContext,
             "aud", // Audience claim, used by service to make sure token is matched with target resource.
             "exp", // Expiration time claims. A token is valid only before its expiration time.
             "iat", // Issued At claim. Added by default. It is not validated by service.
-            "nbf"  // Not Before claim. Added by default. It is not validated by service.
+            "nbf", // Not Before claim. Added by default. It is not validated by service.
+            Constants.ClaimType.UserId // Custom user id claim.
```
vicancy: leverage the `ClaimType.SysPrefix`?

xscript: Yeah, that would be cleaner.

src/Microsoft.Azure.SignalR/HubHost/ServiceConnectionContext.cs
```diff
@@ -115,20 +115,12 @@ public void TickHeartbeat()
...
-                        return true;
-                    }
-                }
-            }
-
-            return false;
+            return claims?.Count > 0 &&
+                   claims.Any(claim =>
+                       !SystemClaims.Contains(claim.Type) &&
+                       !claim.Type.StartsWith(Constants.ClaimType.AzureSignalRSysPrefix, StringComparison.OrdinalIgnoreCase));
```
vicancy: `Ordinal` is enough

xscript: Let's make it safer.

PR #181, opened 2018-08-29
Move ConnectionStringParser and IServiceEndpointGenerator to Common p…

vicancy:
…roject


xscript:
Moving test case too?

PR #183, opened 2018-08-29
[ASP.NET SignalR]Add a sample project for old signalr

vicancy:
This sample includes 
1. a simple one following https://docs.microsoft.com/en-us/aspnet/signalr/overview/getting-started/tutorial-getting-started-with-signalr

This sample is to get ready for switching to Azure 😄


xscript:
Can `dotnet` CLI build such hybrid solution?

PR #184, opened 2018-08-29
Include original query string in negotiate response

xscript:


src/Microsoft.Azure.SignalR.Common/IServiceEndpointGenerator.cs
```diff
@@ -1,12 +1,14 @@
...
 
+using Microsoft.AspNetCore.Http;
+
 namespace Microsoft.Azure.SignalR
 {
     internal interface IServiceEndpointGenerator
     {
         string GetClientAudience(string hubName);
-        string GetClientEndpoint(string hubName, string originalPath);
+        string GetClientEndpoint(string hubName, string originalPath, QueryString queryString);
```
xscript: Thinking about moving it back to `Microsoft.Azure.SignalR` project, as this interface won't be the same as `Microsoft.Azure.SignalR.AspNet`.
/cc @vicancy

vicancy: `Microsoft.Azure.SignalR.AspNet` is not using it. But I am thinking if I should use it so that we can aware such change for `Microsoft.Azure.SignalR.AspNet`, is `queryString` also needed for `Microsoft.Azure.SignalR.AspNet`?

xscript: I don't think you will need it. Client SDK will preserve the original query string when getting redirect response.

vicancy: >QueryString [](start = 70, length = 11)

how about using `string` instead of `QueryString` to get rid of the `Microsoft.AspNetCore.Http` dependency?

xscript: It is doable. Just with `QueryString`, we can save quite some checks.

PR #186, opened 2018-09-03
[ASP.NET SignalR Service]Update the negotiate response

vicancy:
1. `ProtocolVersion` up to 2.0
2. Use `RedirectUrl` instead of already in-use `Url`


davidfowl:
Why 2.0? Isn't the current version 1.4?

PR #187, opened 2018-09-03
try to support /reconnect

vicancy:


src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ServiceConnection.cs
```diff
@@ -90,7 +90,15 @@ protected override Task OnMessageAsync(ConnectionDataMessage connectionDataMessa
...
                     Log.WriteMessageToApplication(_logger, payload.Length, connectionDataMessage.ConnectionId);
-                    transport.OnReceived(GetString(payload));
+                    var message = GetString(payload);
+                    if (message == "Asrs.Reconnect")
+                    {
+                        transport.Reconnected?.Invoke();
+                    }
+                    else
+                    {
+                        transport.OnReceived(GetString(payload));
```
davidfowl: You're calling GetString twice.

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ServiceConnection.cs
```diff
@@ -90,7 +91,15 @@ protected override Task OnMessageAsync(ConnectionDataMessage connectionDataMessa
                 {
                     var payload = connectionDataMessage.Payload;
```
davidfowl: This code needs to be changed to parse text the text message separator.

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ServiceConnection.cs
```diff
@@ -17,6 +17,7 @@ namespace Microsoft.Azure.SignalR.AspNet
 {
     internal partial class ServiceConnection : ServiceConnectionBase
     {
+        private const string ReconnectMessage = "{\"Reconnect\":true}";
```
davidfowl: This makes me semi nervous that we aren't parsing the message

vicancy: Feels no need to ...though

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ServiceConnection.cs
vicancy: already parsed when dispatching message from the runtime side.

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ServiceConnection.cs
davidfowl: Hopefully nobody changes the spacing?

vicancy: they shouldn't. Parsing to Jason is too costly.....

PR #190, opened 2018-09-07
Add application name to the public API

vicancy:
1. Add application name to the public API
2. add comments
3. add an app connection to the service connection
4. add logger factory

### DIFF
--- a/src/Microsoft.Azure.SignalR.AspNet/OwinExtensions.cs
+++ b/src/Microsoft.Azure.SignalR.AspNet/OwinExtensions.cs
@@ -153,7 +153,7 @@ namespace Owin
             }
             else
             {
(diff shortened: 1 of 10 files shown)


src/Microsoft.Azure.SignalR.AspNet/OwinExtensions.cs
```diff
@@ -78,11 +153,11 @@ private static void RunAzureSignalRCore(IAppBuilder builder, HubConfiguration co
             }
             else
             {
-                // TODO: log something
+                logger.CreateLogger<IAppBuilder>().Log(LogLevel.Information, "No hub is found.");
```
davidfowl: Log this as a warning, "No hubs found"

Issue #198, opened 2018-09-13
The 'User-Agent' header must be modified using the appropriate property or method.


hempe:
I'm using Microsoft.Azure.SignalR 1.0.0-preview1-10200, Target platform net461.

When i try to use AzureSignalR I see the following error in the logs, and it seems no connection to the hub is established.

```
[09:24:01 ERR] Failed to connect to the service.
System.ArgumentException: The 'User-Agent' header must be modified using the appropriate property or method.
Parameter name: name
   at System.Net.WebHeaderCollection.ThrowOnRestrictedHeader(String headerName)
   at System.Net.WebHeaderCollection.Set(String name, String value)
   at Microsoft.AspNetCore.Http.Connections.Client.Internal.WebSocketsTransport..ctor(HttpConnectionOptions httpConnectionOptions, ILoggerFactory loggerFactory, Func`1 accessTokenProvider)
   at Microsoft.AspNetCore.Http.Connections.Client.Internal.DefaultTransportFactory.CreateTransport(HttpTransportType availableServerTransports)
   at Microsoft.AspNetCore.Http.Connections.Client.HttpConnection.<StartTransport>d__47.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Http.Connections.Client.HttpConnection.<SelectAndStartTransport>d__44.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Http.Connections.Client.HttpConnection.<StartAsyncCore>d__41.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Microsoft.AspNetCore.Http.Connections.Client.HttpConnection.<StartAsync>d__40.MoveNext()
[...]

PR #200, opened 2018-09-14
rename custom header name to avoid header restriction in asp.net framework

JialinXin:
fix issue [#198](https://github.com/Azure/azure-signalr/issues/198)

### DIFF
--- a/src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
+++ b/src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
@@ -27,7 +27,7 @@ namespace Microsoft.Azure.SignalR
         private readonly IClientConnectionFactory _clientConnectionFactory;
         private readonly string _userId;
         private static readonly string Name = $"ServiceHubDispatcher<{typeof(THub).FullName}>";
-        private static Dictionary<string, string> CustomHeader = new Dictionary<string, string> { { "User-Agent", ProductInfo.GetProductInfo() } };
+        private static Dictionary<string, string> CustomHeader = new Dictionary<string, string> { { "Asrs-UserAgent", ProductInfo.GetProductInfo() } };
 
         public ServiceHubDispatcher(IServiceProtocol serviceProtocol,
             IServiceConnectionManager<THub> serviceConnectionManager,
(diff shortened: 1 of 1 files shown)


src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
```diff
@@ -27,7 +27,7 @@ internal class ServiceHubDispatcher<THub> : IConnectionFactory where THub : Hub
         private readonly IClientConnectionFactory _clientConnectionFactory;
         private readonly string _userId;
         private static readonly string Name = $"ServiceHubDispatcher<{typeof(THub).FullName}>";
-        private static Dictionary<string, string> CustomHeader = new Dictionary<string, string> { { "User-Agent", ProductInfo.GetProductInfo() } };
+        private static Dictionary<string, string> CustomHeader = new Dictionary<string, string> { { "Asrs-UserAgent", ProductInfo.GetProductInfo() } };
```
xscript: Let's reference #198 here and add comment to explain why we are using a custom header.
Also it helps to include a link to the [source](https://referencesource.microsoft.com/#system/net/System/Net/_HeaderInfoTable.cs,92). Because we don't have test to cover this use case for now.

davidfowl:
I'm assuming you tested this and saw it break then saw it work after this change?

src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
```diff
@@ -27,7 +27,9 @@ internal class ServiceHubDispatcher<THub> : IConnectionFactory where THub : Hub
         private readonly IClientConnectionFactory _clientConnectionFactory;
         private readonly string _userId;
         private static readonly string Name = $"ServiceHubDispatcher<{typeof(THub).FullName}>";
-        private static Dictionary<string, string> CustomHeader = new Dictionary<string, string> { { "User-Agent", ProductInfo.GetProductInfo() } };
+        // Fix issue: https://github.com/Azure/azure-signalr/issues/198
+        // .NET Framework has restriction about reserved string as the header name like "User-Agent"
+        private static Dictionary<string, string> CustomHeader = new Dictionary<string, string> { { "Asrs-UserAgent", ProductInfo.GetProductInfo() } };
```
vicancy: to follow the naming convention, is `Asrs-User-Agent` better?

PR #205, opened 2018-09-17
fix claim in negotiate handler.

vwxyzh:
Bug: add our custom user id in the claim when the user id is null, this will prevent using the standard name identify.


davidfowl:
What was the bug?

xscript:
@davidfowl 
The bug is if `ClaimsProvider` is not null, we will use different set of claims to call `IUserIdProvider.GetUserId` for service runtime and application server.
I believe the correct logic is:
1. Invoke the `ClaimProvider` to get all claims.
2. Call `IUserIdProvider.GetUserId` to get the resolved user id and add the custom head.

PR #206, opened 2018-09-17
Make the one from web.config as the default connection string and fix build failure

vicancy:


davidfowl:
Should we also try app settings?

Issue #191, opened 2018-09-09
Preserve client IP address in Context.HttpContext


alexiordan:
How can I see client remote IP address?

`Context.GetHttpContext().Connection.RemoteIpAddress` is null

In asp.net Core SignalR it worked.

xscript:
For now, you can't. When using Azure SignalR Service, the `HttpContext` is not the `"real"` one but a logical one.
The `"real"` HttpContext is maintained in SignalR Service. The one you can access is rebuilt in your app with limited information.

Could you elaborate more how you'd like to use the client IP? We want to understand more about your scenario.

davidfowl:
@xscript yep, I didnt mean roundtrip from the application, it would be sent like how we handle headers.

alexiordan:
Maybe is related to this: https://github.com/Azure/azure-signalr/issues/25

xscript:
With the latest Service SDK (1.0.0), remote IP address is available.
Close this issue.

PR #207, opened 2018-09-17
Add client IP address in Context.HttpContext

vwxyzh:
#191

### DIFF
--- a/samples/ChatSample/Hub/Chat.cs
+++ b/samples/ChatSample/Hub/Chat.cs
@@ -13,10 +13,10 @@ namespace ChatSample
             Clients.All.SendAsync("broadcastMessage", name, message);
         }
 
-        [Authorize]
+        //[Authorize]
         public void Echo(string name, string message)
         {
-            Clients.Client(Context.ConnectionId).SendAsync("echo", name, message + " (echo from server)");
+            Clients.Client(Context.ConnectionId).SendAsync("echo", name, $"{message} (echo from server, Client IP: {Context.GetHttpContext().Connection.RemoteIpAddress})");
         }
     }
 }
--- a/samples/ChatSample/Hub/Chat.cs
+++ b/samples/ChatSample/Hub/Chat.cs
@@ -13,7 +13,6 @@ namespace ChatSample
             Clients.All.SendAsync("broadcastMessage", name, message);
         }
 
-        //[Authorize]
         public void Echo(string name, string message)
         {
             Clients.Client(Context.ConnectionId).SendAsync("echo", name, $"{message} (echo from server, Client IP: {Context.GetHttpContext().Connection.RemoteIpAddress})");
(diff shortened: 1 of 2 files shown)


samples/ChatSample/Hub/Chat.cs
```diff
@@ -13,10 +13,10 @@ public void BroadcastMessage(string name, string message)
             Clients.All.SendAsync("broadcastMessage", name, message);
         }
 
-        [Authorize]
+        //[Authorize]
```
vwxyzh: If require Authorize, echo not work.
Is it a bug in SDK?

src/Microsoft.Azure.SignalR/HubHost/ServiceConnectionContext.cs
```diff
@@ -142,14 +143,18 @@ private HttpContext BuildHttpContext(OpenConnectionMessage message)
...
                 QueryString = message.QueryString,
                 Path = GetOriginalPath(message.QueryString)
             });
             httpContextFeatures.Set<IHttpAuthenticationFeature>(new HttpAuthenticationFeature
             {
                 User = User
             });
+            httpContextFeatures.Set<IHttpConnectionFeature>(new HttpConnectionFeature
+            {
+                RemoteIpAddress = IPAddress.Parse(message.Headers["X-Real-IP"][0])
```
davidfowl: There's a standard header that reverse proxies normally add for the client IP https://en.wikipedia.org/wiki/X-Forwarded-For

src/Microsoft.Azure.SignalR/HubHost/ServiceConnectionContext.cs
```diff
@@ -142,14 +143,18 @@ private HttpContext BuildHttpContext(OpenConnectionMessage message)
...
-                Headers = new HeaderDictionary((Dictionary<string, StringValues>) message.Headers),
+                Headers = new HeaderDictionary((Dictionary<string, StringValues>)message.Headers),
                 QueryString = message.QueryString,
                 Path = GetOriginalPath(message.QueryString)
             });
             httpContextFeatures.Set<IHttpAuthenticationFeature>(new HttpAuthenticationFeature
             {
                 User = User
             });
+            httpContextFeatures.Set<IHttpConnectionFeature>(new HttpConnectionFeature
```
xscript: We should leave it as empty if there is no specified header we need?

samples/ChatSample/Hub/Chat.cs
xscript: It is added for testing in the old days. It will fail intentionally. You can remove it now.
Also we should add test to cover this scenario in the automation test.

src/Microsoft.Azure.SignalR/HubHost/ServiceConnectionContext.cs
```diff
@@ -142,14 +143,21 @@ private HttpContext BuildHttpContext(OpenConnectionMessage message)
...
             });
             httpContextFeatures.Set<IHttpAuthenticationFeature>(new HttpAuthenticationFeature
             {
                 User = User
             });
+            if (message.Headers.ContainsKey("X-Forwarded-For"))
+            {
+                httpContextFeatures.Set<IHttpConnectionFeature>(new HttpConnectionFeature
+                {
+                    RemoteIpAddress = IPAddress.Parse(message.Headers["X-Forwarded-For"][0])
```
xscript: Check for error cases?

PR #210, opened 2018-09-20
Remove myget source from readme

vicancy:
Sounds unnecessary to expose MyGet feeds to end users

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,10 +11,10 @@ This repository contains the open source subset of the .NET SDK.
 
 ## Nuget Packages
 
-Package Name | Target Framework | Version
----|---|---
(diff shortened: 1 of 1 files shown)


davidfowl:
Why not just have both?

vicancy:
MyGet ones in current CI flow, might not ready to publish, it triggers everytime dev branch has commits. feels no need to add it to Readme, it might be misleading. We can add it back when later we change it to a more officially publish workflow, for example, only tagged ones are published.

davidfowl:
I don't agree. I think it's fine to have both there. Github is where developers come to see the source code and use bleeding edge things. No need to hide it from them.

vicancy:
OK, sounds fair

PR #216, opened 2018-09-26
Update Readme to include AspNet SignalR support

vicancy:
[Rendered view](https://github.com/vicancy/azure-signalr/blob/doc/README.md)


sffamily:
One comment: the first sentence we would like it to be more specific: "Azure SignalR Service SDK for .NET helps you build ASP.NET based Azure web application with ASP.NET Core based Azure SignalR Service." 

Others look good.

PR #217, opened 2018-09-27
Add faq.md

xscript:


README.md
```diff
@@ -42,8 +42,8 @@ Samples on how to use Azure SignalR Service can be found [here](https://github.c
 The following documents describe more details about Azure SignalR Service.
 
 - [Use Azure SignalR Service](./docs/use-signalr-service.md)
-- [REST API support](./docs/rest-api.md)
-- [Azure SignalR Service Internals](./docs/internal.md)
+- [REST API in Azure SignalR Service](./docs/rest-api.md)
+- [Internals of Azure SignalR Service](./docs/internal.md)
```
vicancy: `Inside Azure SignalR Service`?

xscript: Why is this better?

docs/faq.md
```diff
@@ -0,0 +1,72 @@
...
+
+Yes if you are using [ASP.NET Core SignalR](https://github.com/aspnet/SignalR) with Azure SignalR Service.
+The generally-available announcement is at [here](https://azure.microsoft.com/en-us/blog/azure-signalr-service-now-generally-available/).
+
+The support of [ASP.NET SignalR](https://github.com/SignalR/SignalR) is still in *public preview*.
+
+<a name="no-server-available"></a>
+## Client connection is closed with error message "No server available". What does that mean?
+
+This error will happen only when clients are sending messages to SignalR Service.
```
vicancy: `This error will happen` => `This error occurs`?

docs/faq.md
```diff
@@ -0,0 +1,72 @@
...
+
+If you don't have any application server and only use SignalR Service's REST APIs, this is **by design**.
+In serverless architecture, client connections are in **LISTEN** mode and should not send any messages to SignalR Service.
+Read more about [REST API](./rest-api.md) in SignalR Service.
+
+If you have application servers, this means there is no application server connected with your SignalR Service instance.
+Possible root causes are:
```
vicancy: Another possibility: the client sends request too early that server has not yet started the connection to service. This can happen when local debugging the web app with Azure Service. Shall we count this possibility in?

docs/faq.md
```diff
@@ -0,0 +1,72 @@
...
+- There are some connectivity issues inside SignalR Service instances.
+It should be transient and recovered automatically within a couple of minutes.
+If it lasts for over an hour, please [open an issue in GitHub](https://github.com/Azure/azure-signalr/issues/new) or [create an Azure support request](https://docs.microsoft.com/en-us/azure/azure-supportability/how-to-create-azure-support-request).
+Then we can work with you to troubleshoot.
+
+<a name="client-server-mapping"></a>
+## When there are multiple application servers, are messages from clients sent to all of them or just one of them?
+
+It is one-to-one mapped between clients and application servers ([server connections](./internal.md#server-connections) accurately).
+So messages from one client will only be sent to the same one application server.
```
vicancy: Shall we also mention when is the mapping determined? Actually the server client first negotiates to is not the same as the one we go through the service runtime maps to. Shall we also point that out?

docs/faq.md
```diff
@@ -0,0 +1,72 @@
...
+In Azure SignalR Service SDK, `HubConnectionContext context` is a logical client connection context.
+The physical client connection is connected to SignalR Service instance.
+So only limited properties are provided.
+For now, only `HubConnectionContext.GetHttpContext()` and `HubConnectionContext.User` is available for access.
+You can find the source code at [here](https://github.com/Azure/azure-signalr/blob/kevinzha/faq/src/Microsoft.Azure.SignalR/ServiceHubConnectionContext.cs).
+
+<a name="configure-transports"></a>
+## Can I configure available transports in Azure SignalR Service like ASP.NET Core SignalR? Say WebSocket transport is disabled.
+
+No.
```
vicancy: Is the `configure available transports` the client side one https://docs.microsoft.com/en-us/aspnet/core/signalr/configuration?view=aspnetcore-2.1#configure-allowed-transports or server side one? I think the client side one is still configurable?

README.md
vicancy: less words? 😄

docs/faq.md
xscript: Also it has been documented at [the internals](https://github.com/Azure/azure-signalr/blob/dev/docs/internal.md#transport-data-between-client-and-server). FAQ shouldn't be the place of detailed documentation.

docs/faq.md
xscript: This is the server side one. Let me make it less confusing.

README.md
davidfowl: Internals of the

Issue #223, opened 2018-10-05
NameClaimType not honored


NArnott:
I have a sample ASP NET Core web app that I'm hooking up to Azure SignalR. For authentication, I'm using an existing Identity Server. When the initial login authentication occurs, the claim type for the user's name is "name" (the JWT claim type name). When adding AddOpenIdConnect, I set x.TokenValidationParameters.NameClaimType = "name";

This works, and when I call HttpContext.SignInAsync(User), the User.Identity.Name is correct.

However, once a call makes it into the Hub class, Context.User.Identity.Name is null, and I discovered that the NameClaimType is no longer "name", but has reverted to "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name"

I removed the AzureSignalR specific calls in Startup.cs and the NameClaimType is persisted as "name". so it seems to be a bug with the Azure implementation of SignalR. Is there some extra configuration I am missing, or is this really a bug?

vicancy:
I believe this is a bug that `x.TokenValidationParameters.NameClaimType` info is not passed through service to server and when creating the `User` the `NameClaimType` is reset to default https://github.com/Azure/azure-signalr/blob/dev/src/Microsoft.Azure.SignalR/HubHost/ServiceConnectionContext.cs#L53

PR #224, opened 2018-10-08
Add custom roletype and nametype to claims

vicancy:
Fix https://github.com/Azure/azure-signalr/issues/223


src/Microsoft.Azure.SignalR.Common/ClaimsUtility.cs
```diff
@@ -0,0 +1,120 @@
...
+            var authenticationType = user?.Identity?.AuthenticationType;
+
+            // No need to pass it when the authentication type is Bearer
+            if (authenticationType != null && authenticationType != DefaultAuthenticationType)
+            {
+                yield return new Claim(Constants.ClaimType.AuthenticationType, authenticationType);
+            }
+
+            // return custom NameClaimType and RoleClaimType
+            if (user?.Identity is ClaimsIdentity identity)
```
davidfowl: I think you can have multiple identities but I'm OK ignoring this until we find that it's required.

vicancy: Add a comment for this

Issue #222, opened 2018-10-03
The SDK having issue to reach out to Azure SignalR - The connection is not active, data cannot be sent to the service.


gynet:
I am having a signalR service to be used send message to a group of clients. So basically, I was testing with only one client connecting to the server overnight,. 

The server instance running overnight, in the morning, I try to push a message to a group client again, it failed and I saw this issue  on the sever side, it's having this error multiply times. 

--------------------------------------------------------------------------------------
fail: Microsoft.Azure.SignalR.ServiceConnection[18]
      Service timeout. 30000.00ms elapsed without receiving a message from service.
fail: Microsoft.Azure.SignalR.ServiceConnection[18]
      Service timeout. 30000.00ms elapsed without receiving a message from service.
fail: Microsoft.Azure.SignalR.ServiceConnection[18]
      Service timeout. 30000.00ms elapsed without receiving a message from service.
fail: Microsoft.Azure.SignalR.ServiceConnection[18]
      Service timeout. 30000.00ms elapsed without receiving a message from service.
fail: Microsoft.Azure.SignalR.ServiceConnection[18]
      Service timeout. 30000.00ms elapsed without receiving a message from service.
fail: Microsoft.Azure.SignalR.ServiceConnection[18]
      Service timeout. 30000.00ms elapsed without receiving a message from service.
fail: Microsoft.Azure.SignalR.ServiceConnection[18]
      Service timeout. 30000.00ms elapsed without receiving a message from service.
fail: Microsoft.Azure.SignalR.ServiceConnection[18]
      Service timeout. 30000.00ms elapsed without receiving a message from service.
fail: Microsoft.Azure.SignalR.ServiceConnection[18]
      Service timeout. 30000.00ms elapsed without receiving a message from service.
fail: Microsoft.Azure.SignalR.ServiceConnection[18]
      Service timeout. 30000.00ms elapsed without receiving a message from service.
fail: Microsoft.Azure.SignalR.ServiceConnection[18]
      Service timeout. 30000.00ms elapsed without receiving a message from service.
fail: Microsoft.Azure.SignalR.ServiceConnection[18]
      Service timeout. 30000.00ms elapsed without receiving a message from service.
fail: Microsoft.Azure.SignalR.ServiceConnection[18]
      Service timeout. 30000.00ms elapsed without receiving a message from service.
fail: Microsoft.Azure.SignalR.ServiceConnection[18]
      Service timeout. 30000.00ms elapsed without receiving a message from service.
[...]

PR #227, opened 2018-10-11
Add warning when heartbeat to service takes longer than expected

vicancy:
Add more hints when thread starvation occurs.

related to https://github.com/Azure/azure-signalr/issues/222


vicancy:
This PR seems not a perfect way mitigating the issue, close this one, and continue investigating the issue when a memory dump from customer is available.

PR #236, opened 2018-10-19
Add ServerCloseMessage to protocol

zackliu:
* As customer can't get any error messages when service terminate the server connections now, it will be very confusing when service reach message limit. Under that circumstance, server will keep reconnecting without any useful reasons.
* ServerCloseMessage is used to send back error messages from service before service abort the connection. There are some changes in service to enable the workflow and in SDK, we just log these error messages.
* This is not a breaking change, old sdk can work with new service and vice versa.

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/ServiceConnectionBase.cs
+++ b/src/Microsoft.Azure.SignalR.Common/ServiceConnectionBase.cs
@@ -41,7 +41,7 @@ namespace Microsoft.Azure.SignalR
         private bool _isStopped;
         private long _lastReceiveTimestamp;
         protected ConnectionContext _connection;
-        protected string _errorMessage;
+        protected string ErrorMessage;
 
         public Task WaitForConnectionStart => _serviceConnectionStartTcs.Task;
 
@@ -86,10 +86,10 @@ namespace Microsoft.Azure.SignalR
             // The lock is per serviceConnection
             await _serviceConnectionLock.WaitAsync();
 
-            if (!string.IsNullOrEmpty(_errorMessage))
+            if (!string.IsNullOrEmpty(ErrorMessage))
             {
                 _serviceConnectionLock.Release();
-                throw new InvalidOperationException(_errorMessage);
+                throw new InvalidOperationException(ErrorMessage);
             }
 
             if (_connection == null)
@@ -131,10 +131,10 @@ namespace Microsoft.Azure.SignalR
             if (!string.IsNullOrEmpty(serviceErrorMessage.ErrorMessage))
             {
                 // When receives service error message, we suppose server -> service connection doesn't work,
-                // and set _errorMessage to prevent sending message from server to service
+                // and set ErrorMessage to prevent sending message from server to service
                 // But messages in the pipe from service -> server should be processed as usual. Just log without
                 // throw exception here.
-                _errorMessage = serviceErrorMessage.ErrorMessage;
+                ErrorMessage = serviceErrorMessage.ErrorMessage;
                 Log.ReceivedServiceErrorMessage(_logger, serviceErrorMessage.ErrorMessage);
             }
 
@@ -152,7 +152,7 @@ namespace Microsoft.Azure.SignalR
                 try
                 {
                     _connection = await CreateConnection();
-                    _errorMessage = null;
+                    ErrorMessage = null;
 
                     if (await HandshakeAsync())
                     {
--- a/src/Microsoft.Azure.SignalR.Common/ServiceConnectionBase.cs
+++ b/src/Microsoft.Azure.SignalR.Common/ServiceConnectionBase.cs
@@ -119,6 +119,16 @@ namespace Microsoft.Azure.SignalR
 
         protected abstract Task OnMessageAsync(ConnectionDataMessage connectionDataMessage);
 
+        protected Task OnServerCloseAsync(ServerCloseMessage serverCloseMessage)
+        {
+            if (!string.IsNullOrEmpty(serverCloseMessage.CloseMessage))
+            {
+                throw new Exception(serverCloseMessage.CloseMessage);
+            }
+
+            return Task.CompletedTask;
+        }
+
         private async Task<bool> StartAsyncCore()
         {
             // Always try until connected
@@ -343,6 +353,8 @@ namespace Microsoft.Azure.SignalR
                     return OnDisconnectedAsync(closeConnectionMessage);
                 case ConnectionDataMessage connectionDataMessage:
                     return OnMessageAsync(connectionDataMessage);
+                case ServerCloseMessage serverCloseMessage:
+                    return OnServerCloseAsync(serverCloseMessage);
                 case PingMessage _:
                     // ignore ping
                     break;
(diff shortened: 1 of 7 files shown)


src/Microsoft.Azure.SignalR.Protocols/ServiceProtocol.cs
```diff
@@ -527,6 +539,13 @@ private static MultiGroupBroadcastDataMessage CreateMultiGroupBroadcastDataMessa
             return new MultiGroupBroadcastDataMessage(groupList, payloads);
         }
 
+        private static ServerCloseMessage createServerCloseMessage(byte[] input, ref int offset)
```
davidfowl: Fix the casing of this method, it should pascal cased.

test/Microsoft.Azure.SignalR.Protocols.Tests/ServiceProtocolFacts.cs
```diff
@@ -159,6 +159,10 @@ public static IEnumerable<object[]> TestDataNames
                         ["messagepack"] = new byte[] {7, 8, 1, 2, 3, 4, 5, 6}
                     }),
                 binary: "kw6Spmdyb3VwNKZncm91cDWCpGpzb27ECAECAwQFBgcIq21lc3NhZ2VwYWNrxAgHCAECAwQFBg=="),
+            new ProtocolTestData(
+                name: "ServerClose",
+                message: new ServerCloseMessage("Message count reach limit: 100000"),
```
davidfowl: This error message should be: "Maximum message count limit reached: 100000"

davidfowl:
Where's the unit test? Does it stop reconnecting as a result?

zackliu:
No it won't, just log an error message.

davidfowl:
Where does the message show up? In the logs? Can you ever recover from this error or will it keep getting sent until a day has passed?

zackliu:
It will show in the log. We can't recover it in the SDK side, we just log it and reconnect. The reasons of aborting connection are various, some of them can be recovered after reconnecting. For those can't be recovered through reconnecting e.g. message count limit reached, we will handle it in service side and turn to a slow reconnecting path.(#233)

davidfowl:
I'm thinking that it would be good to capture this exception and re-throw it from the calls to https://github.com/Azure/azure-signalr/blob/c823e993aba01774e6f19faf78f216e8b71f8e46/src/Microsoft.Azure.SignalR.Common/ServiceConnectionBase.cs#L82. The user would get the message from the service side of a generic message.

zackliu:
https://github.com/Azure/azure-signalr/blob/c823e993aba01774e6f19faf78f216e8b71f8e46/src/Microsoft.Azure.SignalR.Common/ServiceConnectionBase.cs#L82
This function is to write message to service side. But now we get error message from service and soon the transport connection and application connection will be terminated. We have no way to throw it to user.

davidfowl:
> This function is to write message to service side. But now we get error message from service and soon the transport connection and application connection will be terminated. We have no way to throw it to user.

That's not true.  This is directly called by user code here https://github.com/Azure/azure-signalr/blob/c823e993aba01774e6f19faf78f216e8b71f8e46/src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManager.cs. If we capture the exception and rethrow it, the user will be able to observe it in their code.

zackliu:
I see, but we receive error message in `ProcessIncomingAsync`, how can I throw exceptions in `WriteAsync`. When error happens, user may not call any methods. He just find the service connection abort without calling anything, now we've already handle the reconnecting and then just need to tell him why in logs,

davidfowl:
> He just find the service connection abort without calling anything, now we've already handle the reconnecting and then just need to tell him why in logs,

It dosn't need to be one or the other, I don't see why you wouldn't do both. I'd also probably make this a specific exception type.

zackliu:
I think throwing exception when calling `WriteAsync` has been handled by
https://github.com/Azure/azure-signalr/blob/c823e993aba01774e6f19faf78f216e8b71f8e46/src/Microsoft.Azure.SignalR.Common/ServiceConnectionBase.cs#L88
After we receives ServerCloseMessage, the connection will be terminated soon. And this error should be only the connection scoped error. So when a new connection is created, we can't treat it as a bad connection.

davidfowl:
> After we receives ServerCloseMessage, the connection will be terminated soon. And this error should be only the connection scoped error. So when a new connection is created, we can't treat it as a bad connection.

That's my point though. You can give the user the right error message instead of the generic one "The connection is not active, data cannot be sent to the service." becomes the actual error that came from the service.

vicancy:
I love not throwing *generic one "The connection is not active, data cannot be sent to the service."*!

src/Microsoft.Azure.SignalR.Common/ServiceConnectionBase.cs
```diff
@@ -41,6 +41,7 @@ internal abstract class ServiceConnectionBase : IServiceConnection
         private bool _isStopped;
         private long _lastReceiveTimestamp;
         protected ConnectionContext _connection;
+        protected string _errorMessage;
```
davidfowl: Any reason this isn't private?

zackliu: Inheritance class which override `WriteAsync` can use `_errorMessage`

src/Microsoft.Azure.SignalR.Common/ServiceConnectionBase.cs
```diff
@@ -119,6 +126,17 @@ public async virtual Task WriteAsync(ServiceMessage serviceMessage)
...
 
+        protected Task OnServerCloseAsync(ServerCloseMessage serverCloseMessage)
+        {
+            if (!string.IsNullOrEmpty(serverCloseMessage.ErrorMessage))
+            {
+                _errorMessage = serverCloseMessage.ErrorMessage;
+                throw new Exception(serverCloseMessage.ErrorMessage);
+            }
+
+            return Task.CompletedTask;
```
vicancy: >CompletedTask [](start = 24, length = 13)

shall we close the server connection with this message?

src/Microsoft.Azure.SignalR.Common/ServiceConnectionBase.cs
```diff
@@ -119,6 +126,17 @@ public async virtual Task WriteAsync(ServiceMessage serviceMessage)
 
         protected abstract Task OnMessageAsync(ConnectionDataMessage connectionDataMessage);
 
+        protected Task OnServerCloseAsync(ServerCloseMessage serverCloseMessage)
+        {
+            if (!string.IsNullOrEmpty(serverCloseMessage.ErrorMessage))
+            {
+                _errorMessage = serverCloseMessage.ErrorMessage;
+                throw new Exception(serverCloseMessage.ErrorMessage);
```
vicancy: >Exception [](start = 26, length = 9)

throw a more detailed exception?

src/Microsoft.Azure.SignalR.Common/ServiceConnectionBase.cs
```diff
@@ -119,6 +126,21 @@ public async virtual Task WriteAsync(ServiceMessage serviceMessage)
 
         protected abstract Task OnMessageAsync(ConnectionDataMessage connectionDataMessage);
 
+        protected Task OnServiceErrorAsync(ServiceErrorMessage serviceErrorMessage)
+        {
+            if (!string.IsNullOrEmpty(serviceErrorMessage.ErrorMessage))
+            {
+                // When receives service error message, we suppose server -> service connection doesn't work,
```
zackliu: @davidfowl After discussed offline, we renamed the message name and changed to log rather than throw exceptions in `ProcessIncomingAsync`

PR #240, opened 2018-10-19
Expose Action<ServiceOptions> to MapAzureSignalR

vicancy:


davidfowl:
What things on HubConfiguration would break the Azure SignalR support? Is there anything that would?

vicancy:
`EnableJSONP` is one we don't support.

davidfowl (approved):

PR #268, opened 2018-11-20
Use channel in ASP.NET to async handle incoming messages

JialinXin:
In ASP.NET without pipeline , we used to sequentially handle Connection/Data/Disconnection messages which may stuck ping message send and lead to server connection timeout. 
Add a channel to quickly dispatcher messages, so make sure ping send is first priority for stable connection. A single task will async process the messages in the channel.


src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ServiceConnection.cs
```diff
@@ -92,14 +92,7 @@ protected override Task OnMessageAsync(ConnectionDataMessage connectionDataMessa
...
                     var message = GetString(payload);
-                    if (message == ReconnectMessage)
-                    {
-                        transport.Reconnected?.Invoke();
-                    }
-                    else
-                    {
-                        transport.OnReceived(message);
-                    }
+                    transport.Output.WriteAsync(message);
```
vicancy: >message [](start = 48, length = 7)

write `payload` instead?

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ServiceConnection.cs
```diff
@@ -143,6 +140,33 @@ private Task OnConnectedAsyncCore(OpenConnectionMessage message)
             }
         }
 
+        private async Task ProcessingOutgoingMessage(string connectionId)
```
vicancy: >ProcessingOutgoingMessage [](start = 27, length = 25)

nit: naming: `ProcessOutgoingMessage`

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ServiceConnection.cs
vicancy: `_ = ` ?

---
In reply to: [234902146](https://github.com/Azure/azure-signalr/pull/268#discussion_r234902146) [](ancestors = 234902146)

JialinXin: how about write connectionDataMessage? And log can be moved inside ProcessOutgoingMessage().

src/Microsoft.Azure.SignalR.AspNet/Transports/AzureTransport.cs
```diff
@@ -39,6 +44,10 @@ public AzureTransport(HostContext context, IDependencyResolver resolver)
 
         public string ConnectionId { get; set; }
 
+        public ChannelReader<string> Input { get; set; }
+        public ChannelWriter<string> Output { get; set; }
+
```
vicancy: nit: remove extra blank line below and one in between

src/Microsoft.Azure.SignalR.AspNet/Microsoft.Azure.SignalR.AspNet.csproj
```diff
@@ -18,6 +18,7 @@
     <PackageReference Include="Microsoft.Extensions.Logging.TraceSource" Version="$(MicrosoftExtensionsLoggingTraceSourcePackageVersion)" />
     <PackageReference Include="Microsoft.AspNetCore.Connections.Abstractions" Version="$(MicrosoftAspNetCoreConnectionsAbstractionsPackageVersion)" />
     <PackageReference Include="Microsoft.AspNetCore.Http.Connections.Client" Version="$(MicrosoftAspNetCoreHttpConnectionsClientPackageVersion)" />
+    <PackageReference Include="System.Threading.Channels" Version="$(MicrosoftAspNetSystemChannelsPackageVersion)" />
```
BrennanConroy: "SystemThreadingChannelsPackageVersion"
https://github.com/aspnet/SignalR/blob/d0bcdf3dd92afba858fca2871535faf4594c81de/build/dependencies.props#L74

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ServiceConnection.cs
```diff
@@ -111,15 +102,20 @@ protected override Task OnMessageAsync(ConnectionDataMessage connectionDataMessa
...
         }
 
-        private void PerformDisconnectCore(string connectionId)
+        private async Task PerformDisconnectAsyncCore(string connectionId)
         {
             if (_clientConnections.TryRemove(connectionId, out var transport))
             {
-                transport.OnDisconnected();
+                // Wait pending messages to complete
+                while(await transport.Output.WaitToWriteAsync())
```
BrennanConroy: remove the while and the break.

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ServiceConnection.cs
```diff
@@ -132,17 +128,47 @@ private Task OnConnectedAsyncCore(OpenConnectionMessage message)
...
         }
 
+        private async Task ProcessOutgoingMessagesAsync(string connectionId)
+        {
+            if (_clientConnections.TryGetValue(connectionId, out var transport))
+            {
+                try
+                {
+                    // Check if channel is closed
+                    while (await transport.Input.WaitToReadAsync())
```
BrennanConroy: I recommend following the pattern we use in AspNetCore
https://github.com/aspnet/SignalR/blob/2d4fb0af6fd2ef2e3a059a2a15a56484d8800d35/src/Microsoft.AspNetCore.SignalR.Client.Core/HubConnectionExtensions.StreamAsChannelAsync.cs#L275-L277

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ServiceConnection.cs
```diff
@@ -143,6 +149,73 @@ private Task OnConnectedAsyncCore(OpenConnectionMessage message)
...
+                                    break;
+                                default:
+                                    break;
+                            }
+                        }
+                    }
+                }
+                catch (Exception e)
+                {
+                    Log.SendLoopStopped(_logger, connectionId, e);
```
vwxyzh: If loop stopped, should we stop the channel to notify writer?

JialinXin: PerformDisconnect to clean and create a disconnect message to notify service

---
In reply to: [237335224](https://github.com/Azure/azure-signalr/pull/268#discussion_r237335224) [](ancestors = 237335224)

JialinXin:
>                     PerformDisconnectCore(connection.Key);

in cleanup scenaio, no need to await one by one.

---
In reply to: [440679959](https://github.com/Azure/azure-signalr/pull/268#issuecomment-440679959) [](ancestors = 440679959)

---
Refers to: src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ServiceConnection.cs:62 in 7a73d72. [](commit_id = 7a73d7282530f4e6eb603b67654a04e61fc35c7a, deletion_comment = False)

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ServiceConnection.cs
```diff
@@ -69,54 +72,56 @@ protected override Task CleanupConnections()
 
         protected override Task OnConnectedAsync(OpenConnectionMessage openConnectionMessage)
         {
+            // Create channel per client to async process message
+            var connectionId = openConnectionMessage.ConnectionId;
+            var channel = Channel.CreateUnbounded<ServiceMessage>();
+            channel.Writer.TryWrite(openConnectionMessage);
```
vicancy: >TryWrite [](start = 27, length = 8)

other places is `await WriteAsync` while here uses `TryWrite`, any concerns?

JialinXin: TryWrite() means attempt to read or write an item synchronously. To make sure open connection message is first priority before rest. 

---
In reply to: [237729148](https://github.com/Azure/azure-signalr/pull/268#discussion_r237729148) [](ancestors = 237729148)

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ServiceConnection.cs
```diff
@@ -69,54 +72,56 @@ protected override Task CleanupConnections()
...
-            }
-            
-            return Task.CompletedTask;
         }
 
         private void PerformDisconnectCore(string connectionId)
         {
+            if (_clientChannels.TryRemove(connectionId, out var channel))
+            {
+                channel.Writer.Complete();
```
vicancy: `TryComplete`?

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ServiceConnection.cs
```diff
@@ -67,56 +70,67 @@ protected override Task CleanupConnections()
...
         }
 
-        protected override Task OnConnectedAsync(OpenConnectionMessage openConnectionMessage)
+        protected override async Task OnConnectedAsync(OpenConnectionMessage openConnectionMessage)
         {
-            // Writing from the application to the service
-            _ = OnConnectedAsyncCore(openConnectionMessage);
+            // Create channel per client to async process message
+            var connectionId = openConnectionMessage.ConnectionId;
+            var channel = Channel.CreateUnbounded<ServiceMessage>();
```
davidfowl: I'm not sure we should be using an unbounded channel. We may want to make sure we're not flooding the channels unexpectedly. Use a bounded channel instead of some reasonable size. We can use that as a metric if things are backed up on the server (the ability to look at the queue length)

JialinXin: Shall we make this configurable in ServiceOptions since it may depend on customer's app server? What default size is suitable? And what shall we do if it exceeds? And if there're a lot messages pending waiting for the channel available, will this stuck ping message?

davidfowl: Spoke to @vicancy about this and we're going to open an issue and leave it unbounded for now. We'll need to address this behavior in the .NET Core version of the SDK as well.

>  And if there're a lot messages pending waiting for the channel available, will this stuck ping message?

No it won't be stuck. We could make it drop messages.

JialinXin: OK.

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ServiceConnection.cs
```diff
@@ -18,7 +18,8 @@ namespace Microsoft.Azure.SignalR.AspNet
     internal partial class ServiceConnection : ServiceConnectionBase
     {
         private const string ReconnectMessage = "asrs:reconnect";
-        private readonly ConcurrentDictionary<string, AzureTransport> _clientConnections = new ConcurrentDictionary<string, AzureTransport>(StringComparer.Ordinal);
+        private readonly ConcurrentDictionary<string, IServiceTransport> _clientConnections = new ConcurrentDictionary<string, IServiceTransport>(StringComparer.Ordinal);
+        private readonly ConcurrentDictionary<string, Channel<ServiceMessage>> _clientChannels = new ConcurrentDictionary<string, Channel<ServiceMessage>>();
```
davidfowl: Why do we have 2 dictionaries instead of storing the channel as state on the transport?

JialinXin: Because we put create connection message in channel, too. And transport is not created at that time.

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ServiceConnection.cs
```diff
@@ -67,56 +68,67 @@ protected override Task CleanupConnections()
...
+            var connectionId = openConnectionMessage.ConnectionId;
+            var channel = Channel.CreateUnbounded<ServiceMessage>();
+            try
+            {
+                await channel.Writer.WriteAsync(openConnectionMessage);
+                _clientChannels.TryAdd(connectionId, channel);
 
-            return Task.CompletedTask;
+                // Writing from the application to the service
+                _ = ProcessMessageAsync(connectionId);
```
davidfowl: Consider storing and awaiting this task when we complete the channel.

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ServiceConnection.cs
```diff
@@ -50,47 +49,130 @@ protected override Task DisposeConnection()
...
+        {
+            var connectionId = message.ConnectionId;
+            if (_clientConnections.TryGetValue(connectionId, out var transport))
+            {
+                try
+                {
+                    var clientTransport = _clientConnectionManager.CreateConnection(message, this);
+                    // Transfer channel and update client connection dictionary with real transport
+                    clientTransport.MessageChannel = transport.MessageChannel;
+                    _clientConnections.TryUpdate(connectionId, clientTransport, transport);
```
vicancy: >TryUpdate [](start = 39, length = 9)

this looks tricky

davidfowl: Yes we shouldn't have to do this. The original feedback was to use a single dictionary entry instead of 2. We can just store a holder class with 2 fields, the channel and the transport.

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ServiceConnection.cs
```diff
@@ -50,96 +51,174 @@ protected override Task DisposeConnection()
...
         {
-            // Writing from the application to the service
-            _ = OnConnectedAsyncCore(openConnectionMessage);
+            // Create empty transport with only channel for async processing messages
+            var connectionId = openConnectionMessage.ConnectionId;
+            var clientContext = new ClientContext();
+            try
+            {
+                await clientContext.Output.WriteAsync(openConnectionMessage);
+                _clientConnections.TryAdd(connectionId, clientContext);
```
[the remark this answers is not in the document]
JialinXin: The open connection message won't be processed no more message can be send on this connection. Shall we directly call Add and handle the exception like connection failed?

test/Microsoft.Azure.SignalR.AspNet.Tests/Infrastructure/TestConnectionManager.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            var transport = new TestTransport
+            {
+                ConnectionId = message.ConnectionId
+            };
+            CurrentTransports.TryAdd(message.ConnectionId, transport);
+            return transport;
+        }
+
+        public class TestTransport : IServiceTransport
```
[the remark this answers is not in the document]
JialinXin: OK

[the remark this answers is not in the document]
JialinXin: MessageCount is now required to access in tests for validation. Or add a function in TestConnectionManager to check message count?

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ServiceConnection.cs
```diff
@@ -153,5 +239,53 @@ private static string GetString(ReadOnlySequence<byte> buffer)
...
+
+            public Task ApplicationTask { get; set; }
+
+            public CancellationTokenSource CancellationToken { get; set; }
+
+            public async Task DisposeAsync(bool closeGracefully = true)
+            {
+                try
+                {
+                    Output.Complete();
```
davidfowl: TryComplete

JialinXin: If try fails and won't throw exception, it will be blocked here pending waiting for task complete.

test/Microsoft.Azure.SignalR.AspNet.Tests/Infrastructure/TestTransport.cs
```diff
@@ -0,0 +1,44 @@
+﻿using Microsoft.AspNet.SignalR.Transports;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
```
JialinXin: will add copywrite and remove unused.

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ServiceConnection.cs
davidfowl: This should never throw. When would it?

JialinXin: Offline synced and TryComplete() returns false only when the channel is already completed. Will fix it.

test/Microsoft.Azure.SignalR.AspNet.Tests/ServiceConnectionTests.cs
```diff
@@ -46,16 +48,20 @@ public async Task ServiceConnectionDispatchTest()
...
                     await proxy.WriteMessageAsync(new OpenConnectionMessage(clientConnection, new Claim[0]));
-
                     await proxy.WaitForClientConnectAsync(clientConnection).OrTimeout();
-
-                    // TODO: Check response when integrated with ServiceMessageBus
-                    await proxy.WriteMessageAsync(new ConnectionDataMessage(clientConnection, GetPayload("Hello World")));
-
-                    await proxy.WaitForApplicationMessageAsync(clientConnection).OrTimeout();
+                    
+                    while(count < 1000)
```
vicancy: >( [](start = 25, length = 1)

empty space

test/Microsoft.Azure.SignalR.AspNet.Tests/ServiceConnectionTests.cs
```diff
@@ -46,16 +48,20 @@ public async Task ServiceConnectionDispatchTest()
...
+                    {
+                        await proxy.WriteMessageAsync(new ConnectionDataMessage(clientConnection, GetPayload("Hello World")));
+                        await proxy.WaitForApplicationMessageAsync(clientConnection).OrTimeout();
+                        count++;
+                    }
+                    // Wait 1 second for async channel processing messages.
+                    await Task.Delay(TimeSpan.FromMilliseconds(1000));
```
vicancy: >1000 [](start = 63, length = 4)

might cause random test failure

vicancy: Add a tcs into the test transport and `awit tcs` instead

---
In reply to: [239367027](https://github.com/Azure/azure-signalr/pull/268#discussion_r239367027) [](ancestors = 239367027)

vicancy: Or send a close connection message

---
In reply to: [239367182](https://github.com/Azure/azure-signalr/pull/268#discussion_r239367182) [](ancestors = 239367182,239367027)

JialinXin: Close is smart. Will move validation in the end.

src/Microsoft.Azure.SignalR.AspNet/Infrastructure/ServiceConnection.cs
```diff
@@ -50,96 +52,183 @@ protected override Task DisposeConnection()
...
         {
             try
             {
+                var tasks = new List<Task>(_clientConnections.Count);
                 foreach (var connection in _clientConnections)
                 {
-                    PerformDisconnectCore(connection.Key);
+                    tasks.Add(PerformDisconnectCore(connection.Key, false));
                 }
+                await Task.WhenAll(tasks);
```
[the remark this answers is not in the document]
JialinXin: It's ungraceful close and may take some time to break the Input.WaitToRead(). Suggest parallel wait would save time.

test/Microsoft.Azure.SignalR.AspNet.Tests/ServiceConnectionTests.cs
```diff
@@ -46,17 +48,21 @@ public async Task ServiceConnectionDispatchTest()
...
+                        count++;
+                    }
 
                     await proxy.WriteMessageAsync(new CloseConnectionMessage(clientConnection));
-
                     await proxy.WaitForClientDisconnectAsync(clientConnection).OrTimeout();
+
+                    // Validate in transport for 1000 data messages.
+                    _clientConnectionManager.CurrentTransports.TryGetValue(clientConnection, out var transport);
+                    Assert.Equal(transport.MessageCount, count);
```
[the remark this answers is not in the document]
JialinXin: OK.

PR #288, opened 2018-12-07
Handle close when loop break in ProcessOutgoingMessageAsync

JialinXin:
Fix core not well handle close when loop break in ProcessOutgoingMessagesAsync().

### DIFF
--- a/src/Microsoft.Azure.SignalR/HubHost/ServiceConnection.cs
+++ b/src/Microsoft.Azure.SignalR/HubHost/ServiceConnection.cs
@@ -58,7 +58,7 @@ namespace Microsoft.Azure.SignalR
                 {
                     return;
                 }
-                await Task.WhenAll(_connectionIds.Select(s => PerformDisconnectAsyncCore(s.Key)));
+                await Task.WhenAll(_connectionIds.Select(s => PerformDisconnectAsyncCore(s.Key, false)));
             }
             catch (Exception ex)
             {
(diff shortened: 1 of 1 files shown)


src/Microsoft.Azure.SignalR/HubHost/ServiceConnection.cs
```diff
@@ -102,11 +102,18 @@ private async Task ProcessOutgoingMessagesAsync(ServiceConnectionContext connect
             }
             catch (Exception ex)
             {
+                // The exception means applicaion fail to process input anymore
+                // Turn down application output to avoid new messages and notify client to disconnect
                 Log.SendLoopStopped(_logger, connection.ConnectionId, ex);
+                connection.Application.Output.CancelPendingFlush();
```
vicancy: is it enough to set `connection.AbortOnClose`?

src/Microsoft.Azure.SignalR/HubHost/ServiceConnection.cs
```diff
@@ -102,11 +102,18 @@ private async Task ProcessOutgoingMessagesAsync(ServiceConnectionContext connect
...
+                connection.Application.Output.Complete();
+                await WriteAsync(new CloseConnectionMessage(connection.ConnectionId, ex.Message));
             }
             finally
             {
+                // Complete Input to let WaitOnApplicationTask() clean transport.
                 connection.Application.Input.Complete();
+                RemoveClientConnection(connection.ConnectionId);
```
vicancy: call `PerformDisconnectAsyncCore` instead?

JialinXin: Will it make ApplicationTask and current Task waiting for each other?

src/Microsoft.Azure.SignalR/HubHost/ServiceConnection.cs
JialinXin: AbortOnClose is default true. Do you mean no need to WriteAync(CloseMessage) since WaitOnApplicationTask will do this?

src/Microsoft.Azure.SignalR/HubHost/ServiceConnection.cs
JialinXin: After we complete application Input and Output, the only thing in PerformDisconnectAsyncCore() is to RemoveConnection, so I directly call it in the end.

JialinXin: I forget about in normal case, it will also do RemoveClientConnection() which will break normal disconnect. Will fix this.

PR #292, opened 2018-12-10
Fix UT random failure

JialinXin:
Add 1 second waiting after receive ping because send ping is a separate task which might have some delay to update the _lastSendTimestamp and leads to test case failure.


vwxyzh:
`ServiceConnectionDispatchTest` still failed.

PR #303, opened 2018-12-20
Add limitation check for AccessKey length

JialinXin:
Currently, the limitation is that access token total length should not longer than 4K. Throws exception after the check.


src/Microsoft.Azure.SignalR.Common/Utilities/AuthenticationHelper.cs
```diff
@@ -14,6 +14,9 @@ namespace Microsoft.Azure.SignalR
 {
     internal static class AuthenticationHelper
     {
+        // Current limitaion is in Nigix for 4K
```
scottaddie:
```suggestion
         // Current limitation is in Nginx for 4K
```

vicancy: remove the internal implementation  details from the comment, as we might change that and there is no need to expose that

src/Microsoft.Azure.SignalR.Common/Utilities/AuthenticationHelper.cs
```diff
@@ -34,13 +37,20 @@ public static string GenerateAccessToken(string signingKey, string audience, IEn
...
                 audience: audience,
                 claims: claims,
                 expires: expire,
                 signingKey: signingKey,
                 requestId: requestId
             );
+
+            if (jwtToken.Length > MaxTokenLength)
+            {
+                throw new ArgumentException("AccessToken must not be longer than 4K");
```
vicancy: For server connections, if the token too long, we should make sure exception throws.

For client’ negotiation, consider return “413 Request Entity Too Large” or “414 Request-URI Too Long”

JialinXin: Right. Update to return 413 for client which is more accurate. And tested locally with expected behavior.

src/Microsoft.Azure.SignalR.AspNet/Hubs/ServiceHubDispatcher.cs
```diff
@@ -53,11 +53,19 @@ public override Task ProcessRequest(HostContext context)
...
             var claims = BuildClaims(context);
 
             // Redirect to Service
             var url = _endpoint.GetClientEndpoint();
-            var accessToken = _endpoint.GenerateClientAccessToken(claims);
+            try
+            {
+                accessToken = _endpoint.GenerateClientAccessToken(claims);
+            }
+            catch (ArgumentException)
```
vicancy: this error is quite general, consider a specific exception type?

src/Microsoft.Azure.SignalR.AspNet/Hubs/ServiceHubDispatcher.cs
```diff
@@ -53,11 +53,19 @@ public override Task ProcessRequest(HostContext context)
...
+            {
+                context.Response.StatusCode = 413;
```
vicancy: also add some response message ?

src/Microsoft.Azure.SignalR.Common/Exception/AccessTokenTooLongException.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Microsoft.Azure.SignalR.Common
+{
+    public class AccessTokenTooLongException : Exception
```
vicancy: `[Serializable]` exception, a sample https://github.com/aspnet/SignalR/blob/master/src/Microsoft.AspNetCore.SignalR.Common/HubException.cs#L13. 
As it is `public`, make the naming more specific, e.g. `AzureSignalRAccessTokenTooLongException`?

src/Microsoft.Azure.SignalR/NegotiateHandler.cs
```diff
@@ -36,9 +37,9 @@ public NegotiationResponse Process(HttpContext context, string hubName)
             {
                 accessToken = _endpointProvider.GenerateClientAccessToken(hubName, claims);
             }
-            catch (ArgumentException)
+            catch (AccessTokenTooLongException)
```
vicancy: no need to catch as it simply throws again?

src/Microsoft.Azure.SignalR/ServiceRouteBuilder.cs
```diff
@@ -33,6 +36,9 @@ public ServiceRouteBuilder(RouteBuilder routes)
             _routes = routes;
             _serviceProvider = _routes.ServiceProvider;
             _negotiateHandler = _serviceProvider.GetRequiredService<NegotiateHandler>();
+
+            var loggerFactory = _serviceProvider.GetService<ILoggerFactory>();
+            _logger = loggerFactory.CreateLogger<ServiceRouteBuilder>();
```
vicancy: can be a null loggerFactory? `loggerFactory?.Create... ?? NullLogger...`

src/Microsoft.Azure.SignalR.Common/Exception/AccessTokenTooLongException.cs
JialinXin: Get it. Do we need to add summary comments as well?

---
In reply to: [243783574](https://github.com/Azure/azure-signalr/pull/303#discussion_r243783574) [](ancestors = 243783574)

src/Microsoft.Azure.SignalR/NegotiateHandler.cs
```diff
@@ -8,6 +8,7 @@
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Connections;
 using Microsoft.AspNetCore.SignalR;
+using Microsoft.Azure.SignalR.Common;
 using Microsoft.Extensions.Options;
```
JialinXin: Will revert the change in this file

test/Microsoft.Azure.SignalR.Common.Tests/AuthenticationHelperTest.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            var claims = GenerateClaims(100);
+            var exception = Assert.Throws<AzureSignalRAccessTokenTooLongException>(() => AuthenticationHelper.GenerateAccessToken(SigningKey, Audience, claims, DefaultLifetime, string.Empty));
+
+            Assert.Equal("AccessToken must not be longer than 4K.", exception.Message);
+        }
+
+        private Claim[] GenerateClaims(int count)
+        {
+            var claims = new List<Claim>();
```
vicancy: another way:
```
return Enumerable.Range(0, count).Select(s=>new Claim($"ClaimSubject{s}",$"ClaimValue{count}")).ToArray();
```

src/Microsoft.Azure.SignalR.AspNet/Hubs/ServiceHubDispatcher.cs
```diff
@@ -25,13 +27,15 @@ internal class ServiceHubDispatcher : HubDispatcher
...
+        private readonly ILogger _logger;
 
         private IServiceEndpointProvider _endpoint;
 
-        public ServiceHubDispatcher(HubConfiguration configuration, string appName, ServiceOptions options) : base(configuration)
+        public ServiceHubDispatcher(HubConfiguration configuration, string appName, ServiceOptions options, ILoggerFactory loggerFactory) : base(configuration)
         {
             _appName = appName ?? throw new ArgumentException(nameof(appName));
             _claimsProvider = options?.ClaimsProvider;
+            _logger = loggerFactory.CreateLogger<ServiceHubDispatcher>();
```
vicancy: `loggerFactory?.Create... ??`

Issue #270, opened 2018-11-21
Check if the server connection is healthy before sending message to this connection


vicancy:
Related issues #266

Currently, we don't check if the server connection is in `connected` mode or not before sending messages to this connection:
1. For group related messages, messages are always round-robined to one specific connection (because we want to preserve the order the group actions), so if this connection encounters any network issue, messages are lost in the recovering period
2. For `echo` messages, we always write back to the incoming connection, it is possible if this incoming connection is not in a healthy state.
2. For other messages, we randomly choose one connection to send messages, so there is also chance that this one connection is not in a healthy state.

To resolve this, consider add a `status` property into the service connection, 
1. For group related messages, if the round-robined connection is not healthy, fallback to another connection. It is possible that group messages are **mis-ordered**.
2. For `echo` messages, if the incoming connection is not healthy, fallback to another connection. It is better if we have the ability to fallback to a server connection from the same Service Instance as the incoming connection.
2. For other messages, skip the unhealthy server-connection.

PR #308, opened 2018-12-26
skip not-connected service connections when writing message

vicancy:
Fix https://github.com/Azure/azure-signalr/issues/270
1. Refactor to use ServiceConnectionContainer for both core and aspnet,
2. Change algorithm to skip not-connected service connections
     1. for random pick, retry 10 times until a valid connection is picked, otherwise throw
     2. for partitioned pick, go through all the service connections in a fixed order until a valid connection is picked. It is possible that group messages are not in order however it is better then message lose


src/Microsoft.Azure.SignalR.Common/ServiceConnectionContainer.cs
```diff
@@ -0,0 +1,110 @@
...
+                throw new ArgumentException($"{nameof(count)} must be greater than 0.");
+            }
+
+            _serviceConnections = new List<IServiceConnection>(count);
+            for (int i = 0; i < count; i++)
+            {
+                _serviceConnections.Add(generator());
+            }
+
+            _count = _serviceConnections.Count;
```
JialinXin: >_count = _serviceConnections.Count [](start = 12, length = 34)

_count = count ?

src/Microsoft.Azure.SignalR.Common/Exceptions/ServiceConnectionNotActiveException.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ServiceConnectionNotActiveException"/> class.
+        /// </summary>
+        public ServiceConnectionNotActiveException() : base("The connection is not active, data cannot be sent to the service.")
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="AzureSignalRException"/> class.
```
JialinXin: >AzureSignalRException [](start = 57, length = 21)

ServiceConnectionNotActiveException ?

src/Microsoft.Azure.SignalR.Common/Exceptions/ServiceConnectionNotActiveException.cs
```diff
@@ -0,0 +1,34 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Runtime.Serialization;
+
+namespace Microsoft.Azure.SignalR.Common
+{
+    /// <summary>
+    /// The exception thrown when AccessToken is too long.
```
JialinXin: >The exception thrown when AccessToken is too long. [](start = 8, length = 50)

update description

src/Microsoft.Azure.SignalR.Common/ServiceConnectionContainer.cs
```diff
@@ -0,0 +1,110 @@
...
+            var retry = 0;
+            while (retry < maxRetry)
+            {
+                var index = nextSelector(retry);
+                var connection = _serviceConnections[index];
+                if (connection != null && connection.Status == ServiceConnectionStatus.Connected)
+                {
+                    try
+                    {
+                        // still possible the connection is not valid
```
JialinXin: >// still possible the connection is not valid [](start = 24, length = 45)

Just question, when would Status set not accurate?

vicancy: because the check is not atomic

src/Microsoft.Azure.SignalR.Common/Exceptions/AzureSignalRException.cs
```diff
@@ -0,0 +1,41 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Runtime.Serialization;
+
+namespace Microsoft.Azure.SignalR.Common
+{
+    /// <summary>
+    /// The exception thrown when AccessToken is too long.
```
JialinXin: Update or remove?

vicancy: Oops

src/Microsoft.Azure.SignalR.Common/ServiceConnectionContainer.cs
```diff
@@ -0,0 +1,118 @@
...
+        private Task WriteToRandomAvailableConnection(ServiceMessage sm)
+        {
+            return WriteWithRetry(sm, StaticRandom.Next(-_count, _count));
+        }
+
+        private async Task WriteWithRetry(ServiceMessage sm, int initial)
+        {
+            var retry = 0;
+            var index = (initial & int.MaxValue) % _count;
+            var direction = initial > 0 ? 1 : -1;
```
vwxyzh: `initial > 0?1:_count-1` then we can remove `+ _count` in line 112?

PR #323, opened 2019-01-07
Add http request handler to call rest api

realwanpengli:


src/Microsoft.Azure.SignalR.Management/HttpRequest.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.SignalR.Management
+{
+    internal class HttpRequest
```
vicancy: >HttpRequest [](start = 19, length = 11)

what this class for?

src/Microsoft.Azure.SignalR.Management/HttpRequest.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        private static readonly IHttpClientFactory _clientFactory = null;
+
+        static HttpRequest()
+        {
+            if (_clientFactory == null)
```
vwxyzh: remove if

src/Microsoft.Azure.SignalR.Management/HttpRequest.cs
vicancy: The naming is too general

---
In reply to: [245872032](https://github.com/Azure/azure-signalr/pull/323#discussion_r245872032) [](ancestors = 245872032)

realwanpengli: For call rest api

vicancy: >class [](start = 13, length = 5)

is a static class enough?

src/Microsoft.Azure.SignalR.Management/HttpRequest.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+using Newtonsoft.Json;
+
+namespace Microsoft.Azure.SignalR.Management
+{
+    internal class HttpRequest
+    {
+        private static readonly IHttpClientFactory _clientFactory = null;
```
vicancy: > = null [](start = 65, length = 7)

can remove

src/Microsoft.Azure.SignalR.Management/HttpRequest.cs
realwanpengli: Yes.

src/Microsoft.Azure.SignalR.Management/HttpRequest.cs
realwanpengli: OK

src/Microsoft.Azure.SignalR.Management/HttpRequest.cs
realwanpengli: > The naming is too general
> 
> In reply to: [245872032](https://github.com/Azure/azure-signalr/pull/323#discussion_r245872032) [](ancestors = 245872032)

How about `RestApiRequest`

vicancy: `...Helper` or `...Utility`?

realwanpengli: OK. sounds `..Helper` / `..Utility` is more usual

test/Microsoft.Azure.SignalR.Management.Tests/HttpRequestFacts.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Net.Http;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Microsoft.Azure.SignalR.Management.Tests
+{
+    public class HttpRequestFacts
+    {
+        [Fact]
+        public async Task RequestAndResponse()
```
vicancy: >RequestAndResponse [](start = 26, length = 18)

Fact naming convention: `{Action}{State}{ExpectedResult}`.
Also, for UT we use mock httpclient instead of rely on network. This case fails when the computer is offline.

realwanpengli: Remove this test, since it doesn't test code in SDK.
Thanks for comments for naming.

PR #341, opened 2019-01-15
Enable multiple endpoint from configs

vicancy:
1. For Core One, support setting multiple endpoints from `IConfiguration` with key starts with
`Azure:SignalR:ConnectionString:` or `ConnectionStrings:Azure:SignalR:ConnectionString:`
2. For AspNet one, support setting multiple endpoints from `config` file, from `ConnectionStrings` section, and `AppSettings` section as fallback, with key starts with `Azure:SignalR:ConnectionString:`


src/Microsoft.Azure.SignalR.AspNet/ServiceOptions.cs
```diff
@@ -36,10 +37,75 @@ public class ServiceOptions
...
+            if (string.IsNullOrEmpty(connectionString))
+            {
+                connectionString = endpoints.FirstOrDefault()?.ConnectionString;
+            }
+
+            ConnectionString = connectionString;
+            Endpoints = endpoints.ToArray();
+        }
+
+        private static (bool, ServiceEndpoint) GetEndpoint(string key, Func<string, string> valueGetter)
```
vwxyzh: add name for value tuple.

src/Microsoft.Azure.SignalR.AspNet/ServiceOptions.cs
```diff
@@ -36,10 +37,75 @@ public class ServiceOptions
...
+
+        public ServiceOptions()
+        {
+            var count = ConfigurationManager.ConnectionStrings.Count;
+            string connectionString = null;
+            var endpoints = new List<ServiceEndpoint>();
+            for (int i = 0; i < count; i++)
+            {
+                var setting = ConfigurationManager.ConnectionStrings[i];
+                var se = GetEndpoint(setting.Name, () => setting.ConnectionString);
```
vwxyzh: `var (endpoint , isDefault) = ...`

PR #344, opened 2019-01-17
Add fallback to GetPrimaryEndpoints when all primary endpoints are offline

vicancy:
When `connection` is null, consider it as `online` to be compatible with current behavior. 

TODO:
- [ ] 1. Add a mechanism for `/negotiate` to wait or throw when the connection is initializing
- [ ] 2. Add UT to cover `/negotiate`

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpointManagerBase.cs
+++ b/src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpointManagerBase.cs
@@ -55,9 +55,9 @@ namespace Microsoft.Azure.SignalR
 
         public abstract IServiceEndpointProvider GetEndpointProvider(ServiceEndpoint endpoint);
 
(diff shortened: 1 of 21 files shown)


src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpointManagerBase.cs
```diff
@@ -58,18 +55,36 @@ internal ServiceEndpointManagerBase(ServiceEndpoint[] endpoints, ILogger logger)
 
         public abstract IServiceEndpointProvider GetEndpointProvider(ServiceEndpoint endpoint);
 
-        public IReadOnlyList<ServiceEndpoint> GetAvailableEndpoints()
+        public virtual IReadOnlyList<ServiceEndpoint> GetAvailableEndpoints()
         {
-            return Endpoints;
+            return Endpoints.Where(s => s.Online).ToArray();
```
vwxyzh: return `IEnumerable`?

PR #347, opened 2019-01-18
Fix bug and expose multiple endpoint in ServiceOptions

vicancy:


src/Microsoft.Azure.SignalR/ServiceOptions.cs
```diff
@@ -39,8 +39,6 @@ public class ServiceOptions : IServiceEndpointOptions
         /// TODO: expose to customer
```
vwxyzh: remove

PR #349, opened 2019-01-21
Use NegotiateMiddleware instead of hacking into PersistentConnection

vicancy:


src/Microsoft.Azure.SignalR.AspNet/Internals/OwinEnvironmentExtensions.cs
```diff
@@ -0,0 +1,91 @@
...
+            object value;
+            if (environment.TryGetValue(OwinConstants.HostTraceOutputKey, out value))
+            {
+                return value as TextWriter;
+            }
+
+            return null;
+        }
+
+        internal static bool SupportsWebSockets(this IDictionary<string, object> environment)
```
davidfowl: You don't need this.

src/Microsoft.Azure.SignalR.AspNet/OwinExtensions.cs
```diff
@@ -273,5 +277,94 @@ private static IReadOnlyList<string> GetAvailableHubNames(HubConfiguration confi
...
+        /// </summary>
+        /// <param name="builder"></param>
+        /// <param name="applicationName"></param>
+        /// <param name="configuration"></param>
+        /// <param name="options"></param>
+        /// <returns></returns>
+        private static IAppBuilder InitAzureSignalRHost(this IAppBuilder builder, string applicationName, HubConfiguration configuration, ServiceOptions options)
+        {
+            // 1. Init SignalR Host 
+            EnsureValidCulture();
```
davidfowl: This isn't needed anymore since we removed the performance counters

src/Microsoft.Azure.SignalR.AspNet/OwinExtensions.cs
```diff
@@ -273,5 +277,94 @@ private static IReadOnlyList<string> GetAvailableHubNames(HubConfiguration confi
...
+            // the app builder respects the OwinMiddleware base class
+            SignatureConversions.AddConversions(builder);
+
+            var env = builder.Properties;
+
+            // If we don't get a valid instance name then generate a random one
+            string instanceName = env.GetAppInstanceName() ?? Guid.NewGuid().ToString();
+
+            // always use empty ProtectedData
+            var protectedData = new EmptyProtectedData();
```
davidfowl: Do we still need this? We added an override to PersistentConnection to allow overriding this right? https://github.com/SignalR/SignalR/issues/4175

src/Microsoft.Azure.SignalR.AspNet/ClientConnections/ClientConnectionManager.cs
```diff
@@ -108,6 +107,49 @@ internal static string GetContentAndDispose(MemoryStream stream)
...
+                message = null;
+                statusCode = 400;
+
+                var tokens = connectionToken.Split(SplitChars, 2);
+
+                connectionId = tokens[0];
+                string tokenUserName = tokens.Length > 1 ? tokens[1] : String.Empty;
+                string userName = GetUserIdentity(context);
+
+                if (!String.Equals(tokenUserName, userName, StringComparison.OrdinalIgnoreCase))
```
JialinXin: >String.Equals(tokenUserName, userName, StringComparison.OrdinalIgnoreCase) [](start = 21, length = 74)

could it be possible that invalid token returns String.Empty and GetUserIdentity() also fail to get Identity makes it turn to equal?

vicancy: Update to a simpler one

src/Microsoft.Azure.SignalR.AspNet/OwinExtensions.cs
```diff
@@ -192,32 +194,46 @@ private static void RunAzureSignalRCore(IAppBuilder builder, string applicationN
...
+            }
+
+            var resolver = configuration.Resolver ?? throw new ArgumentException("A dependency resolver must be specified.");
 
-            ILoggerFactory logger;
-            var traceManager = configuration.Resolver.Resolve<ITraceManager>();
-            if (traceManager != null)
+            // Ensure we have the conversions for MS.Owin so that
+            // the app builder respects the OwinMiddleware base class
+            SignatureConversions.AddConversions(builder);
```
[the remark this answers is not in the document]
vicancy: Isn't that required for the app builder to use `OwinMiddleware` base class when the app builder is not Microsoft.Owin's AppBuilder?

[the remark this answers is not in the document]
davidfowl: Yes but it's added by default https://github.com/aspnet/AspNetKatana/blob/e2b18ec84ceab7ffa29d80d89429c9988ab40144/src/Microsoft.Owin/Builder/AppBuilder.cs#L42, do you need to explicitly call it here?

PR #351, opened 2019-01-21
Handle Exception for rest api mode

realwanpengli:


realwanpengli:
Discussed with @vwxyzh : 
1. Exceptions inherits `AzureSignalRException`. Users can simply catch it.
2. More detail information in exception message: e.g. 404 Not found: endpoint not found
3. For batch requests (e.g. send to users/groups), use `AggregateException` to store multiple exceptions

@chenkennt @vicancy what do you think

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
```diff
@@ -79,44 +100,85 @@ public override Task SendGroupsAsync(IReadOnlyList<string> groupNames, string me
...
+            }
         }
 
-        public Task UserRemoveFromGroupAsync(string userId, string groupName, CancellationToken cancellationToken = default)
+        public async Task UserRemoveFromGroupAsync(string userId, string groupName, CancellationToken cancellationToken = default)
         {
             var api = _restApiProvider.GetUserGroupManagementEndpoint(userId, groupName);
-            return CallRestApi(api, HttpMethod.Delete, null, null, cancellationToken);
-        }
+            var response = await CallRestApi(api, HttpMethod.Delete, null, null, cancellationToken);
```
vicancy: >CallRestApi [](start = 33, length = 11)

move into try..catch block? and use `if (response.IsSuccessStatusCode)` instead?

realwanpengli: But I need exception thrown by `EnsureSuccessStatusCode`

realwanpengli: Do you mean : 
``` C#
            RestApiEndpoint api = null;
            HttpResponseMessage response = null;

            try
            {
                api = _restApiProvider.GetUserGroupManagementEndpoint(userId, groupName);
                response = await CallRestApi(api, HttpMethod.Delete, null, null, cancellationToken);
                response.EnsureSuccessStatusCode();
            }
            catch(Exception ex)
            {
                if (api != null && response != null)
                {
                    ThrowException(ex, response.StatusCode, api.Audience);
                }
                else
                {
                    throw;
                }
            }
```
?

But I don't get the point. I just want to set the HttpRequestException (if response status is failed) to the inner exception of my exception. I don't care about other exceptions.

src/Microsoft.Azure.SignalR.Management/Exceptions/AzureSignalRBadRequestException.cs
```diff
@@ -0,0 +1,23 @@
...
+using Microsoft.Azure.SignalR.Common;
+
+namespace Microsoft.Azure.SignalR.Management
+{
+    [Serializable]
+    public class AzureSignalRBadRequestException : AzureSignalRException
+    {
+        private const string _message = "Bad Request. Caused by one or more of the following reasons: Invalid hub name, invalid group name, null or empty method name, and invalid message.";
+
+        public AzureSignalRBadRequestException(Exception ex, string requestUri) : base(String.IsNullOrEmpty(requestUri) ? _message : $"{_message} Request Uri: {requestUri}", ex)
```
vwxyzh: order: `string requestUri, Exception innerException`

src/Microsoft.Azure.SignalR.Management/Exceptions/AzureSignalRUnauthorizationException.cs
```diff
@@ -0,0 +1,23 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Runtime.Serialization;
+using Microsoft.Azure.SignalR.Common;
+
+namespace Microsoft.Azure.SignalR.Management
+{
+    [Serializable]
+    public class AzureSignalRUnauthorizationException : AzureSignalRException
```
chenkennt: How is this one different than `UnauthorizedAccessException`

src/Microsoft.Azure.SignalR.Common/Exceptions/AzureSignalRUnableToAccessException.cs
```diff
@@ -0,0 +1,23 @@
...
+using System;
+using System.Runtime.Serialization;
+using Microsoft.Azure.SignalR.Common;
+
+namespace Microsoft.Azure.SignalR.Common
+{
+    [Serializable]
+    public class AzureSignalRUnableToAccessException : AzureSignalRException
+    {
+        private const string ErrorMessage = "Unable to access SignalR service. May Caused by one or more of the following reasons: Incorrect endpoint or DNS error.";
```
vicancy: >C [](start = 83, length = 1)

lowercase

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
```diff
@@ -42,10 +44,11 @@ public override Task RemoveFromGroupAsync(string connectionId, string groupName,
             throw new NotImplementedException();
         }
 
-        public override Task SendAllAsync(string methodName, object[] args, CancellationToken cancellationToken = default)
+        public override async Task SendAllAsync(string methodName, object[] args, CancellationToken cancellationToken = default)
         {
             var api = _restApiProvider.GetBroadcastEndpoint();
-            return CallRestApi(api, HttpMethod.Post, methodName, args, cancellationToken);
+            var request = BuildRequest(api, HttpMethod.Post, methodName, args, cancellationToken);
+            await CallRestApiAsync(request, api?.Audience);
```
vicancy: >await [](start = 12, length = 5)

directly return

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
```diff
@@ -79,44 +83,94 @@ public override Task SendGroupsAsync(IReadOnlyList<string> groupNames, string me
...
         }
 
-        public Task UserRemoveFromGroupAsync(string userId, string groupName, CancellationToken cancellationToken = default)
+        public async Task UserRemoveFromGroupAsync(string userId, string groupName, CancellationToken cancellationToken = default)
         {
             var api = _restApiProvider.GetUserGroupManagementEndpoint(userId, groupName);
-            return CallRestApi(api, HttpMethod.Delete, null, null, cancellationToken);
+            var request = BuildRequest(api, HttpMethod.Delete, null, null, cancellationToken);
+            await CallRestApiAsync(request, api?.Audience);
```
vicancy: >api [](start = 44, length = 3)

pass in `api`?

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
```diff
@@ -79,44 +83,94 @@ public override Task SendGroupsAsync(IReadOnlyList<string> groupNames, string me
...
+                ThrowException(ex, HttpStatusCode.NotFound, audience);
+            }
+
+            try
+            {
+                response.EnsureSuccessStatusCode();
+            }
+            catch (Exception ex)
+            {
+                var detail = await response.Content.ReadAsStringAsync();
```
vicancy: >ReadAsStringAsync [](start = 52, length = 17)

can this throw?

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
```diff
@@ -79,44 +98,129 @@ public override Task SendGroupsAsync(IReadOnlyList<string> groupNames, string me
             throw new NotImplementedException();
         }
 
-        public override Task SendUserAsync(string userId, string methodName, object[] args, CancellationToken cancellationToken = default)
+        public override async Task SendUserAsync(string userId, string methodName, object[] args, CancellationToken cancellationToken = default)
         {
+            if (methodName == null)
+            {
+                throw new ArgumentNullException(nameof(methodName));
+            }
```
vwxyzh: what about empty?

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
```diff
@@ -79,44 +98,129 @@ public override Task SendGroupsAsync(IReadOnlyList<string> groupNames, string me
...
+                    }
+                case HttpStatusCode.Unauthorized:
+                    {
+                        throw new AzureSignalRUnauthorizationException(requestUri, innerException);
+                    }
+                case HttpStatusCode.NotFound:
+                    {
+                        throw new AzureSignalRUnableToAccessException(requestUri, innerException);
+                    }
+                default:
```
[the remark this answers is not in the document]
realwanpengli: I tried, the case null will go to `default` block

src/Microsoft.Azure.SignalR.Common/Exceptions/AzureSignalRRuntimeException.cs
```diff
@@ -0,0 +1,22 @@
...
+
+using System;
+using System.Runtime.Serialization;
+
+namespace Microsoft.Azure.SignalR.Common
+{
+    [Serializable]
+    public class AzureSignalRRuntimeException : AzureSignalRException
+    {
+        public const string ErrorMessage = "Azure SignalR service runtime error.";
```
chenkennt: private?

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
```diff
@@ -92,31 +123,119 @@ public override Task SendUsersAsync(IReadOnlyList<string> userIds, string method
...
+            }
+
+            try
+            {
+                detail = await response.Content.ReadAsStringAsync();
+                response.EnsureSuccessStatusCode();
+            }
+            catch (HttpRequestException ex)
+            {
+                ThrowException(ex, response?.StatusCode, request.RequestUri.ToString(), detail);
```
chenkennt: response cannot be null in this case?

src/Microsoft.Azure.SignalR.Common/Exceptions/AzureSignalRUnableToAccessException.cs
```diff
@@ -0,0 +1,23 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Runtime.Serialization;
+using Microsoft.Azure.SignalR.Common;
+
+namespace Microsoft.Azure.SignalR.Common
+{
+    [Serializable]
+    public class AzureSignalRUnableToAccessException : AzureSignalRException
```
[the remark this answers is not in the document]
realwanpengli: How about `AzureSignalRInaccessibleEndpointException`?

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
```diff
@@ -92,31 +123,119 @@ public override Task SendUsersAsync(IReadOnlyList<string> userIds, string method
...
+        {
+            var httpClient = HttpClientFactory.CreateClient();
+            HttpResponseMessage response = null;
+            string detail = "";
+
+            try
+            {
+                response = await httpClient.SendAsync(request);
+            }
+            catch (Exception ex)
```
chenkennt: don't catch Exception

src/Microsoft.Azure.SignalR.Common/Exceptions/AzureSignalRBadRequestException.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Runtime.Serialization;
+
+namespace Microsoft.Azure.SignalR.Common
+{
+    [Serializable]
+    public class AzureSignalRBadRequestException : AzureSignalRException
```
chenkennt: This also need a better name

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
```diff
@@ -92,31 +123,119 @@ public override Task SendUsersAsync(IReadOnlyList<string> userIds, string method
...
+                        throw new AzureSignalRUnableToAccessException(requestUri, innerException);
+                    }
+                default:
+                    {
+                        throw new AzureSignalRRuntimeException(requestUri, innerException);
+                    }
+            }
+        }
+
+        private static void ThrowException(Exception innerException, string requestUri, string detail = null)
```
chenkennt: This is confusing, what's the difference between the two ThrowExceptions? The name of method should be more self-explain.

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
```diff
@@ -92,31 +123,119 @@ public override Task SendUsersAsync(IReadOnlyList<string> userIds, string method
...
+            try
+            {
+                detail = await response.Content.ReadAsStringAsync();
+                response.EnsureSuccessStatusCode();
+            }
+            catch (HttpRequestException ex)
+            {
+                ThrowExceptionOnResponseFailure(ex, response.StatusCode, request.RequestUri.ToString(), detail);
+            }
+            catch (Exception ex)
```
chenkennt: Do not catch Exception

src/Microsoft.Azure.SignalR.Common/Exceptions/AzureSignalRInaccessibleEndpointException.cs
```diff
@@ -0,0 +1,22 @@
...
+
+using System;
+using System.Runtime.Serialization;
+
+namespace Microsoft.Azure.SignalR.Common
+{
+    [Serializable]
+    public class AzureSignalRInaccessibleEndpointException : AzureSignalRException
+    {
+        private const string ErrorMessage = "Unable to access SignalR service. May caused by one or more of the following reasons: Incorrect endpoint or DNS error.";
```
chenkennt: I'm still thinking the possible reason may not be accurate. The inner exception may be self-explain enough.

realwanpengli: The inner exception exception is `HttpRequestException`, the message is "No such host is known" (in Windows) and "No such device or address " (in Linux). I think the inner exception is not enough and message is different in different platform.

realwanpengli: @chenkennt How about this?

message format: phenomenon + reason in inner exception + recommend action

$"Unable to access SignalR service. {innerExceptionMessage}. Please make sure the endpoint or DNS setting is correct."

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
realwanpengli: `ReadAsStringAsync` might throw `LimitMemoryStream`, `HttpRequestException`, `ArgumentOutOfRangeException`, `InvalidOperationException` ...
Except for `HttpRequestException`, other exceptions are only related to customer's machine. Thus I think catch `HttpRequestException` is enough. What do you think? @chenkennt

chenkennt: Then you don't need to catch those exceptions.

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
```diff
@@ -92,31 +123,115 @@ public override Task SendUsersAsync(IReadOnlyList<string> userIds, string method
...
+                        throw new AzureSignalRInaccessibleEndpointException(requestUri, innerException);
+                    }
+                default:
+                    {
+                        throw new AzureSignalRRuntimeException(requestUri, innerException);
+                    }
+            }
+        }
+
+        private static void ThrowExceptionOnNetworkConnectivityAndOtherIssues(Exception innerException, string requestUri, string detail = null)
```
chenkennt: Then you don't need this.

realwanpengli: done

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
```diff
@@ -92,31 +123,100 @@ public override Task SendUsersAsync(IReadOnlyList<string> userIds, string method
...
+            HttpResponseMessage response = null;
+            string detail = "";
+
+            try
+            {
+                response = await httpClient.SendAsync(request);
+            }
+            catch (HttpRequestException ex)
+            {
+                ThrowExceptionOnResponseFailure(ex, HttpStatusCode.NotFound, request.RequestUri.ToString());
```
chenkennt: I would rather directly throw instead of pass in a NotFound status code.

realwanpengli: @chenkennt updated

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
```diff
@@ -44,8 +47,14 @@ public override Task RemoveFromGroupAsync(string connectionId, string groupName,
 
         public override Task SendAllAsync(string methodName, object[] args, CancellationToken cancellationToken = default)
         {
+            if (String.IsNullOrEmpty(methodName))
+            {
+                throw new ArgumentException(NullOrEmptyStringErrorMessage, nameof(methodName));
```
vicancy: why not throw `ArgumentNullException`?

vicancy: also use `string` instead of `String`

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
```diff
@@ -92,31 +123,100 @@ public override Task SendUsersAsync(IReadOnlyList<string> userIds, string method
...
+        {
+            var payload = httpMethod == HttpMethod.Post ? new PayloadMessage { Target = methodName, Arguments = args } : null;
+            return GenerateHttpRequest(endpoint.Audience, payload, endpoint.Token, HttpMethod.Post);
+        }
+
+        private static async Task CallRestApiAsync(HttpRequestMessage request)
+        {
+            var httpClient = HttpClientFactory.CreateClient();
+            HttpResponseMessage response = null;
+            string detail = "";
```
vicancy: nit: `var` is enough

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
realwanpengli: > why not throw `ArgumentNullException`?
The argument may be "" but not null

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
```diff
@@ -92,31 +123,100 @@ public override Task SendUsersAsync(IReadOnlyList<string> userIds, string method
 
         public Task UserAddToGroupAsync(string userId, string groupName, CancellationToken cancellationToken = default)
         {
+            if (string.IsNullOrEmpty(userId))
```
ccic: The validation of userId and groupName happens more than once. Could you please merge them to a standalone validation method?

public void ValidateUserIdAndGroupName(string userId, string groupName)
{
    xxxxx
}

Then in other place, just call ValidateUserIdAndGroupName(userId, groupName).

PR #352, opened 2019-01-22
Remove EmptyProtectedData as it is not needed anymore

vicancy:


test/Microsoft.Azure.SignalR.AspNet.Tests/ClientConnectionManagerTests.cs
```diff
@@ -22,28 +21,37 @@ public class ClientConnectionManagerTests
...
 
         [Theory]
-        [InlineData("?transport=webSockets")]
-        [InlineData("?connectionToken=invalidOne")]
-        public void TestCreateConnectionWithCustomQueryStringSucceeds(string queryString)
+        [InlineData("?transport=webSockets", "a", true)]
+        [InlineData("?connectionToken=invalidOne", "a", false)]
+        [InlineData("?connectionToken=conn2:user1%32c", "conn1", false)]
+        [InlineData("connectionToken=conn2:user1%32c", "conn3", false)]
+        public void TestCreateConnectionAlwaysUsesConnectionIdInOpenConnectionMessage(string queryString, string expectedConnectionId, bool throws)
```
davidfowl: I would split these into valid and invalid tests

PR #370, opened 2019-01-29
Update service connection lifetime management

zackliu:


src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -124,6 +141,16 @@ internal static TimeSpan GetRetryDelay(int retryCount)
             return TimeSpan.FromSeconds(1 << retryCount) + ReconnectInterval;
         }
 
+        protected void ReplaceFixedConnections(int index, IServiceConnection serviceConnection)
+        {
+            lock (_lock)
+            {
+                var newImmutableConnections = FixedServiceConnections.Select(c => c).ToList();
```
vwxyzh: remove `Select(c => c)`

src/Microsoft.Azure.SignalR.Common/ServiceConnections/StrongServiceConnectionContainer.cs
```diff
@@ -29,6 +33,28 @@ internal class StrongServiceConnectionContainer : ServiceConnectionContainerBase
...
+            }
+
+            int index = 0;
+            while (index < pingMessage.Messages.Length - 1)
+            {
+                if (pingMessage.Messages[index] == PingTargetKey &&
+                    !string.IsNullOrEmpty(pingMessage.Messages[index + 1]))
+                {
+                    var connection = CreateOnDemandServiceConnection();
+                    await StartCoreAsync(connection, pingMessage.Messages[index + 1]);
```
vicancy: >await [](start = 20, length = 5)

`return` is enough?

PR #374, opened 2019-01-31
Fix ASP.NET LeaveGroup hung error

vicancy:


src/Microsoft.Azure.SignalR.AspNet/ServerConnections/ServiceConnectionManager.cs
```diff
@@ -57,10 +57,13 @@ public void Initialize(Func<string, IServiceConnectionContainer> connectionGener
 
                 _appConnection = connectionGenerator(_appName);
 
-                foreach (var hub in _hubs)
+                if (_hubs != null)
```
vwxyzh: what about empty collection?

src/Microsoft.Azure.SignalR.AspNet/ServerConnections/ServiceConnectionManager.cs
```diff
@@ -24,7 +25,8 @@ internal class ServiceConnectionManager : IServiceConnectionManager
 
         public ServiceConnectionManager(string appName, IReadOnlyList<string> hubs)
         {
-            if (hubs.Contains(appName))
+            _hubs = hubs ?? EmptyHubs;
```
vwxyzh: `Array.Empty<string>`?

PR #382, opened 2019-02-12
Wanl/mng api update

realwanpengli:


src/Microsoft.Azure.SignalR.Management/ServiceManager.cs
```diff
@@ -54,9 +54,16 @@ public Task<IServiceHubContext> CreateHubContextAsync(string hubName)
             }
         }
 
-        public string GenerateClientAccessToken(string hubName, IList<Claim> claims, TimeSpan? lifeTime = null)
+        public string GenerateClientAccessToken(string hubName, string userId, IList<Claim> claims, TimeSpan? lifeTime = null)
```
chenkennt: User Id can be optional?

src/Microsoft.Azure.SignalR.Management/IServiceManager.cs
```diff
@@ -12,6 +12,8 @@ public interface IServiceManager
     {
         Task<IServiceHubContext> CreateHubContextAsync(string hubName);
 
-        string GenerateClientAccessToken(string hubName, IList<Claim> claims, TimeSpan? lifeTime = null);
+        string GenerateClientAccessToken(string hubName, string userId, IList<Claim> claims, TimeSpan? lifeTime = null);
+
+        string GenerateClientEndpoint(string hubName);
```
chenkennt: Better to be get

src/Microsoft.Azure.SignalR.Management/ServiceManager.cs
realwanpengli: If the client connection only listens to `broadcast` and `group broadcast`, then Yes. That may be one scenario.

I think both `userId` and `claims` can be optional.

PR #390, opened 2019-02-14
[After 1.0.7] Add protocols for ack-able messages

zackliu:
* Add three new message type: `JoinGroupWithAckMessage`, `LeaveGroupWithAckMessage`, `AckMessage`
* Add `Status` for `AckMessage` to return the ack-able message operation status from service side
* As discussed, remove ack-able send group message because it's high cost. But it will result in the operation sequence like `send to group -> join group` out of order. But sequence like `join group -> send to group` or `leave group -> send to group` still keep in order.


src/Microsoft.Azure.SignalR.Protocols/ServiceMessage.cs
```diff
@@ -12,6 +12,14 @@ public abstract class ServiceMessage
     {
     }
 
+    /// <summary>
+    /// Interface of ack-able message 
+    /// </summary>
+    public interface IAckableMessage
+    {
+        int AckId { get; set; }
```
vwxyzh: should we return some messages from service, e.g. operation success or failed?

src/Microsoft.Azure.SignalR.Protocols/ServiceMessage.cs
```diff
@@ -119,4 +127,31 @@ public ServiceErrorMessage(string errorMessage)
...
+    {
+        /// <summary>
+        /// Gets or sets the ack id.
+        /// </summary>
+        public int AckId { get; set; }
+
+        /// <summary>
+        /// Gets or sets the status code
+        /// </summary>
+        public int Status { get; set; }
```
vwxyzh: Also add a message property?

davidfowl:
I suggest updating the spec in lock step with the code.

test/Microsoft.Azure.SignalR.Protocols.Tests/ServiceMessageEqualityComparer.cs
```diff
@@ -59,6 +59,12 @@ public bool Equals(ServiceMessage x, ServiceMessage y)
                         (MultiGroupBroadcastDataMessage)y);
                 case ServiceErrorMessage serviceErrorMessage:
                     return ServiceErrorMessageEqual(serviceErrorMessage, (ServiceErrorMessage)y);
+                case JoinGroupWithAckMessage joinGroupWithAckMessage:
+                    return JoinGroupWithAckMessageEqual(joinGroupWithAckMessage, (JoinGroupWithAckMessage) y);
```
vicancy: remove extra space

PR #410, opened 2019-02-25
Add on-demand connection specific handshake error

zackliu:


src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -603,6 +612,9 @@ private static class Log
             private static readonly Action<ILogger, string, Exception> _unexpectedExceptionInStop =
                 LoggerMessage.Define<string>(LogLevel.Warning, new EventId(29, "UnexpectedExceptionInStop"), "Connection {ServiceConnectionId} got unexpected exception in StopAsync.");
 
+            private static readonly Action<ILogger, string, Exception> _onDemandConnectionHandshakeError =
+                LoggerMessage.Define<string>(LogLevel.Information, new EventId(30, "OnDemandConnectionHandshakeError"), "Service returned handshake error: {Error}");
```
vicancy: don't use `error` for this one, how about `Close on demand connection: {Error}`?

zackliu: It's duplicated with the context of `Message`. What about just `Service returned handshake response: {Message}`

PR #442, opened 2019-03-19
stop connection when disposing service hub context

realwanpengli:


src/Microsoft.Azure.SignalR.Management/ServiceHubContext.cs
```diff
@@ -28,14 +28,28 @@ public ServiceHubContext(IHubContext<Hub> hubContext, IHubLifetimeManagerForUser
 
         public void Dispose()
         {
+            if (GetServiceConnectionStatus() != ServiceConnectionStatus.Disconnected)
+            {
+                StopConnectionAsync().Wait();
```
davidfowl: No please. Never use .Wait or .Result.

realwanpengli: OK. How about `_ = StopConnectionAsync()`

davidfowl: Nope, if you need to do asynchronous work, then await the result. Remove the IDisposable implementation and add a DisposeAsync that callers can use.

realwanpengli: OK. I will remove the interface `IDisposable` and its implementation and keep `DisposeAsync`.

src/Microsoft.Azure.SignalR.Management/ServiceHubContext.cs
```diff
@@ -26,15 +25,25 @@ public ServiceHubContext(IHubContext<Hub> hubContext, IHubLifetimeManagerForUser
...
             _serviceProvider?.Dispose();
         }
 
-        public Task DisposeAsync()
+        private Task StopConnectionAsync()
         {
-            // todo: stop service connection
-            Dispose();
+            var serviceContainer = _serviceProvider.GetService<IServiceConnectionContainer>();
+            if (serviceContainer == null)
```
realwanpengli: In Transient mode, no service of `IServiceConnectionContainer`

Issue #348, opened 2019-01-18
Use app name as a prefix to hub so that different apps with same hub name can share the same Azure SignalR


vicancy:
Quote from our user:
> I would like to understand what app name does and if I can use it for different environments. 
>
> Eg. In the code below environment would be ‘staging’, ‘prod’, ‘dev’
> map.RunAzureSignalR("DrTalkSignalR" + environment, new HubConfiguration()
>
> Would this work and be fully isolated?

kroymann:
A feature like this would be extremely useful to me.  Our service has multiple different parallel environments (prod, test, staging, etc.).  In the prod environment we use Azure SignalR to handle connection capacity, but none of the other environments have enough concurrent users to warrant paying the additional money for each one to have it's own dedicated Azure SignalR instance.  Given that we can't currently isolate the different environments from each other within a shared Azure SignalR instance, this forces us to only use Azure SignalR in production, which raises the chances of encountering bugs that only repro in production. :(

I envision something simple like this with the AppName getting prepended to each of the HubNames internally:

```
services
    .AddSignalR()
    .AddAzureSignalR(opts =>
    {
        opts.ConnectionString = azureSignalRConnectionString;
        **opts.AppName = $"{appName}_{envName}"**;
    });
```

nmehlei:
I would find this also very useful, for the same reasons as @kroymann already described. According to #192 , it's already possible to re-use the same Azure SignalR Service for different App Services (and thus environments) as long as the hub names are different. With a setter for a prefix (or something similar) this could be improved upon.

steverash:
I needed this, so have done the work and created a pull request.   

#449 

Works in the way that @kroymann describes.

vicancy:
Already supported

PR #449, opened 2019-03-21
Add ApplicationName to options and use it to prefix the HubName

steverash:
Added a HubPrefix property to the options class.  

This allows it to be configured to prefix every HubName with a prefix and therefore allows applications to share the same Azure SignalR service without worrying about multiple applications having the same hub names.   

#348


vicancy:
Hi @steverash ,really appreciate your effort on improving the SDK. For this particular feature, we tend to follow the initial proposal in #348 to leverage the `appName` input parameter considering:
1. `appName` is already a required parameter for ASP.NET support, introducing that concept into the ASP.NET Core one as an optional parameter seems consistent and easy to understand.
2. We'd also like to add an `appName` parameter into REST API endpoint. If using`HubPrefix`, users who directly call the REST API would be aware that the actual hub name is `HubPrefix + HubName`. The `appName` approach sounds comparatively straight-forward.

Please just let us know if you have any concerns about the proposal.

kroymann:
@vicancy I'm not familiar with the `appName` parameter that you are referring to.  Aside from being named differently than `HubPrefix`, how would this behavior differ from what I proposed (and what @steverash then implemented)?  I'm pretty sure the only thing that both he and I care about is the ability to inject an arbitrary string into the ASRS configuration that is then used to distinguish/isolate hub connections that share the same HubName.  This could be the same app running in different environments (prod/test/dev/etc), or it could be multiple different applications that happen to use the same name HubName.

vicancy:
`applicatonName` is a required parameter when using the SDK for ASP.NET SignalR https://github.com/Azure/azure-signalr/blob/a101b4384d7578cd14a9f941ab0c048fd88f7f06/src/Microsoft.Azure.SignalR.AspNet/OwinExtensions.cs#L22

Yeah, from the implementation perspective, it is the same as `HubPrefix`.

steverash:
@vicancy thanks for taking the time to review this.  

Is it just renaming hubPrefix to appName, but still implementing in the same way via the ServiceOptions?   If so, I'm happy to go back through the code and do that?       

I can see the code around the ASP.NET version, but putting in the ServiceOptions seemed like a good fit and in keeping with the ASP.NET Core patterns.  The ASP.NET/Owin code seems to work in a quite different way and does not utilise the ASP.NET Core Options pattern?   

I notice there is a seperate ASP.NET and Core ServiceOptions class that both implement IServiceEndpointOptions.   I can see this would be confusing on the ASP.NET side as there would be 2 places in the Owin startup for AppName.

At the moment even on the ASP.NET/Owin side, which has ApplicationName, I can't see that it actually provides any isolation of the Hubs?   Is ApplicationName a work in progress in that respect?  

Also an important point for me is the ability to keep the signalr client routes the same whilst separating the Azure SignalR usage by the appName (hubPrefix).   

If you can confirm that it is the naming needing changing (HubPrefix -> AppName) and if you'd like me to proceed, I'm happy to update the PR tomorrow.

vicancy:
Thanks @steverash, your points make sense. 

`applicationName` in ASP.NET is for other usage but we feel that it sounds great to promote its usage to also isolate hubs.

How about updating the PR to:
1. Rename `HubPrefix` to `ApplicationName`
2. For ASP.NET, leverage the existing `applicationName`

cc @chenkennt @vwxyzh Any comments?

steverash:
@vicancy - I am happy with that plan.   

On a slight side issue, I've just noticed that the behaviour of ASP.NET and ASP.NET Core ServiceEndpointProvider differ.   On ASP.NET the client endpoints don't include the hub query string parameter, but the ASP.NET Core ones do.    Is this intentional or a bug?

steverash:
I've made the change to update it to ApplicaitonName.

On the ASP.NET side, it was a struggle to reconcile the applicationName in the RunAzureSignalR function parameters and the new one in the options class.   Given the Options class is shared between Core and ASP.NET it cannot be removed from ASP.NET easily.    I've added a check to say that if it is specified in the Options class, it must be the same as the one passed to the RunAzureSignalR function.    Not perfect, but hopefully the best option.

src/Microsoft.Azure.SignalR/ServiceOptionsSetup.cs
```diff
@@ -10,31 +10,53 @@ namespace Microsoft.Azure.SignalR
...
+                (connectionString, endpoints) = GetEndpoint(configuration, Constants.ConnectionStringSecondaryKey, Constants.ConnectionStringSecondaryKeyPrefix, _appName);
             }
 
             _connectionString = connectionString;
             _endpoints = endpoints.ToArray();
         }
 
+        private string GetAppName(IConfiguration configuration)
+        {
+            foreach (var pair in configuration.AsEnumerable())
```
vicancy: simply use `configuration[Constants.ApplicationNameDefaultKey]` and no need to support prefix?

src/Microsoft.Azure.SignalR/EndpointProvider/DefaultServiceEndpointGenerator.cs
```diff
@@ -46,21 +50,27 @@ public string GetClientEndpoint(string hubName, string originalPath, string quer
...
-            InternalGetEndpoint(ServerPath, hubName);
+            InternalGetEndpoint(ServerPath, hubName, ApplicationName);
 
-        private string InternalGetEndpoint(string path, string hubName) =>
-            Port.HasValue ?
-            $"{Endpoint}:{Port}/{path}/?hub={hubName.ToLower()}" :
-            $"{Endpoint}/{path}/?hub={hubName.ToLower()}";
+        private string InternalGetEndpoint(string path, string hubName, string applicationName)
+        {
+            var prefixedHubName = string.IsNullOrEmpty(applicationName) ? hubName.ToLower() : $"{applicationName.ToLower()}_{hubName.ToLower()}";
```
vicancy: extract to a common method for calculate `prefixedHubName`?

src/Microsoft.Azure.SignalR.AspNet/EndpointProvider/ServiceEndpointProvider.cs
```diff
@@ -95,9 +98,10 @@ public string GetClientEndpoint(string hubName = null, string originalPath = nul
 
         public string GetServerEndpoint(string hubName)
         {
+            var prefixedHubName = string.IsNullOrEmpty(_appName) ? hubName.ToLower() : $"{_appName.ToLower()}_{hubName.ToLower()}";
             return _port.HasValue ?
-                $"{_endpoint}:{_port}/{ServerPath}/?hub={hubName.ToLower()}" :
-                $"{_endpoint}/{ServerPath}/?hub={hubName.ToLower()}";
+                $"{_endpoint}:{_port}/{ServerPath}/?hub={prefixedHubName.ToLower()}" :
+                $"{_endpoint}/{ServerPath}/?hub={prefixedHubName.ToLower()}";
```
vicancy: no need to `ToLower` as `prefixedHubName` is already lowered

src/Microsoft.Azure.SignalR.AspNet/ServiceOptions.cs
```diff
@@ -25,6 +25,16 @@ public class ServiceOptions : IServiceEndpointOptions
         /// </summary>
         public int ConnectionCount { get; set; } = 5;
 
+        /// <summary>
+        /// Gets applicationName, which will be used as a prefix to apply to each hub name
+        /// </summary>
+        public string ApplicationName{ get; internal set; }
```
vicancy: How about using EII? For ASP.NET one, `ApplicationName` is not exposed using: 
`string IServiceEndpointOptions.ApplicationName { get; } = "";` 
or 
```cs
internal string ApplicationName { get; set; }
string IServiceEndpointOptions.ApplicationName => ApplicationName;
```

src/Microsoft.Azure.SignalR/ServiceOptionsSetup.cs
steverash: Sorted.

src/Microsoft.Azure.SignalR/EndpointProvider/DefaultServiceEndpointGenerator.cs
steverash: done

src/Microsoft.Azure.SignalR.AspNet/EndpointProvider/ServiceEndpointProvider.cs
steverash: fixed

src/Microsoft.Azure.SignalR.AspNet/ServiceOptions.cs
steverash: done.

steverash:
I've made the changes suggested, and I've also changed the .NET Core project to remove the ApplicationName from the ServiceEndPoint and use the IOptions to supply the ApplicationName in the ServiceEndPointProvider.   I think this is better and means ApplicationName is being provided more directly from the IOptions.

src/Microsoft.Azure.SignalR.AspNet/EndpointProvider/ServiceEndpointProvider.cs
```diff
@@ -57,7 +59,8 @@ public string GenerateServerAccessToken(string hubName, string userId, TimeSpan?
                 };
             }
 
-            var audience = $"{_endpoint}/{ServerPath}/?hub={hubName.ToLower()}";
+            var prefixedHubName = string.IsNullOrEmpty(_appName) ? hubName.ToLower() : $"{_appName.ToLower()}_{hubName.ToLower()}";
```
vicancy: also extract the method here

src/Microsoft.Azure.SignalR.AspNet/OwinExtensions.cs
```diff
@@ -181,6 +181,15 @@ private static void RunAzureSignalRCore(IAppBuilder builder, string applicationN
                 throw new ArgumentException(nameof(applicationName), "Empty application name is not allowed.");
             }
 
+            if (options.UseHubNamePrefix)
```
vicancy: no need to `UseHubNamePrefix`, let's just enable it

src/Microsoft.Azure.SignalR.Management/RestApiProvider.cs
```diff
@@ -10,18 +10,28 @@ internal class RestApiProvider
...
+        public RestApiProvider(string connectionString, string hubName, string appName)
         {
             string accessKey;
             (_baseEndpoint, accessKey, _, _port) = ConnectionStringParser.Parse(connectionString);
             _hubName = hubName;
+            _appName = appName.ToLower();
             _restApiAccessTokenGenerator = new RestApiAccessTokenGenerator(accessKey);
-            _requestPrefix = _port == null ? $"{_baseEndpoint}/api/v1/hubs/{_hubName}" : $"{_baseEndpoint}:{_port}/api/v1/hubs/{_hubName}";
-            _audiencePrefix = $"{_baseEndpoint}/api/v1/hubs/{_hubName}";
+            if (string.IsNullOrEmpty(_appName))
```
vicancy: keep it the similar way as the other places using: `var prefixedHubName = string.IsNullOrEmpty(_appName) ? hubName.ToLower() : $"{_appName.ToLower()}_{hubName.ToLower()}";`?

src/Microsoft.Azure.SignalR/EndpointProvider/ServiceEndpointProvider.cs
```diff
@@ -31,21 +34,31 @@ public ServiceEndpointProvider(ServiceEndpoint endpoint, TimeSpan? ttl = null)
...
+            _serviceOptions = serviceOptions;
+            _appName = appName;
 
             var port = endpoint.Port;
             var version = endpoint.Version;
 
             _generator = new DefaultServiceEndpointGenerator(_endpoint, _accessKey, version, port);
         }
 
+        internal string ApplicationName
```
vicancy: is private field enough?

steverash: In theory IOptions.Value can change over time, so I evaluate each time.

src/Microsoft.Azure.SignalR.Management/RestApiProvider.cs
steverash: done

src/Microsoft.Azure.SignalR.AspNet/EndpointProvider/ServiceEndpointProvider.cs
steverash: done

src/Microsoft.Azure.SignalR.AspNet/OwinExtensions.cs
steverash: OK, I've removed this switch

steverash: I've just recalled why I added this.   I was concerned that if there is someone with multiple app servers using different versions of the library (say during rolling out an update), then some will be connected to prefixed hub names and some not prefixed?

vicancy: Nice catch! Yes, that's a breaking change in deed. I will point that out in release note.

test/Microsoft.Azure.SignalR.Management.Tests/UnitTests/ServiceManagerFacts.cs
```diff
@@ -67,5 +70,14 @@ internal void GenerateClientEndpointTest()
 
             Assert.Equal(_clientEndpoint, clientEndpoint);
         }
+
+        [Fact]
+        internal void GenerateClientEndpointWithAppNameTest()
+        {
+            var manager = new ServiceManager(new ServiceManagerOptions() { ConnectionString = _testConnectionString, ApplicationName = _appName });
```
realwanpengli: Could you apply the `ApplicationName` to all other tests in this file?
Simply add `ApplicationName` in _serviceManagerOptions  (line 23)?

realwanpengli:
Could you add simply add `ApplicationName` [here](https://github.com/Azure/azure-signalr/blob/dev/test/Microsoft.Azure.SignalR.Management.Tests/E2ETest/ServiceHubContextE2EFacts.cs#L207) for E2E test?

realwanpengli:
Apply `ApplicationName` to related tests and fix a bug

TimVinkemeier:
Thank you for your work, I was actually waiting for exactly this feature and have been watching this PR for a while 😃 
@vicancy Any idea when it will be available via NuGet (e.g. as a preview or a nightly myget)?

realwanpengli:
@TimVinkemeier The myget package should be the latest dev branch (we publish a new package when the dev branch changes), you can install you the package in myget. Just click the myget package link in this [section](https://github.com/Azure/azure-signalr#nuget-packages), and follow the installation guide to install it.

PR #470, opened 2019-04-02
[Preview][AspNetCore3.0] Add startup to easy switch between Local/Azure SignalR.

JialinXin:
In AspNetCore3.0, SignalR added endpoint routing support which allows Azure SignalR be able to use same startup entry by implementation late bound version of some methods and hosting startup assembly to identify an external library.

Add a configuration `Azure:SignalR:Enabled` to easy switch local/Azure without code changes

1. Value `true` and assembly assigned to be able to use Azure SignalR of simply call AddSignalR()
2. Value `false` to fallback local SignalR when call AddAzureSignalR()

Add a sample chat room targeting aspnetcore 3.0 for tests.

> Notes. AspNetCore SDK version >= 3.0.100-preview4-011024 is required to run and tests.


src/Microsoft.Azure.SignalR/DependencyInjectionExtensions.cs
```diff
@@ -27,7 +30,19 @@ public static class AzureSignalRDependencyInjectionExtensions
         /// </remarks>
         public static ISignalRServerBuilder AddAzureSignalR(this ISignalRServerBuilder builder)
         {
+
             builder.Services.AddSingleton<IConfigureOptions<ServiceOptions>, ServiceOptionsSetup>();
+
+#if NETCOREAPP3_0
+            var serviceOptions = builder.Services
```
davidfowl: What is this?

src/Microsoft.Azure.SignalR/Startup/NegotiateMatcherPolicy.cs
```diff
@@ -0,0 +1,84 @@
...
+using Microsoft.AspNetCore.Routing.Matching;
+using Microsoft.AspNetCore.SignalR;
+
+namespace Microsoft.Azure.SignalR.Startup
+{
+    internal class NegotiateMatcherPolicy : MatcherPolicy, IEndpointSelectorPolicy
+    {
+        // This caches the replacement endpoints for negotiate so they are not recomputed on every request
+        private readonly ConcurrentDictionary<Endpoint, Endpoint> _negotiateEndpointCache = new ConcurrentDictionary<Endpoint, Endpoint>();
+        private readonly ConcurrentDictionary<string, List<IAuthorizeData>> _hubAuthorizePolicy = new ConcurrentDictionary<string, List<IAuthorizeData>>();
```
davidfowl: Remove this, we no longer need to manually apply auth policies. That *should* be done by the authorization middleware.

src/Microsoft.Azure.SignalR/DependencyInjectionExtensions.cs
JialinXin: Build the configurations, so it could fallback to local SignalR with the flag Azure:SignalR:Enabled = false configuration.

davidfowl: We shouldn't be building the service provider before it's ready. This isn't the right place to do this.

JialinXin: I'd prefer to get the configuration in an early stage before unused dependencies add and build. Would use ConfigurationBuilder() to load appsettings.json directly better?

JialinXin: Will revert this part and split to a separate PR.

src/Microsoft.Azure.SignalR/Utilities/ServiceRouteHelper.cs
```diff
@@ -0,0 +1,93 @@
...
+{
+    internal class ServiceRouteHelper
+    {
+        public static async Task RedirectToService(HttpContext context, string hubName, IList<IAuthorizeData> authorizationData)
+        {
+            var handler = context.RequestServices.GetRequiredService<NegotiateHandler>();
+            var loggerFactory = context.RequestServices.GetService<ILoggerFactory>();
+            var logger = loggerFactory.CreateLogger<ServiceRouteHelper>();
+
+#if !NETCOREAPP3_0
```
davidfowl: You can remove this ifdef. Just check for null or empty authorizationData, it's a bit cleaner that way

src/Microsoft.Azure.SignalR/Utilities/AuthorizeHelper.cs
```diff
@@ -66,5 +67,21 @@ public static async Task<bool> AuthorizeAsync(HttpContext context, IList<IAuthor
             }
             return false;
         }
+
+        public static List<IAuthorizeData> BuildAuthorizePolicy(Type hub)
```
davidfowl: You can also remove the ifdef here.

JialinXin: Users may still call UseAzureSignalR() in Core 3.0 which will build ServiceRouteBuilder and then come here. So better to add a check here to skip it completely under 3.0.

src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
```diff
@@ -58,6 +59,26 @@ public void Start(ConnectionDelegate connectionDelegate)
             _ = _serviceConnectionManager.StartAsync();
         }
 
+#if NETCOREAPP3_0
+        public void Start(ConnectionDelegate connectionDelegate, Endpoint endpoint)
```
davidfowl: I think we should make this logic a bit more generic and instead of passing the endpoint, pass an `Action<HttpContext>`. That will allow for future modifications without coupling this logic to adding an endpoint specifically. That way we can keep the ifdef in the AzureSignalRHostedService and all of this code can be decoupled from Endpoint.

src/Microsoft.Azure.SignalR/Utilities/AuthorizeHelper.cs
```diff
@@ -16,7 +16,7 @@ internal static class AuthorizeHelper
     {
         public static async Task<bool> AuthorizeAsync(HttpContext context, IList<IAuthorizeData> policies)
         {
-            if (policies != null && policies.Count == 0)
+            if (policies == null || policies.Count == 0)
```
davidfowl: Hmm, this change looks suspicious. Why change this now?

src/Microsoft.Azure.SignalR/Startup/ServiceHubDispatcher.cs
```diff
@@ -23,11 +24,21 @@ public ServiceHubDispatcher(IServiceProvider serviceProvider)
         public void Start(Endpoint endpoint, Type hubType, ConnectionDelegate app)
         {
             var type = _serviceDispatcherType.MakeGenericType(hubType);
-            var startMethod = type.GetMethod("Start", new Type[] { typeof(ConnectionDelegate), typeof(Endpoint) });
+            var startMethod = type.GetMethod("Start", new Type[] { typeof(ConnectionDelegate), typeof(Action<HttpContext>) });
+
+            var configureContext = new Action<HttpContext>(c => c.Features.Set<IEndpointFeature>(new EndpointFeature
```
davidfowl: Very nice!

src/Microsoft.Azure.SignalR/Utilities/AuthorizeHelper.cs
JialinXin: in 3.0, users might still call UseAzureSignalR() and will get null policies (https://github.com/Azure/azure-signalr/blob/corestartup/src/Microsoft.Azure.SignalR/ServiceRouteBuilder.cs#L47), so null should be also regarded as success?
Normally IAuthorizeData built from Hub class wouldn't be null.

davidfowl: If you call UseAzureSignalR in 3.0 it should work exactly like it did in 2.x

JialinXin: Right. But `AuthorizationMiddleware` is added in Core 3.0 which to reduce the effect in our SDK. Should it be skipped as here I checks (https://github.com/Azure/azure-signalr/blob/corestartup/src/Microsoft.Azure.SignalR/Utilities/AuthorizeHelper.cs#L73-#L76)? 
Or just let `UseAzureSignalR()` do the authorization build and check in 3.0?

davidfowl: I think it should just do what it did before. It means there’ll be a slight performance hit but it’ll make upgrades work. The worst thing we could do is remove the check then have apps upgrade and result in less secure code by default

JialinXin: Make sense. Will update.

samples/ChatSample/ChatSample.CoreApp3/Program.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            CreateHostBuilder(args).Build().Run();
+        }
+
+        public static IWebHostBuilder CreateHostBuilder(string[] args) =>
+            WebHost.CreateDefaultBuilder(args).UseStartup<Startup>();
```
davidfowl: This should reflect the 3.0 template. It uses Host instead of WebHost:

```c#
using System;
using System.Linq;
using System.Threading.Tasks;
using Microsoft.AspNetCore.Hosting;
using Microsoft.Extensions.Configuration;
using Microsoft.Extensions.Hosting;

namespace WebApplication174
{
    public class Program
    {
        public static void Main(string[] args)
        {
            CreateHostBuilder(args).Build().Run();
        }

        public static IHostBuilder CreateHostBuilder(string[] args) =>
            Host.CreateDefaultBuilder(args)
                .ConfigureWebHostDefaults(webBuilder =>
                {
                    webBuilder.UseStartup<Startup>();
                });
    }
}
```

src/Microsoft.Azure.SignalR/ClientConnections/ClientConnectionFactory.cs
```diff
@@ -1,15 +1,17 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using Microsoft.AspNetCore.Http;
 using Microsoft.Azure.SignalR.Protocol;
+using System;
```
davidfowl: nit: System usings should be first

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
```diff
@@ -146,7 +148,8 @@ private void RemoveClientConnection(string connectionId)
 
         protected override Task OnConnectedAsync(OpenConnectionMessage message)
         {
-            var connection = _clientConnectionFactory.CreateConnection(message);
```
davidfowl: Why this change?

src/Microsoft.Azure.SignalR/Startup/ServiceHubDispatcher.cs
```diff
@@ -0,0 +1,45 @@
...
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Features;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Azure.SignalR.Startup
+{
+    internal class ServiceHubDispatcher
+    {
+        // A late bound version of ServiceHubDispatcher<THub>.
+        private readonly Type _serviceDispatcherType = typeof(ServiceOptions).Assembly.GetType("Microsoft.Azure.SignalR.ServiceHubDispatcher`1");
```
davidfowl: Sine we're inside of the same assembly, this could be a bit cleaner: 

```C#
private readonly Type _serviceDispatcherType = typeof(ServiceHubDispatcher<>);
```

src/Microsoft.Azure.SignalR/Startup/ServiceHubDispatcher.cs
davidfowl: Style nit:

```C#
Action<HttpContext> configureContext = c => c.Features.Set<IEndpointFeature>(new EndpointFeature
{
    Endpoint = endpoint
});
```

src/Microsoft.Azure.SignalR/Utilities/AuthorizeHelper.cs
```diff
@@ -66,5 +67,16 @@ public static async Task<bool> AuthorizeAsync(HttpContext context, IList<IAuthor
             }
             return false;
         }
+
+        public static List<IAuthorizeData> BuildAuthorizePolicy(Type hub)
+        {
+            var authorizeAttributes = hub.GetCustomAttributes(typeof(AuthorizeAttribute), inherit: true);
```
davidfowl: nit: There's a generic overload of GetCustomAttributes.

```C#
var authorizeAttributes = hub.GetCustomAttributes<AuthorizeAttribute>(inherit: true);
```

src/Microsoft.Azure.SignalR/Startup/AzureSignalRHostingStartup.cs
```diff
@@ -0,0 +1,29 @@
...
+#if NETCOREAPP3_0
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.Azure.SignalR.Startup;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+[assembly: HostingStartup(typeof(AzureSignalRHostingStartup))]
+
+namespace Microsoft.Azure.SignalR.Startup
```
davidfowl: Put this in the `Microsoft.Azure.SignalR` namespace and get rid of the .Startup.

src/Microsoft.Azure.SignalR/Startup/NegotiateMatcherPolicy.cs
```diff
@@ -0,0 +1,75 @@
...
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Connections;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.AspNetCore.Routing.Matching;
+using Microsoft.AspNetCore.SignalR;
+
+namespace Microsoft.Azure.SignalR.Startup
```
davidfowl: Fix the namespace (similar to the above comment)

src/Microsoft.Azure.SignalR/Startup/ServiceHubDispatcher.cs
```diff
@@ -0,0 +1,45 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#if NETCOREAPP3_0
+using System;
+using Microsoft.AspNetCore.Connections;
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Http.Features;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Azure.SignalR.Startup
```
davidfowl: Same here, fix the namespace.

src/Microsoft.Azure.SignalR/Startup/AzureSignalRConnectionBuilderExtensions.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Reflection;
+using Microsoft.AspNetCore.Connections;
+using Microsoft.AspNetCore.SignalR;
+
+namespace Microsoft.Azure.SignalR.Startup
```
davidfowl: Fix the namespace.

src/Microsoft.Azure.SignalR/Startup/AzureSignalRHostedService.cs
```diff
@@ -0,0 +1,50 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#if NETCOREAPP3_0
+using System;
+using Microsoft.AspNetCore.Connections;
+using Microsoft.AspNetCore.Http.Connections;
+using Microsoft.AspNetCore.Routing;
+using Microsoft.AspNetCore.SignalR;
+
+namespace Microsoft.Azure.SignalR.Startup
```
davidfowl: Same here, namespace

src/Microsoft.Azure.SignalR/Startup/NegotiateMatcherPolicy.cs
```diff
@@ -0,0 +1,75 @@
...
+            return Task.CompletedTask;
+        }
+
+        private Endpoint CreateNegotiateEndpoint(Endpoint endpoint)
+        {
+            var routeEndpoint = (RouteEndpoint)endpoint;
+            var hubMetadata = endpoint.Metadata.GetMetadata<HubMetadata>();
+
+            // Replaces the negotiate endpoint with one that does the service redirect
+            var routeEndpointBuilder = new RouteEndpointBuilder(async context =>
```
BrennanConroy: We should be setting the `DisplayName` for this endpoint.

src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManager.cs
```diff
@@ -27,10 +27,13 @@ internal class ServiceLifetimeManager<THub> : ServiceLifetimeManagerBase<THub> w
             ILogger<ServiceLifetimeManager<THub>> logger, AzureSignalRMarkerService marker)
             : base(serviceConnectionManager, protocolResolver)
         {
+            // after core 3.0 UseAzureSignalR() is not required.
+#if !NETCOREAPP3_0
```
BrennanConroy: I'd recommend changing this so there is a compile error in future TFM changes

https://github.com/aspnet/AspNetCore/blob/648bdf706e823bab9f244f3a06d378de6961f221/src/SignalR/clients/csharp/Http.Connections.Client/src/Internal/WebSocketsTransport.cs#L234

src/Microsoft.Azure.SignalR/Startup/NegotiateMatcherPolicy.cs
JialinXin: Thank you @BrennanConroy for helping review. Is there any naming convention we should follow ? Is **RouteAzureSignalRService** good?

BrennanConroy: The norm seems to be the route path, SignalR sets it to `"/<hub path>/negotiate"` and `"/<hub path>"` for the two endpoints

PR #472, opened 2019-04-03
adjusting the connection retry back off delay

KKhurin:
The current logic (1, 2, 4, 8, 16, 32, 60 seconds) works well when there is only 1 connection. However when there are many connections to the service (and all of them try and fail to connect due to an intermittent network failure or a service being unavailable), then each attempt from each connection will increase the back off time for the next connection. If say 10 connections try and immediately fail, the first 6 will already reach 60s back off delay so the other 4 will only try again after 60 seconds. So if the network was unavailable for 5 seconds only 1 connection will reconnect after 8 seconds, another one will do in 16 seconds, etc. It will take over 60 seconds to get everything reconnected.  The intended logic however is to gradually increase the back off time for each connection. This change restores the intended logic when there are multiple connections.


KKhurin:
My only concern is if FixedConnectionCount can ever be equal to 0. On one hand if it is 0 then nothing will work. On the other hand nothing actually prevents it from being set to 0 in config. So maybe add some kind of config validation?

ccic:
Open question: Is it valid if connection count is 0? Even server connection is 0, app server can play a role of "router" which routes client connections to ASRS.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -113,7 +113,9 @@ protected virtual async Task OnConnectionComplete(IServiceConnection serviceConn
 
         private async Task RestartServiceConnectionCoreAsync(int index)
         {
-            await Task.Delay(GetRetryDelay(_defaultConnectionRetry));
+            // multiple concurrent connection attempts can quickly reach long back off delays
+            // so we adjust the effective retry count according to the number of connections
+            await Task.Delay(GetRetryDelay(_defaultConnectionRetry/FixedConnectionCount));
```
chenkennt: I think after a recent change connection count does not always equal to FixedConnectionCount

chenkennt: Also format the code

vwxyzh: @chenkennt On demand connections won't reconnect. This fix is OK for normal case.
@KKhurin I think it is better to connect server connections one by one, defend for a large `FixedConnectionCount` value (e.g. 6000, as the result we will try to create about 6000 connections per minute, 100 per second, it is still unreasonable reconnect rate).

chenkennt: Make sense. Then what really needs to be fixed is to make this method synchronized?

zackliu: We shall use a queue to implement a producer-consumer model. When a connection failed, it just put itself(index) to the queue and use a seperate task to do the real reconnection and let it control the retry delay.

KKhurin: I see. There are several requirements:
1. Reconnect all dropped fixed connections relatively fast in case of an intermittent short term outage. Currently a 5 seconds outage will result in most of the connections not recovering for a minute. We'd like all the connections to reconnect much faster, ideally with some predictable pattern.

2. Throttle the number of reconnect attempts in case of a large number of configured connections. Like @vwxyzh noted 100 retries for second is unacceptable (and unnecessary). I ran a quick test with the app server configured to have 1500 connections and retrying to connect to the service after a few seconds of outage consumed ~30 seconds of the process CPU time (~6000 retries).

How about this optimization: 
Only one retry happens each 1, 2, 4, 8, 16, 32, 60 seconds but as soon as a connection retry succeeds all other connections will stop waiting, all will try to connect, those which failed will start trying with the back off interval reset to 1s. In other words the exponentially increasing probing is performed with only 1 attempt per back off interval, but the back off interval gets reset as soon as a successful connection is made.

KKhurin:
Here is the summary of new changes:
1.	Try to re-connect immediately when disconnect detected, start back off only after a failure to reconnect. Here are the reasons supporting this change: 
+  a) often there is already some back off time built-in when disconnect happens (ping times out, tcp times out, internal retries failed, etc), 
+  b) we’re connecting to a service behind a load balancer with at least some guarantee of availability so a connection failure does not always mean the whole service is down which makes the initial wait less useful
+  c) this is a self-imposed back off as opposed to a throttling back off imposed by the endpoint (so trying earlier has a good chance to succeed), 
+ d) the algorithm described below allows us to throttle the number of immediate attempts to reconnect and thus no initial back off is required .

2.	Rather than having all disconnected connections trying to re-connect all at once we delegate a “probing” connection which will either succeed, fail, or exceed a time out while trying to connect (e.g. it hit a slow pod). In case of a success we let all other disconnected connections to try to connect. In case of a failure, we back off and initiate another single probe to try to connect. In case of a timeout (which is equal to the current back off time) we initiate a new probing connection attempt. 
This approach makes sure there is no rush of failed connect attempts (e.g. app server network is down). At the same time it promotes restoring several connections in parallel rather than restoring them one by one (in which case one slow or faulty pod may delay a long list of connections waiting to reconnect).

3.	All connections trying to reconnect follow the same back off algorithm. This was the original goal of the fix – make sure we don’t back off too aggressively when recovering from multiple disconnects.

4.	Changes to the tests: adjusted to no back off before the initial failure.

Finally, using multiple instances of BackOffPolicy class allows us to separately control connections with multiple services. It also makes it easy to change to use multiple initial probes (if the experiments show that running a single initial connection probe does not scale in all scenarios).

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -25,8 +25,8 @@ internal abstract class ServiceConnectionContainerBase : IServiceConnectionConta
         protected readonly IConnectionFactory ConnectionFactory;
         protected volatile List<IServiceConnection> FixedServiceConnections;
         protected readonly int FixedConnectionCount;
+        protected BackOffPolicy backOffPolicy = new BackOffPolicy();
```
vwxyzh: `private readonly`

src/Microsoft.Azure.SignalR.Common/Utilities/BackOffPolicy.cs
```diff
@@ -0,0 +1,112 @@
...
+                    }
+                    finally
+                    {
+                        Interlocked.Exchange(ref _currentProbeTcs, null);
+                        // indicate the result to others
+                        myTcs.SetResult(probeSuccess);
+                    }
+
+                    // take care of unfinished probe task
+                    if (awaitProbeTask)
```
chenkennt: Should this be inside the lock?

ccic: It is already protected by https://github.com/Azure/azure-signalr/pull/472/files#diff-bc394dd7d4dd9b3f15a06df56aabd424R38, which plays the role of lock.

chenkennt: That's the 'lock' I'm taking about. My question is: in case of probe task timed out, the lock will be released before the probe task finishes, then there will be multiple tasks running at the same time. Is it expected?

ccic: If the probe task does not succeed (failed or timedout), all the disconnected tasks try to get the 'lock'. The one who fetches the lock starts reconnecting. 
This is a little different from our expected: only 1 disconnected task tries, the others are not involved unless it is successful.
But since the probeTask has risk of hang, if only 1 disconnected task try reconnecting, and it hangs, then the whole reconnecting will hang. So, I think current implementation is Ok.
We should add a cancellation token when StartAsync in https://github.com/Azure/azure-signalr/blob/dev/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ConnectionFactory.cs#L48 to eliminate the potential hang risk.

KKhurin: Here is the scenario @chenkennt is likely concerned about:
Connection 1 is chosen as a probe (e.g. enters the 'lock') and the probe hangs. Eventually we time out and release the 'lock' but keep awaiting for the probe's task. At this point another probe can start executing and it will run in parallel with the connection 1 probe. If this is another connection (another index in the array of fixed connections) then this is what we actually want to do: try again without waiting for the slow probe for index 1. The question is: what happens if we initiate another probe for index 1? 

Let's assume this is possible (e.g. someone calls [RestartServiceConnectionCoreAsync](https://github.com/Azure/azure-signalr/blob/8a3c20fca36804290a578bc13c44b95cb5d58df2/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs#L114) for connection 1 while we're awaiting for connection 1 probe). The probe func will create a new connection, replace it in FixedServiceConnections and try to connect. From the service perspective this should not create problems since this is a separate instance. From structural integrity point this is also safe because we already have parallel execution between the retry and un-awaited task from `_ = StartCoreAsync(connection)` call for the same index and the array is protected by the lock. The worst outcome is we "leak" a connection - e.g. if the original probe eventually succeeds it will be unaccounted in the FixedServiceConnections list. In fact this was the reason why it took me so long to create this PR - I made a mistake and instead of awaiting for `awaitProbeTask` task I awaited for a new call to the `probe()` (that is probe became recursive). This worked but it created an excessive number of connections. 

In reality this does not happen because the only place where we probe again is in the finally block of `StartCoreAsync` - that is in the probe func itself.

I agree with @ccic - we should bring the cancellation token (perhaps with some universal timeout). Then the code in `CallProbeWithBackOffAsync` can handle it the same way it handles timeouts.

KKhurin: Another concern is - shall we even try to connect multiple connections in parallel? I believe we should. The shorter the disconnected time is, the fewer messages get lost, the better balancing between app servers.

chenkennt: I understand even there're multiple connections at the same time their rate is still controlled by the backoff policy. Your explanation makes sense.

KKhurin: Sorry I think I misunderstood the original question. If the probe task times out then the `ongoingProbeTcs.Task.Result` is `false` and then whoever was [ awaiting for ongoingProbeTcs.Task](https://github.com/Azure/azure-signalr/blob/8a3c20fca36804290a578bc13c44b95cb5d58df2/src/Microsoft.Azure.SignalR.Common/Utilities/BackOffPolicy.cs#L95) will not call the probe all at once. They will have to select a new shared ongoing probe and await for the new ongoingProbeTcs.Task. It is only in case of a successful probe that everyone else will try to connect at the same time. The thinking is if one has connected successfully then everyone is going to try (network got recovered? service is back?). If the connection attempt failed then we control the rate of retries.

chenkennt: This part I already understand. If the probe succeeds all pending probes will connect at the same time. This sounds like a reasonable design.

chenkennt: But do we want to control the rate even if it's a successful one? Say thousands of connections connecting at the same time, will it create too much pressure?

vicancy:
Hi @KKhurin we are going to release a new version of SDK this week, would you like this change to be in?

src/Microsoft.Azure.SignalR.Common/Utilities/BackOffPolicy.cs
KKhurin: Successfully establishing a new connection is relatively lightweight. It is not CPU intensive (~2ms of CPU time per connection in my tests), not memory intensive (e.g. no large allocations or spikes during handshake, 5000 connections used 600MB), and not network intensive either. 

Dropping connections could be more expensive (due to contentions while throwing exceptions), but in my quick tests it was still pretty lightweight, about the same as connecting: ~10s of total CPU time for dropping 5000 connections (~2-3 seconds on 4 core server on a local network).  So even creating thousands of them at the same time does not create much pressure.

There could be other consequences of creating too many connections fast (spikes in logging, running out of TCP ports, etc). But these are also the same consequences of simply having too many connections anyway (and these are the reason why app servers want to use the service instead of having clients connecting to them directly). Throttling the rate of connections may help in some cases, but it won’t eliminate the fundamental problem of using too many connections. If in some special case the app server has to have this many connections, I would assume it would rather have them all (re)connected as fast as possible or otherwise it may not serve its purpose.

KKhurin: If we ever decide to add some connection rate throttling, the existing code can be adopted to introduce such throttling in 2 small changes:
-	After awaiting for [ongoingProbeTcs.Task](https://github.com/Azure/azure-signalr/pull/472/files#diff-bc394dd7d4dd9b3f15a06df56aabd424R95) one will just continue the loop (instead of calling the probe right away) so its probe func will get queued again along with other callers. This would effectively throttle to 1 concurrent call at a time with the same back off algorithm.
-	Whoever needs a more than 1 concurrent calls can create N instances of BackOffPolicy and round-robin between them. This will allow up to N concurrent connection attempts while maintaining the lock-free nature of the code.

In general case I would assume the number of connections is small and the connection rate throttling isn’t required on app server side.

KKhurin:
@vicancy yes, if there are no outstanding questions.

src/Microsoft.Azure.SignalR.Common/Utilities/BackOffPolicy.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// - probe's task changes state to task.IsCompletedSuccessfully == true and task.Result == true
+        /// - probe's task.Result == false / faulted / cancelled and the delay defined by getRetryDelay has passed
+        /// </returns>
+        public async Task<bool> CallProbeWithBackOffAsync(Func<Task<bool>> probe, Func<int, TimeSpan> getRetryDelay)
+        {
+            bool calledProbeOnce = false;
+            bool probeSuccess = false;
+            do
+            {
+                var myTcs = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
```
davidfowl: Why is this making a new tcs in a loop?

src/Microsoft.Azure.SignalR.Common/Utilities/BackOffPolicy.cs
```diff
@@ -0,0 +1,112 @@
...
+                    bool awaitProbeTask = false;
+                    try
+                    {
+                        Debug.Assert(!calledProbeOnce);
+                        calledProbeOnce = true;
+                        probeTask = probe();
+
+                        using (CancellationTokenSource delayCts = new CancellationTokenSource())
+                        {
+                            var d = getRetryDelay(_currentRetryCount++);
```
davidfowl: variable naming? delay

src/Microsoft.Azure.SignalR.Common/Utilities/BackOffPolicy.cs
```diff
@@ -0,0 +1,112 @@
...
+                            var delayTask = Task.Delay(d, delayCts.Token);
+                            await Task.WhenAny(delayTask, probeTask);
```
davidfowl: Why isn't this looking at which task caused the yield?

KKhurin: Even if it does we'd still need to handle 3 cases (success, timeout, failure) separately so this doesn't provide much benefit in this particular case.

davidfowl: I dunno, looking at the code, it feels like it could be simplified

src/Microsoft.Azure.SignalR.Common/Utilities/BackOffPolicy.cs
KKhurin: Agreed. Get rid of it completely?
~~~
var delayTask = Task.Delay(getRetryDelay(_currentRetryCount++), delayCts.Token);
~~~

src/Microsoft.Azure.SignalR.Common/Utilities/BackOffPolicy.cs
KKhurin: Hmm... There is no reason indeed. I guess I missed it because the cost is very low comparing to the new connection attempt. In the best case it gets allocated only once per call.

src/Microsoft.Azure.SignalR.Common/Utilities/BackOffPolicy.cs
KKhurin: I've seen both styles in .net libraries. The main goal was to keep the logic relatively easy to follow (successful connection, timeout, everything else).

vicancy:
@KKhurin how is it going? ready for release?

KKhurin:
The only outstanding feedback was from David to handle which task finishes first and I tried coding it the other way without much improvement in clarity or simplicity.  So I think it is ready to get in.

Issue #452, opened 2019-03-22
500 on negotiate when deployed to load balanced servers


DavidMCarek:
Currently I am using Microsoft.Azure.SignalR.AspNet.1.0.0-preview1-10330 and when I deploy to load balanced servers I get a 500 in the browser when the /signalr/negotiate endpoint is hit. The really confusing part of this is that it works locally just fine. I tried a setup with 2 iis app pools running the same website and it worked great. I also was able to get this working on the load balanced servers but only once I downgraded the nuget package to 10275. So something has changed that broke deployment to remote load balanced servers.

vicancy:
The changes in 10330 is that it throws early when the app server is not connected to the Azure Service. For 10275, even if the app server is not connected to the Azure Service, the client not throw when /negotiate but throw on /connect. Could you make sure that your app server is connected to the Azure SignalR Service?

DavidMCarek:
I checked the Azure portal and looks like there's no connection to the app server when the site is spun up on the load balanced server. Is there anything in particular that I need to call to make that connection other than app.MapAzureSignalR or do you think its more likely an infrastructure issue since the code seems to work locally?

vicancy:
Does the load balancer allow web-socket connections? Some proxy need special configuration to enable websocket feature.

DavidMCarek:
Since the solution was working on 10275 I assume the websocket connections are working just fine on the server but I'll double check tomorrow.

DavidMCarek:
Just updated to the latest version and I'm still having the same issue. Its working locally but when it gets deployed I get a 500 in my network tab when the request to signalr/negotiate is executed. If I navigate to the url for that request I get the error message "Azure SignalR Service is not connected yet, please try again later.". I also see no connections made to the signalr service on the azure portal when deployed. 10275 still works just fine. This is what my startup.cs looks like.

```
public class Startup
{
    public void Configuration(IAppBuilder app)
    {
        var client = new HttpClient();
        client.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));

        var container = AutofacConfig.Make(client);

        var enableLogging = bool.Parse(ConfigurationManager.AppSettings["Azure.Signalr.EnableLogging"]);
        if(!int.TryParse(ConfigurationManager.AppSettings["Azure.Signalr.PerServerConnectionCount"], out var perServerConnectionCount))
            perServerConnectionCount = 1;

[...]

PR #499, opened 2019-04-25
separate ASP.NET and the core one productInfo

vicancy:
to get rid of assemblybinding issue

To resolve https://github.com/Azure/azure-signalr/issues/452


src/Microsoft.Azure.SignalR.AspNet/ProductInfo.cs
```diff
@@ -0,0 +1,24 @@
...
+        /// For now, disable usage of System.Runtime.InteropServices.RuntimeInformation
+        /// </summary>
+        /// <returns></returns>
+        public static string GetProductInfo()
+        {
+            var assembly = Assembly.GetCallingAssembly();
+            var packageId = assembly.GetName().Name;
+            var version = assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>().InformationalVersion;
+            
+            return $"{packageId}/{version}(Asp.Net SignalR)";
```
JialinXin: >(Asp.Net SignalR) [](start = 42, length = 17)

Can be removed, that packageId already contains the information

PR #501, opened 2019-04-26
Releasing ASP.NET SignalR and remove preview

vicancy:


src/Microsoft.Azure.SignalR.AspNet/Microsoft.Azure.SignalR.AspNet.csproj
```diff
@@ -4,8 +4,6 @@
     <TargetFramework>net461</TargetFramework>
     <RootNamespace>Microsoft.Azure.SignalR.AspNet</RootNamespace>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
-    <!-- Fix AspNet version to preview1 -->
-    <PackageVersion>1.0.0-preview1-$(BuildNumber)</PackageVersion>
```
chenkennt: keep 1.0.0?

PR #508, opened 2019-04-28
add productInfoAttr

realwanpengli:


src/Microsoft.Azure.SignalR.Common/Utilities/ProductInfoAttribute.cs
```diff
@@ -0,0 +1,17 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using System.Text;
+
+namespace Microsoft.Azure.SignalR.Common
+{
+    [AttributeUsage(AttributeTargets.Assembly)]
+    class ProductInfoAttribute : Attribute
```
vicancy: add `internal` explicitly

src/Microsoft.Azure.SignalR/ProductInfo.cs
```diff
@@ -1,22 +1,49 @@
...
 
 namespace Microsoft.Azure.SignalR
 {
     internal static class ProductInfo
     {
+        private const int MaxProductInfoLength = 256;
         public static string GetProductInfo()
         {
-            var assembly = Assembly.GetCallingAssembly();
+            var allAssemblies = AppDomain.CurrentDomain.GetAssemblies();
```
vicancy: Maybe just make it for Management API?

src/Microsoft.Azure.SignalR/ProductInfo.cs
```diff
@@ -1,22 +1,49 @@
...
+
+        public static string GetProductInfoCore(Assembly assembly)
+        {
             var packageId = assembly.GetName().Name;
             var version = assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>().InformationalVersion;
             var runtime = RuntimeInformation.FrameworkDescription?.Trim();
             var operatingSystem = RuntimeInformation.OSDescription?.Trim();
             var processorArchitecture = RuntimeInformation.ProcessArchitecture.ToString().Trim();
-            
             return $"{packageId}/{version} ({runtime}; {operatingSystem}; {processorArchitecture})";
```
JialinXin: >({runtime}; {operatingSystem}; {processorArchitecture}) [](start = 43, length = 55)

For multiple assembilies, the runtime information will be duplicated?

src/Microsoft.Azure.SignalR/ProductInfo.cs
realwanpengli: But management SDK reuse service SDK, so there seems no way to tell which one is quering product info if we don't change api.

realwanpengli:
cons: what if someone use both service sdk and management sdk?

discussed and figure out a better solution: add a method that input a assembly in builder,

PR #513, opened 2019-04-29
product info v3

realwanpengli:


src/Microsoft.Azure.SignalR/ProductInfo.cs
```diff
@@ -8,16 +8,27 @@ namespace Microsoft.Azure.SignalR
 {
     internal static class ProductInfo
     {
+        private const int MaxLength = 128;
```
JialinXin: >128 [](start = 38, length = 3)

128 is too small. Better to use 256.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ConnectionFactory.cs
```diff
@@ -29,7 +29,7 @@ public ConnectionFactory(string hubName, IServiceEndpointProvider provider, ISer
             _hubName = hubName;
         }
 
-        public async Task<ConnectionContext> ConnectAsync(TransferFormat transferFormat, string connectionId, string target, CancellationToken cancellationToken = default, IDictionary<string, string> headers = null)
+        public virtual async Task<ConnectionContext> ConnectAsync(TransferFormat transferFormat, string connectionId, string target, CancellationToken cancellationToken = default, IDictionary<string, string> headers = null)
```
vicancy: [Composition over inheritance](https://en.wikipedia.org/wiki/Composition_over_inheritance)

src/Microsoft.Azure.SignalR.Management/ConnectionFactory.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        public ConnectionFactory(string productInfo, string hubName, IServiceEndpointProvider provider, IServerNameProvider nameProvider, ILoggerFactory loggerFactory)
+            : base(hubName, provider, nameProvider, loggerFactory)
+        {
+            _productInfo = productInfo;
+        }
+
+        public override Task<ConnectionContext> ConnectAsync(TransferFormat transferFormat, string connectionId, string target, CancellationToken cancellationToken = default, IDictionary<string, string> headers = null)
+        {
+            return base.ConnectAsync(transferFormat, connectionId, target, cancellationToken, new Dictionary<string, string> { { Constants.AsrsUserAgent, _productInfo } });
```
vicancy: the input `header` is always ignored?

realwanpengli: Yes, since this one is always used by management sdk, the management provides product info.

src/Microsoft.Azure.SignalR/ProductInfo.cs
```diff
@@ -8,16 +8,27 @@ namespace Microsoft.Azure.SignalR
...
+        {
             var packageId = assembly.GetName().Name;
             var version = assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>().InformationalVersion;
             var runtime = RuntimeInformation.FrameworkDescription?.Trim();
             var operatingSystem = RuntimeInformation.OSDescription?.Trim();
             var processorArchitecture = RuntimeInformation.ProcessArchitecture.ToString().Trim();
-            
-            return $"{packageId}/{version} ({runtime}; {operatingSystem}; {processorArchitecture})";
+            var packageInfo = $"{packageId}/{version}";
+            return $"{TruncateString(packageInfo)} ({runtime}; {operatingSystem}; {processorArchitecture})";
```
vicancy: >TruncateString [](start = 22, length = 14)

which one could probabaly exceed max?

vicancy: append `...` to indicate it is truncated?

---
In reply to: [279599840](https://github.com/Azure/azure-signalr/pull/513#discussion_r279599840) [](ancestors = 279599840)

src/Microsoft.Azure.SignalR/ProductInfo.cs
```diff
@@ -8,16 +8,27 @@ namespace Microsoft.Azure.SignalR
...
     {
+        private const int MaxLength = 128;
+
         public static string GetProductInfo()
         {
-            var assembly = Assembly.GetCallingAssembly();
+            return GetProductInfo(Assembly.GetCallingAssembly());
+        }
+
+        public static string GetProductInfo(Assembly assembly)
```
vicancy: >assembly [](start = 53, length = 8)

support null as default?

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
```diff
@@ -11,6 +13,7 @@ namespace Microsoft.Azure.SignalR.Management
     public class ServiceManagerBuilder : IServiceManagerBuilder
     {
         private readonly ServiceManagerOptions _options = new ServiceManagerOptions();
+        private Assembly _assembly = Assembly.GetExecutingAssembly();
```
vicancy: >GetExecutingAssembly [](start = 46, length = 20)

no need?

realwanpengli: We need it. If `WithCallingAssembly`() isn't called, we need to track the Management SDK itself.
Will move it to `Build()` to avoid calling twice.

PR #520, opened 2019-05-07
Enable server sticky mode and add docs

vicancy:


docs/use-signalr-service.md
```diff
@@ -92,6 +92,14 @@ By default, all claims from `HttpContext.User` of the negotiate request will be
 They can be accessed at [`Hub.Context.User`](https://github.com/aspnet/SignalR/blob/release/2.2/src/Microsoft.AspNetCore.SignalR.Core/HubCallerContext.cs#L29).
 - Normally you should leave this option as is. Make sure you understand what will happen before customizing it.
 
+#### `ServerStickyMode`
+- Default value is `Disabled`.
+- This option specifies the mode for **server sticky**. When the client is always routed to the server which it first `/negotiate` with, we call it **server sticky**. This options can be useful when you want to store some client context when `/negotaite` for `hub` to use.
+- In distributed scenarios, there can be multiple app servers connecting to one Azure SignalR instance. Inside one Azure SignalR instance, there are many running pods, each of them take some server connections from some app server. When the client comes in, Azure SignalR chooses the server connection for the client traffic to route to based on the routing strategy.
```
chenkennt: Since preferred is unpredictable, in which case this will more useful than disabled?

chenkennt: should be preferred rather than prefered.

docs/use-signalr-service.md
```diff
@@ -92,6 +92,16 @@ By default, all claims from `HttpContext.User` of the negotiate request will be
...
+- In distributed scenarios, there can be multiple app servers connected to one Azure SignalR instance. As [internals of client connections](../internal.md#client-connections) explains, client first negotiates with the app server, and then redirects to Azure SignalR to establish the persistent connection. Azure SignalR then finds one app server to serve the client, as [Transport Data between client and server](../internal.md#transport-data-between-client-and-server) explains.
+    - When `Disabled`, the client routes to a random app server. In general, app servers have balanced client connections with this mode. If your scenarios are *broadcast* or *group send*, use this default option is enough.
+    - When `Prefered`, Azure SignalR tries to find the app server which the client first negotiate with in a way that no additional cost or global routing is needed. This one can be useful when your scenario is *send to connection*. *Send to connection* can have better performance and lower latency when the sender and the receiver are routed to the same app server.
```
chenkennt: prefered -> preferred

test/Microsoft.Azure.SignalR.AspNet.Tests/RunAzureSignalRTests.cs
```diff
@@ -584,7 +584,7 @@ public async Task TestStickyServerInServiceOptionsTakeEffect()
                 using (WebApp.Start(ServiceUrl, a => a.RunAzureSignalR(AppName, hubConfiguration, options =>
                 {
                     options.IsolateApplication = true;
-                    options.ServerStickyMode = ServerStickyMode.Prefered;
+                    options.ServerStickyMode = ServerStickyMode.Preferred;
```
chenkennt: Did you change the enum definition?

docs/use-signalr-service.md
```diff
@@ -92,6 +92,16 @@ By default, all claims from `HttpContext.User` of the negotiate request will be
 They can be accessed at [`Hub.Context.User`](https://github.com/aspnet/SignalR/blob/release/2.2/src/Microsoft.AspNetCore.SignalR.Core/HubCallerContext.cs#L29).
 - Normally you should leave this option as is. Make sure you understand what will happen before customizing it.
 
+<a name="server-sticky-mode"></a>
+
+#### `ServerStickyMode` 
+- Default value is `Disabled`.
```
davidfowl: Seems like the default should be preferred no?

docs/use-signalr-service.md
```diff
@@ -92,6 +92,16 @@ By default, all claims from `HttpContext.User` of the negotiate request will be
...
+    - When `Disabled`, the client routes to a random app server. In general, app servers have balanced client connections with this mode. If your scenarios are *broadcast* or *group send*, use this default option is enough.
+    - When `Preferred`, Azure SignalR tries to find the app server which the client first negotiate with in a way that no additional cost or global routing is needed. This one can be useful when your scenario is *send to connection*. *Send to connection* can have better performance and lower latency when the sender and the receiver are routed to the same app server.
+    - When `Required`, Azure SignalR always tries to find the app server which the client first negotiate with. This options can be useful when some client context is fetched from `negotaite` step and stored in memory, and then to be used inside `Hub`s. However, this option may have performance drawbacks because it requires Azure SignalR to take additional efforts to find this particular app server globally, and to keep globally routing traffics between client and server.
```
davidfowl: negotaite -> negotiate

docs/use-signalr-service.md
vicancy: `Preferred` might have side effect that client connections are not balanced to the app servers, it can have extra burden to the negotiate server, not sure if that is what users want, so we keep it `Disabled` as it is the current behavior.

PR #522, opened 2019-05-08
Add always reconnect logic to netcore js client sample

vicancy:
A simple reconnect strategy that it always reconnect when error occurs


samples/ChatSample/ChatSample/wwwroot/index.html
```diff
@@ -135,27 +146,41 @@ <h2 class="text-center" style="margin-top: 0; padding-top: 30px; padding-bottom:
...
+                        }, getRandom(200, 2000));
+            }
+
+            function startConnectionWithRetry(hubConnection) {
+                hubConnection.start()
+                    .then(function () {
+                        onConnected(hubConnection);
+                    })
+                    .catch(error => {
+                        console.error(error);
```
chenkennt: just call onConnectionError?

PR #532, opened 2019-05-16
Fix multi-endpoint bug that per hub per container should have its own…

vicancy:
… endpoint as endpoint contains connectioncontainer info


src/Microsoft.Azure.SignalR.Common/EndpointRouters/DefaultMessageRouter.cs
```diff
@@ -53,30 +51,12 @@ public IEnumerable<ServiceEndpoint> GetEndpointsForGroup(string groupName, IEnum
...
-
-            // All primary endpoints are offline, fallback to the first online secondary endpoint
-            var secondary = endpoints.Where(s => s.Online && s.EndpointType == EndpointType.Secondary).ToArray();
-            if (secondary.Length == 0)
-            {
-                throw new AzureSignalRNotConnectedException();
-            }
-
-            return secondary;
+            return endpoints.Where(e => e.Online);
```
JialinXin: >endpoints.Where(e => e.Online) [](start = 19, length = 30)

would it be 0 and shall we throw an exception?

vicancy: How about not when it is for sending messages?

JialinXin: Suggest to give customer some information so that they would be aware of sending failure is caused by no endpoint online.

vicancy: Keep the original behavior then, fix the issue later.

PR #534, opened 2019-05-17
Adding webproxy from ServiceEndpoint to HttpConnection

tebeco:
## Context :
* The goal of this pull request is to run an application hosting a `Hub` and `AddAzureSignalR` 
* As part of a company that still refuse to use transparent proxy we have to explicitly "point" at the proxy for outgoing request
* This is done under the hood by this Sdk when the application start.

## No goal
Other idea was suggested before this PR but are not really possible :
* Targeting `netfx`
  ==> NO GO "because it has magic code that interact with windows"
  ==> Require to setup the proxy in IE
  ==> Will not work if not on windows
* using `HTTP_PROXY`
  ==> Pretty bad idea when your application communicate mostly in the LAN 
  ==> Especially when the only place you need a `Proxy` is for the connection to `Azure Signalr`
* using `NO_PROXY`
  ==> would be too complexe to filter out internal network
* `netcoreapp3.0` 
  ==> `netcoreapp3.0` add `HTTP_SUPPORT` (see previous point)
  ==> We arleady have a running app on `netcoreapp2.2` waiting for `netcoreapp3.0` would require much more change only for this

## Target
The PR is targeting `master` because it "look like" the current version that is already published on nuget.org and is the one targeting `netstandard2.0` so we could use it while still targetting `netcoreapp2.x`

## A bit of details

If we take a look as the CSharp client for signalr, it seems that the class [`HttpConnectionOptions`](https://github.com/aspnet/AspNetCore/blob/master/src/SignalR/clients/csharp/Http.Connections.Client/src/HttpConnectionOptions.cs) contains code for :
* [`public IWebProxy Proxy { get; set; }`](https://github.com/aspnet/AspNetCore/blob/master/src/SignalR/clients/csharp/Http.Connections.Client/src/HttpConnectionOptions.cs#L119)
* [`public Func<HttpMessageHandler, HttpMessageHandler> HttpMessageHandlerFactory { get; set; }`](https://github.com/aspnet/AspNetCore/blob/master/src/SignalR/clients/csharp/Http.Connections.Client/src/HttpConnectionOptions.cs#L39)
* [`public Action<ClientWebSocketOptions> WebSocketConfiguration { get; set; }`](https://github.com/aspnet/AspNetCore/blob/master/src/SignalR/clients/csharp/Http.Connections.Client/src/HttpConnectionOptions.cs#L104)
  with also contains : `public IWebProxy Proxy { get; set; }`

Now if we take a look at how `Microsoft.Azure.SignalR` connects to the Azure Signalr backplane :
As i'm not really familiar with library, it look like it's [here](https://github.com/Azure/azure-signalr/blob/dev/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ConnectionFactory.cs#L32-L53) :
```csharp
public async Task<ConnectionContext> ConnectAsync(TransferFormat transferFormat, string connectionId, string target, CancellationToken cancellationToken = default, IDictionary<string, string> headers = null)
{
    var httpConnectionOptions = new HttpConnectionOptions
    {
        Url = GetServiceUrl(connectionId, target),
        AccessTokenProvider = () => Task.FromResult(_provider.GenerateServerAccessToken(_hubName, _userId)),
        Transports = HttpTransportType.WebSockets,
        SkipNegotiation = true,
        Headers = headers,
        Proxy = _provider.GetServerProxy()
    };
    var httpConnection = new HttpConnection(httpConnectionOptions, _loggerFactory);
    try
    {
        await httpConnection.StartAsync(transferFormat, cancellationToken);
        return httpConnection;
    }
    catch
    {
        await httpConnection.DisposeAsync();
        throw;
    }
}
```

As we can see the connection re-use the code from the `HttpConnectionOptions` + `HttpConnection` that is already supposed to works 🤞 

In a naive approach it look like the "code flow execution" from "Configure" to "ConnectAsync" look like :
* `AddAzureSignalr(options => options.ServiceEndpoints.xxxxx)`
* `ServiceEndpoint` is injected in `ServiceEndpointProvider`
* `ServiceEndpointProvider` is injected in `ConnectionFactory`
* the `ConnectionFactory` creates on the fly a `HttpConnectionOptions`

This very naive PR just try to add an `IWebProxy` to the only "Object" we can manipulate before the bootstrap => `ServiceEndpoint` and flow the information to the connection

## More ?
As this PR is really naive and i would love feedback on it, as you can see i did not edited a single unit test yet.
This is of course the next step if this PR is still ok


vicancy:
Thanks for the contribution!

The idea looks pretty good to me. Some minor comments towards the PR:
1. Targeting `dev` is enough. We publish 2 packages from `dev` branch, one for `netstandard2.0` and one for `netcore3.0`
2. How about adding the `IWebProxy` to [ServiceOptions](https://github.com/Azure/azure-signalr/blob/dev/src/Microsoft.Azure.SignalR/ServiceOptions.cs) so it can be customized by for example `AddAzureSignalr(options => options.ServerWebProxy)`?

tebeco:
hi @vicancy

1. i'll rebase on `dev` then. My main pre-occupation is to have it before the release of `netcoreapp3.0` of September
  i think it could even be tagged as `blocking-partner` honestly

2. yes of course, i'll take a deeper look at how `ServiceOptions` is consumed and do the change ASAP

3. what design do you want for the test ? simple field checking since it flows through code and copy reference ?

vicancy:
Hi @tebeco:
1. We generally release stable `netstandard20` version every 3-4 weeks. For every changes committed to dev branch, we always publish 2 packages, one `1.1.0-preview...` targeting for `netcoreapp3.0`, one `1.0.x-preview...` targeting for `netstandard20`, the current readme is confusing because it only lists the latest `netcoreapp3.0` version. But don't worry, the `netstandard20` version is always in its pace of releasing stable version to Nuget, current version is `1.0.9`, next one would be `1.0.10`.

2. We have two `ServiceOptions` one for [ASP.NET](https://github.com/Azure/azure-signalr/blob/dev/src/Microsoft.Azure.SignalR.AspNet/ServiceOptions.cs) and one for [ASP.NET Core](https://github.com/Azure/azure-signalr/blob/dev/src/Microsoft.Azure.SignalR/ServiceOptions.cs), I think the change can be applied to both.

3. I don't think current code is decoupled very well to unit test this change.. It would be easier if we DI `ConnectionFactory` instead of creating it on the fly. I will try to refactor the code a little bit to make it easier. For this PR, I think it is fine to go without test changes if it is hard to do.

tebeco:
Injecting `IOptions<ServiceOptions>` in `ServiceEndpointProvider` is fine then ?
I attached it to the nearest `Connection` "thing" as it would let people choose different proxy for each `ServiceEndpoint` 
it sounds a bit stupid from me as the `Endpoint` is always `Azure` right ?

tebeco:
is `git push --force` ok for this PR ?

tebeco:
Also, i don't understand why in `dev` or `master` i have an error in ServiceLifetimeManagerFacts.cs
```
ServiceLifetimeManagerFacts.cs(32,25): error CS0246: The type or namespace name 'NewtonsoftJsonHubProtocol' could not be found (are you missing a using directive or an assembly reference?) [C:\workspace\github\azure\azure-signalr\test\Microsoft.Azure.SignalR.Tests\Microsoft.Azure.SignalR.Tests.csproj]
```

I have to comment it locally in order to be able to build, but it seems the CI here have no issue whatsoever.

vicancy:
@tebeco Sure, your can push to your forked branch using `--force`

For `NewtonsoftJsonHubProtocol`, my local resolved one is from `netcoreapp3.0` dotnet references. Do you have the prerequisites installed https://github.com/Azure/azure-signalr/blob/dev/docs/build-source.md#install-pre-requistes?

tebeco:
```
Install pre-requistes
Building Azure SignalR Service SDK requires:

Visual Studio 2019 Preview. https://visualstudio.com
To install the exact required components, run eng/scripts/InstallVisualStudio.ps1.
PS> ./eng/scripts/InstallVisualStudio.ps1
Git. https://git-scm.org
AspNetCore 3.0 Preview Runtime (Version >= 3.0.0-preview3-27503-5) https://dotnet.microsoft.com/download/dotnet-core/3.0
AspNetCore 3.0 SDK (Version >= 3.0.100-preview4-011136). Install from AspNetCore code repo before release https://github.com/dotnet/core-sdk#installers-and-binaries.
```

* Vs2019 `16.0.4` OK
* Vs 2019 Preview `16.1.0 Preview 3.0` OK
* `PS> ./eng/scripts/InstallVisualStudio.ps1` <== error 404 (but i guess Vs Workloads are installed already, else i suppose vs2019 would not be nice on solution load)
* Git, seems ok or i could not have pushed ^^
* AspNetCore 3.0 Preview Runtime (Version >= 3.0.0-preview3-27503-5) ==> SDK 3.0.0-preview5 installed

### Do i need to install this on top of `dotnet sdk preview5` ?
* AspNetCore 3.0 SDK (Version >= 3.0.100-preview4-011136). Install from AspNetCore code repo before release

tebeco:
There are now issue with the `ServiceEndpointOptions`, for a reason i can't understand i see :
`public ServiceEndpointProvider(ServiceEndpoint endpoint, string appName = "", IOptions<ServiceOptions> serviceOptions = null, TimeSpan? ttl = null)`

It's nullable -_-'
And one of the call look like :
`_endpointProvider = new ServiceEndpointProvider(_endpoint, appName: _serviceManagerOptions.ApplicationName);`

So if the call originate from `ServiceManager`, the `IOptions` will be discarded

tebeco:
@vicancy is that what you had in mind ?

vicancy:
Thanks @tebeco, this PR looks good to me. The `ServiceEndpointOptions` is indeed a special case though.

tebeco:
Thanks.

PR #538, opened 2019-05-21
delete unnecessary parameter

realwanpengli:
for https://github.com/Azure/azure-signalr/pull/534#issuecomment-493949170

### DIFF
--- a/src/Microsoft.Azure.SignalR.Management/ServiceManager.cs
+++ b/src/Microsoft.Azure.SignalR.Management/ServiceManager.cs
@@ -14,6 +14,7 @@ using Microsoft.Azure.SignalR.Common.ServiceConnections;
 using Microsoft.Azure.SignalR.Protocol;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.Azure.SignalR.Management
 {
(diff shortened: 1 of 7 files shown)


src/Microsoft.Azure.SignalR.Management/ServiceManager.cs
```diff
@@ -30,7 +31,13 @@ internal ServiceManager(ServiceManagerOptions serviceManagerOptions, string prod
         {
             _serviceManagerOptions = serviceManagerOptions;
             _endpoint = new ServiceEndpoint(_serviceManagerOptions.ConnectionString, EndpointType.Secondary);
-            _endpointProvider = new ServiceEndpointProvider(_endpoint, appName: _serviceManagerOptions.ApplicationName);
+            var optionWrapper = new OptionsWrapper<ServiceOptions>(new ServiceOptions
```
vicancy: use `Options.Create`

src/Microsoft.Azure.SignalR/EndpointProvider/ServiceEndpointProvider.cs
```diff
@@ -22,7 +22,7 @@ internal class ServiceEndpointProvider : IServiceEndpointProvider
         private readonly TimeSpan _accessTokenLifetime;
         private readonly IServiceEndpointGenerator _generator;
 
-        public ServiceEndpointProvider(ServiceEndpoint endpoint, string appName = "", IOptions<ServiceOptions> serviceOptions = null, TimeSpan? ttl = null)
+        public ServiceEndpointProvider(ServiceEndpoint endpoint, IOptions<ServiceOptions> serviceOptions, TimeSpan? ttl = null)
         {
             var connectionString = endpoint.ConnectionString;
```
vicancy: for `serviceOptions`, check null

realwanpengli: But we alreadly check `connectionString` in next line (line 28)

realwanpengli: I think passing serviceOption is unnecessary, the provider uses ApplicationName only

realwanpengli: already update in commit a3513f7

PR #543, opened 2019-05-22
remove unnecessary parameters

realwanpengli:


src/Microsoft.Azure.SignalR/EndpointProvider/ServiceEndpointProvider.cs
```diff
@@ -11,48 +11,37 @@ namespace Microsoft.Azure.SignalR
 {
     internal class ServiceEndpointProvider : IServiceEndpointProvider
     {
-        internal static readonly string ConnectionStringNotFound =
+        public static readonly string ConnectionStringNotFound =
```
vicancy: why to public?

src/Microsoft.Azure.SignalR/EndpointProvider/ServiceEndpointProvider.cs
```diff
@@ -11,48 +11,37 @@ namespace Microsoft.Azure.SignalR
...
             if (string.IsNullOrEmpty(connectionString))
             {
                 throw new ArgumentException(ConnectionStringNotFound);
             }
 
-            _accessTokenLifetime = ttl ?? Constants.DefaultAccessTokenLifetime;
+            _accessTokenLifetime = serviceOptions.Value.AccessTokenLifetime;
 
-            _endpoint = endpoint.Endpoint;
+            var endpointStr = endpoint.Endpoint;
```
vicancy: better naming

src/Microsoft.Azure.SignalR/EndpointProvider/ServiceEndpointProvider.cs
realwanpengli: also used in `ServiceEndpointManager`

PR #552, opened 2019-05-29
add more e2e for sdk (core)

realwanpengli:


test/Microsoft.Azure.SignalR.AspNet.Tests/E2ETest/TestClientSet.cs
```diff
@@ -58,5 +75,14 @@ public Task StopAsync()
 
             return Task.CompletedTask;
         }
+
+        public Task ManageGroupAsync(string methodName, IDictionary<int, string> connectionGroupMap)
+        {
+            return Task.WhenAll(from entry in connectionGroupMap
+                let proxyInd = entry.Key
+                let groupName = entry.Value
+                where connectionGroupMap.ContainsKey(proxyInd)
```
zackliu: Why it need to check `ContainsKey` again.

test/Microsoft.Azure.SignalR.AspNet.Tests/E2ETest/AspNetSignalRServiceE2EFacts.cs
```diff
@@ -6,11 +6,11 @@
...
 namespace Microsoft.Azure.SignalR.AspNet.Tests
 {
-    public class AspNetSignalRServiceE2EFacts : ServiceE2EFactsBase
-    {
-        public AspNetSignalRServiceE2EFacts(ITestOutputHelper output)
-            : base(new TestServer(), (string serverUrl, int count) => new TestClientSet(serverUrl, count), output)
-        {
-        }
-    }
+    //public class AspNetSignalRServiceE2EFacts : ServiceE2EFactsBase
```
zackliu: Comment is strange. Do we still need this class?

test/Microsoft.Azure.SignalR.Tests/Infrastructure/TestHub.cs
```diff
@@ -1,15 +1,88 @@
...
+        }
+
+        public void Broadcast(string message)
+        {
+            Clients.All.SendAsync(nameof(Broadcast), message);
+        }
+
+        public void JoinGroup(string groupName)
+        {
+            Groups.AddToGroupAsync(Context.ConnectionId, groupName);
```
zackliu: `AddToGroupAsync` and `RemoveFromGroupAsync` can not guarantee order now if they're not `await`. Do we need to change to `public Task JoinGroup()`

test/Microsoft.Azure.SignalR.AspNet.Tests/E2ETest/AspNetSignalRServiceE2EFacts.cs
realwanpengli: Will uncomment it. Yes, we need this to run test in net461 framework.

test/Microsoft.Azure.SignalR.Tests/Infrastructure/TestHub.cs
realwanpengli: But `await Groups.AddToGroupAsync(Context.ConnectionId, groupName);` still cannot guarantee the order, later messages might be sent in other server connections, is that right?

For now, I await `DefaultDelayMilliseconds` seconds after join/leave group to "make sure" the connection joined/leaved the group.

zackliu: I've merged ack-able group, so it can guarantee the order now.

realwanpengli: great!

test/Microsoft.Azure.SignalR.Tests/E2ETest/TestServer.cs
```diff
@@ -30,7 +32,16 @@ protected override Task StartCoreAsync(string serverUrl, ILoggerFactory loggerFa
 
         public override async Task StopAsync()
         {
+
             await _host.StopAsync();
+
+            // dispose client connections
+            var clientConnections = _host.Services.GetRequiredService<IClientConnectionManager>().ClientConnections;
+            clientConnections.Clear();
```
vicancy: >clientConnections [](start = 12, length = 17)

do we need to dispose the connections before clear the set?

realwanpengli: These connections are `ServiceConnectionContext`, which don't implement `IDisposable`, so we don't need to dispose them.

PR #562, opened 2019-06-10
Wanl/more asp e2e test

realwanpengli:


test/Microsoft.Azure.SignalR.AspNet.Tests/E2ETest/TestHub.cs
```diff
@@ -1,15 +1,96 @@
...
+using System.Web.UI.WebControls;
 using Microsoft.AspNet.SignalR;
+using Serilog;
+using Serilog.Core;
 
 namespace Microsoft.Azure.SignalR.AspNet.Tests
 {
     public class TestHub : Hub
     {
+        private static ConcurrentDictionary<string, bool> _connectedConnections = new ConcurrentDictionary<string, bool>();
```
vicancy: >static [](start = 16, length = 6)

why `static`?

test/Microsoft.Azure.SignalR.AspNet.Tests/E2ETest/TestServer.cs
```diff
@@ -2,27 +2,59 @@
...
+        protected override Task StartCoreAsync(string serverUrl, ITestOutputHelper output)
+        {
+            TestHub.ClearConnectedConnectionAndUser();
+
+            _loggerFactory = new LoggerFactory().AddXunit(output);
+
+            _webApp = WebApp.Start(new StartOptions(serverUrl), app =>
+            {
+                var hubConfiguration = Utility.GetActualHubConfig(_loggerFactory);
+                hubConfiguration.Resolver.Register(typeof(IUserIdProvider), () => Activator.CreateInstance(typeof(UserIdProvider)));
```
vicancy: >CreateInstance [](start = 92, length = 14)

singleton is enough

test/Microsoft.Azure.SignalR.AspNet.Tests/E2ETest/TestServer.cs
```diff
@@ -2,27 +2,59 @@
...
+                GlobalHost.TraceManager.Switch.Level = SourceLevels.Information;
+            });
             return Task.CompletedTask;
         }
 
         public override Task StopAsync()
         {
-            _webApp.Dispose();
+            _webApp?.Dispose();
+            _loggerFactory.Dispose();
```
vicancy: >Dispose [](start = 27, length = 7)

`?.` as `_loggerFactory` is not inited until `StartCoreAsync`

test/Microsoft.Azure.SignalR.AspNet.Tests/E2ETest/TestHub.cs
realwanpengli: When you call a method in hub, the hub is created by reflection, so the non-static member is created and disposed for every call/request. So I need to kepp it static to use the data across methods.

vicancy: How about using a `ConnectionManager` instead?

PR #567, opened 2019-06-13
Upgrade korebuild and build doc for .NET Core Preview6

JialinXin:


docs/build-source.md
```diff
@@ -7,14 +7,14 @@ Building Azure SignalR Service SDK from source allows you tweak and customize th
 
 Building Azure SignalR Service SDK requires:
 
-* Visual Studio **2019 Preview**. <https://visualstudio.com>
+* Visual Studio **2019 (Preview)**. <https://visualstudio.com>
```
vicancy: 2019 is already released

JialinXin: Right. But Preview can experience latest features. Since we're now keep up with latest .NET Core SDK and suggest to work with latest Visual Studio.

PR #570, opened 2019-06-17
Change endpoint online to push mode

vicancy:


src/Microsoft.Azure.SignalR.Common/EndpointRouters/DefaultMessageRouter.cs
```diff
@@ -9,40 +9,41 @@ namespace Microsoft.Azure.SignalR
     internal class DefaultMessageRouter : IMessageRouter
     {
         /// <summary>
-        /// Broadcast to all available endpoints
+        /// Broadcast to all available endpoints including online and offline, when the endpoint is offline, there is a need to notify user about it because there is potential message lose
```
JialinXin: >lose [](start = 184, length = 4)

typo: loss? lost?

test/Microsoft.Azure.SignalR.Tests/TestServiceConnection.cs
```diff
@@ -14,19 +15,27 @@ internal sealed class TestServiceConnection : IServiceConnection
...
+        public TestServiceConnection(ServiceConnectionStatus status = ServiceConnectionStatus.Connected, bool throws = false, TaskCompletionSource<object> writeAsyncTcs = null)
         {
             Status = status;
             _throws = throws;
+            _writeAsyncTcs = writeAsyncTcs;
         }
 
         public Task StartAsync(string target = null)
         {
+            ConnectionStatusChanged?.Invoke(new StatusChange(ServiceConnectionStatus.Inited, Status));
```
JialinXin: >Inited [](start = 85, length = 6)

why added an "Inited" status but only used in test? not Connecting?

vicancy: `inited` status is added for ServiceConnectionContainer to have an initial state `Inited` instead of `Disconnected` as the initial state, so that ServiceConnectionContainer can get notified when the inner server connections has status changed to `Disconnected`. And I am not able to reuse `Connecting` as for backward compatibility, the status of `ServiceConnectionContainer` should be considered as online until it starts the inner service connection.

vicancy: So with `inited` we explicitly defined a stage for the server connection between it is created to it is `connected`. I think it is clear, as for now, we consider the endpoint as `online` in this stage, which is when the app server starts. And later, we can easily change the endpoint status for this stage if there is a requirement.

PR #576, opened 2019-06-24
Add service active check to weak connections

vicancy:
The service returns `active` ping message when there is any client connected to it, and returns `inactive` ping message when there is no client connected to it. 

For weak connections, there is no need to send server->client messages to a `inactive` service.

From SDK side, we check service status every 10 seconds, whenever we get an `active` ping message, we consider the service as `active`. When we get at least 5 continuous `inactive` ping messages in at least 10 minutes, we consider the service as `inactive`.

Due to the delay of the `active`/`inactive` status, we only enable this for weak connections.

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/ServiceConnections/WeakServiceConnectionContainer.cs
+++ b/src/Microsoft.Azure.SignalR.Common/ServiceConnections/WeakServiceConnectionContainer.cs
@@ -15,17 +15,17 @@ namespace Microsoft.Azure.SignalR.Common.ServiceConnections
         private const int CheckWindow = 5;
         private static readonly TimeSpan DefaultGetServiceStatusInterval = TimeSpan.FromSeconds(10);
         private static readonly long DefaultGetServiceStatusTicks = DefaultGetServiceStatusInterval.Seconds * Stopwatch.Frequency;
+        private static readonly TimeSpan CheckTimeSpan = TimeSpan.FromMinutes(10);
 
-        private readonly TimeSpan CheckTimeSpan = TimeSpan.FromMinutes(10);
         private readonly object _lock = new object();
         private int _inactiveCount;
         private DateTime? _firstInactiveTime;
         private long _lastSendTimestamp;
 
-        // active ones are those who client connections connected to the whole endpoint
+        // active ones are those whose client connections connected to the whole endpoint
         private volatile bool _active = true;
 
-        private TimerAwaitable _timer;
+        private readonly TimerAwaitable _timer;
 
         protected override ServerConnectionType InitialConnectionType => ServerConnectionType.Weak;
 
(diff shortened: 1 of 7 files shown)


src/Microsoft.Azure.SignalR.Common/ServiceConnections/WeakServiceConnectionContainer.cs
```diff
@@ -9,17 +12,189 @@ namespace Microsoft.Azure.SignalR.Common.ServiceConnections
...
+
+                    // Inactive it only when it checks over 5 times and elapsed for over 10 minutes
+                    if (_inactiveCount >= checkWindow && DateTime.UtcNow - _firstInactiveTime >= checkTimeSpan)
+                    {
+                        return false;
+                    }
+                    else
+                    {
+                        return true;
+                    }
```
JialinXin: Add log?

vicancy: logged in L44

src/Microsoft.Azure.SignalR.Common/ServiceMessages/ServicePingMessageExtension.cs
```diff
@@ -0,0 +1,48 @@
...
+                var item1 = pingMessage.Messages[index];
+                var item2 = pingMessage.Messages[index + 1];
+                if (item1 == key)
+                {
+                    value = item2;
+                    return true;
+                }
+
+                index += 2;
+            }
```
JialinXin: would normal ping mixed together, then the order not work?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/WeakServiceConnectionContainer.cs
```diff
@@ -9,17 +12,189 @@ namespace Microsoft.Azure.SignalR.Common.ServiceConnections
...
+
+            public static void EndpointInactive(ILogger logger, HubServiceEndpoint endpoint)
+            {
+                _endpointInactive(logger, endpoint, endpoint.Hub, null);
+            }
+
+            public static void EndpointActive(ILogger logger, HubServiceEndpoint endpoint)
+            {
+                _endpointActive(logger, endpoint, endpoint.Hub, null);
+            }
```
JialinXin: logs not use.

src/Microsoft.Azure.SignalR.Common/ServiceMessages/ServicePingMessageExtension.cs
vicancy: all the ping follows the same pattern, we use key to distinguish different kind of pings

JialinXin: get it.

---
In reply to: [297487067](https://github.com/Azure/azure-signalr/pull/576#discussion_r297487067) [](ancestors = 297487067)

src/Microsoft.Azure.SignalR.Common/ServiceConnections/WeakServiceConnectionContainer.cs
```diff
@@ -9,17 +12,189 @@ namespace Microsoft.Azure.SignalR.Common.ServiceConnections
...
+        private static readonly TimeSpan DefaultGetServiceStatusInterval = TimeSpan.FromSeconds(10);
+        private static readonly long DefaultGetServiceStatusTicks = DefaultGetServiceStatusInterval.Seconds * Stopwatch.Frequency;
+
+        private readonly TimeSpan CheckTimeSpan = TimeSpan.FromMinutes(10);
+        private readonly object _lock = new object();
+        private int _inactiveCount;
+        private DateTime? _firstInactiveTime;
+        private long _lastSendTimestamp;
+
+        // active ones are those who client connections connected to the whole endpoint
```
JialinXin: >who [](start = 33, length = 3)

typo: whose

PR #579, opened 2019-06-27
move signalr and management e2e test to standalone project

realwanpengli:
avoid HttpClientConnection version conflict


test/Microsoft.Azure.SignalR.Tests/Infrastructure/TestHub.cs
```diff
@@ -1,73 +1,11 @@
...
-using System;
-using System.Collections.Concurrent;
-using System.Linq;
-using System.Threading.Tasks;
 using Microsoft.AspNetCore.SignalR;
-using Microsoft.Azure.SignalR.Tests.Common;
 
 namespace Microsoft.Azure.SignalR.Tests
 {
     internal class TestHub : Hub
```
JialinXin: Is it still in use? or can be removed?

PR #585, opened 2019-07-02
Add back group partition send

vicancy:


vicancy:
Perf test indicates that performance impact is acceptable.

PR #595, opened 2019-07-12
Fix aspnet application exception cases

vicancy:
1. Case 1: , when OnConnected throws, the exception is swallowed, and Disconnected will never trigger
2. Case 2: when client connection closed, the client manager never cleans the disconnected connection

### DIFF
--- a/src/Microsoft.Azure.SignalR.AspNet/ClientConnections/ClientConnectionManager.cs
+++ b/src/Microsoft.Azure.SignalR.AspNet/ClientConnections/ClientConnectionManager.cs
@@ -3,8 +3,10 @@
 
 using System;
 using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Threading.Tasks;
 using Microsoft.AspNet.SignalR;
 using Microsoft.AspNet.SignalR.Hosting;
 using Microsoft.AspNet.SignalR.Hubs;
@@ -28,7 +30,7 @@ namespace Microsoft.Azure.SignalR.AspNet
             _logger = loggerFactory?.CreateLogger<ClientConnectionManager>() ?? NullLogger<ClientConnectionManager>.Instance;
         }
 
-        public IServiceTransport CreateConnection(OpenConnectionMessage message, IServiceConnection serviceConnection)
+        public async Task<IServiceTransport> CreateConnection(OpenConnectionMessage message, IServiceConnection serviceConnection)
         {
             var dispatcher = new ClientConnectionHubDispatcher(_configuration, message.ConnectionId);
             dispatcher.Initialize(_configuration.Resolver);
@@ -39,7 +41,7 @@ namespace Microsoft.Azure.SignalR.AspNet
             if (dispatcher.Authorize(hostContext.Request))
             {
                 // ProcessRequest checks if the connectionToken matches "{connectionid}:{userName}" format with context.User
-                _ = dispatcher.ProcessRequest(hostContext);
+                await dispatcher.ProcessRequest(hostContext);
 
                 // TODO: check for errors written to the response
                 if (hostContext.Response.StatusCode != 200)
@@ -61,6 +63,13 @@ namespace Microsoft.Azure.SignalR.AspNet
         {
             return _clientConnections.TryGetValue(key, out serviceConnection);
         }
+
+        public bool TryRemoveServiceConnection(string connectionId, out IServiceConnection connection)
+        {
+            return _clientConnections.TryRemove(connectionId, out connection);
+        }
+
+        public IReadOnlyDictionary<string, IServiceConnection> ClientConnections => _clientConnections;
 
         internal HostContext GetHostContext(OpenConnectionMessage message, Stream responseStream, IServiceConnection serviceConnection)
         {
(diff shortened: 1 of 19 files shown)


src/Microsoft.Azure.SignalR.AspNet/ServerConnections/ServiceConnection.cs
```diff
@@ -67,61 +67,47 @@ protected override Task CleanupConnections()
...
+            return ForwardMessageToApplication(closeConnectionMessage.ConnectionId, closeConnectionMessage);
         }
 
-        protected override async Task OnMessageAsync(ConnectionDataMessage connectionDataMessage)
+        protected override Task OnMessageAsync(ConnectionDataMessage connectionDataMessage)
+        {
+            return ForwardMessageToApplication(connectionDataMessage.ConnectionId, connectionDataMessage);
+        }
+
+        private async Task ForwardMessageToApplication(string connectionId, ServiceMessage message)
```
JialinXin: >ForwardMessageToApplication(string connectionId, ServiceMessage message) [](start = 27, length = 72)

Can simplify to ForwardMessageToApplication(ServiceMessage message) as ConnectionId is always get from ServiceMessage.

src/Microsoft.Azure.SignalR.AspNet/ServerConnections/ServiceConnection.cs
```diff
@@ -67,61 +67,47 @@ protected override Task CleanupConnections()
...
             if (_clientConnections.TryGetValue(connectionId, out var clientContext))
             {
                 try
                 {
-                    await clientContext.Output.WriteAsync(connectionDataMessage);
+                    await clientContext.Output.WriteAsync(message);
                 }
                 catch (Exception e)
                 {
                     Log.FailToWriteMessageToApplication(Logger, connectionId, e);
```
JialinXin: >FailToWriteMessageToApplication [](start = 24, length = 31)

Shall we add a message type in the log so it would be clear to understand which message fail to write?

src/Microsoft.Azure.SignalR.AspNet/ClientConnections/ClientConnectionManager.cs
```diff
@@ -62,6 +64,13 @@ public bool TryGetServiceConnection(string key, out IServiceConnection serviceCo
             return _clientConnections.TryGetValue(key, out serviceConnection);
         }
 
+        public bool TryRemoveServiceConnection(string connectionId, out IServiceConnection connection)
+        {
+            return _clientConnections.TryRemove(connectionId, out connection);
+        }
+
+        public IReadOnlyDictionary<string, IServiceConnection> ClientConnections => _clientConnections;
```
JialinXin: >ClientConnections [](start = 63, length = 17)

only use for tests?

src/Microsoft.Azure.SignalR.AspNet/ServerConnections/ServiceConnection.cs
```diff
@@ -176,43 +163,41 @@ private void ProcessOutgoingMessages(ClientContext clientContext, ConnectionData
             }
         }
 
-        private async Task ProcessMessageAsync(string connectionId)
+        private async Task ProcessMessageAsync(ClientContext clientContext)
         {
             // Check if channel is created.
```
JialinXin: >// Check if channel is created. [](start = 12, length = 31)

can removed.

src/Microsoft.Azure.SignalR.AspNet/ServerConnections/ServiceConnection.cs
vicancy: However `ServiceMessage` class does not contain `ConnectionId` property, need a type check if we omit `connectionId`

src/Microsoft.Azure.SignalR.AspNet/ClientConnections/ClientConnectionManager.cs
vicancy: For now it is. Intent to keep to some extent consistent with the core version.

src/Microsoft.Azure.SignalR.AspNet/ServerConnections/ServiceConnection.cs
```diff
@@ -51,105 +55,129 @@ protected override Task DisposeConnection()
...
-        protected override Task CleanupConnections()
+        protected override async Task CleanupConnections()
         {
             try
             {
-                foreach(var connection in _clientConnections)
-                {
-                    PerformDisconnectCore(connection.Key);
-                }
+                await Task.WhenAll(_clientConnections.Select(s => PerformDisconnectCore(s.Key, true)));
```
JialinXin: >true [](start = 95, length = 4)

do cleanup still need to wait?

src/Microsoft.Azure.SignalR.AspNet/ServerConnections/ServiceConnection.cs
```diff
@@ -51,105 +55,129 @@ protected override Task DisposeConnection()
...
                 try
                 {
-                    await clientContext.Output.WriteAsync(closeConnectionMessage);
+                    await clientContext.Output.WriteAsync(message);
                 }
                 catch (Exception e)
                 {
-                    Log.FailToWriteMessageToApplication(Logger, connectionId, e);
+                    Log.FailToWriteMessageToApplication(Logger, message.GetType().Name, connectionId, e);
+                    await PerformDisconnectCore(connectionId, true);
```
JialinXin: >await PerformDisconnectCore(connectionId, true); [](start = 20, length = 48)

Seems only under the case which probably like channel broken or overflow that need to wait application, then will this be necessary to wait under this case? Normally wait won't get any better.

PR #596, opened 2019-07-15
Add connectionId reletated rest api and tests

zackliu:


test/Microsoft.Azure.SignalR.Tests.Common/Microsoft.Azure.SignalR.Tests.Common.csproj
```diff
@@ -15,5 +15,6 @@
     <PackageReference Include="Microsoft.AspNetCore.Testing" Version="$(MicrosoftAspNetCoreTestingVersion)" />
     <PackageReference Include="Microsoft.Extensions.Configuration.UserSecrets" Version="$(MicrosoftConfigurationUserSecretVersion)" />
     <PackageReference Include="Microsoft.Extensions.Logging.Testing" Version="$(MicrosoftExtensionsLoggingPackageVersion)" />
+    <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.3" />
```
realwanpengli: Why we need it? BTW, versions should be defined in `dependencies.props`

zackliu: I moved a util class to a more underlayer project. And we need to add reference to let this project to use ValueTask

test/Microsoft.Azure.SignalR.Tests.Common/E2ETest/TestHubConnectionManager.cs
```diff
@@ -58,5 +72,11 @@ public void RemoveUser(string user)
                 throw new InvalidOperationException($"Failed to remove user {user}. Connected users: {string.Join(", ", _connectedUsers.Keys)}");
             }
         }
+
+        public Task<string> WaitForConnectionCount(int n)
```
realwanpengli: WaitForConnectionCountAsync?

test/Microsoft.Azure.SignalR.Tests.Common/ParameterDelegator.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Azure.SignalR.Tests.Common
+{
+    public class ParameterDelegator
```
realwanpengli: can we use `ConfigurationBuilder().AddInMemoryCollection()` like https://msdn.microsoft.com/en-us/magazine/mt814420.aspx?f=255&MSPPError=-2147217396 instead of `ParameterDelegator`?

test/Microsoft.Azure.SignalR.E2ETests/Management/ServiceHubContextE2EFacts.cs
```diff
@@ -31,9 +32,11 @@ public class ServiceHubContextE2EFacts : VerifiableLoggedTest
         private static readonly string[] _groupNames = GetTestStringList("Group", GroupCount);
         private static readonly ServiceTransportType[] _serviceTransportType = new ServiceTransportType[] { ServiceTransportType.Transient, ServiceTransportType.Persistent };
         private static readonly string[] _appNames = new string[] { "appName", "", null };
+        private readonly ITestServerFactory _testServerFactory;
```
realwanpengli: How about make it static and constrcut here?

test/Microsoft.Azure.SignalR.Tests.Common/E2ETest/ITestServer.cs
```diff
@@ -1,14 +1,15 @@
...
+using System.Collections.Generic;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Azure.SignalR.Tests.Common
 {
     public interface ITestServer
     {
-        Task<string> StartAsync();
+        Task<string> StartAsync(Dictionary<string, string> configuration = null);
```
realwanpengli: Could you make the `HubConnectionManager` in the interface instead of casting, and also update https://github.com/Azure/azure-signalr/blob/dev/test/Microsoft.Azure.SignalR.Tests.Common/E2ETest/ServiceE2EFactsBase.cs which is also related to test server?

test/Microsoft.Azure.SignalR.E2ETests/Management/ServiceHubContextE2EFacts.cs
zackliu: I think it seems not the field should be static

test/Microsoft.Azure.SignalR.Tests.Common/E2ETest/ITestServer.cs
zackliu: Add to interface is fine, but scenarios in `ServiceE2EFactsBase` is more complicated. We need to use another pr to update it.

test/Microsoft.Azure.SignalR.E2ETests/SignalR/TestServer.cs
```diff
@@ -6,25 +6,32 @@
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Configuration;
 using Xunit.Abstractions;
+using System.Collections.Generic;
```
vicancy: >System [](start = 6, length = 6)

order

test/Microsoft.Azure.SignalR.Tests.Common/E2ETest/ITestServer.cs
realwanpengli: ok

Issue #539, opened 2019-05-21
Orphaned connection


alexiordan:
For some reasons, rarely the Hub doesn't get notified about a client disconnect. Today it happened related to some issues ( I don't know if they are network or server related) that caused several disconnects (10-20 times) between 09:30 - 11:00 UTC aprox.

Are there any means to know that a specific connection is no longer active?

vicancy:
One related bug https://github.com/Azure/azure-signalr/issues/416

vicancy:
We beleive 1.0.12 fixed the issue. Please reopen it if you still encounter such issue. Thanks for your support.

PR #612, opened 2019-07-23
Handle InstanceOfflinePing and close related clients

### LINKED ISSUE #416: SDK cannot detect some clients dropped when ASRS instance crashed.
Condition:
* ASRS instance crash.
* The ASRS instance for client connection is different from the instance for server connection.

Problem:
1. These clients will send ping, and app server do not notify offline for these clients.
2. No error will be throw.
3. These clients can be removed when server connection stopped.

JialinXin:
1. Store client-instance mapping from OpenConnectionMessages
2. Handle offline ping message to disconnect clients on crashes service instances

Fix #416, #539


src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -211,6 +213,10 @@ protected Task OnServiceErrorAsync(ServiceErrorMessage serviceErrorMessage)
 
         protected Task OnPingMessageAsync(PingMessage pingMessage)
         {
+            if (pingMessage.TryGetValue(Constants.ServicePingMessageKey.OfflineKey, out var instanceId) && !string.IsNullOrEmpty(instanceId))
+            {
+                return DisconnectClientConnections(instanceId);
```
vicancy: >DisconnectClientConnections [](start = 23, length = 27)

add some log

src/Microsoft.Azure.SignalR.AspNet/ServerConnections/ServiceConnection.cs
```diff
@@ -278,14 +279,48 @@ private static string GetString(ReadOnlySequence<byte> buffer)
 
             return Encoding.UTF8.GetString(buffer.ToArray());
         }
-        
+
+        protected override Task DisconnectClientConnections(string instanceId)
+        {
+            if (_clientConnections.Count() == 0 || string.IsNullOrEmpty(instanceId))
```
vicancy: >() [](start = 40, length = 2)

don't use the linq version `Count()`, use the property `Count`

src/Microsoft.Azure.SignalR.AspNet/ServerConnections/ServiceConnection.cs
```diff
@@ -278,14 +279,48 @@ private static string GetString(ReadOnlySequence<byte> buffer)
...
+            {
+                return Task.CompletedTask;
+            }
+            try
+            {
+                // Connected service instance is down, close client gracefully
+                var connections = _clientConnections.Where(c => c.Value.InstanceId == instanceId);
+                foreach (var connection in connections)
+                {
+                    var serviceMessage = new CloseConnectionMessage(connection.Value.ConnectionId, errorMessage: "Service transient error");
```
vicancy: >Service transient error [](start = 114, length = 23)

`The client is disconnected due to service internal error`?

src/Microsoft.Azure.SignalR.AspNet/ServerConnections/ServiceConnection.cs
```diff
@@ -278,14 +279,48 @@ private static string GetString(ReadOnlySequence<byte> buffer)
...
+            catch (Exception ex)
+            {
+                Log.FailedToCleanupConnections(Logger, ex);
+            }
+            return Task.CompletedTask;
+        }
+
+        private string GetInstanceId(IDictionary<string, StringValues> header)
+        {
+            if (header.ContainsKey(Constants.AsrsInstanceId))
```
vicancy: >ContainsKey [](start = 23, length = 11)

use TryGetValue to access the dict once

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
```diff
@@ -82,6 +83,30 @@ protected override async Task CleanupConnections()
...
+            }
+            try
+            {
+                // Connected service instance is down, close client gracefully
+                var connections = _connectionIds.Where(c => c.Value == instanceId);
+                foreach (var connection in connections)
+                {
+                    var serviceMessage = new CloseConnectionMessage(connection.Key, errorMessage: "Service transient error.");
+                    _ = PerformDisconnectAsyncCore(connection.Key, false);
+                    _ = WriteAsync(serviceMessage);
```
vicancy: >WriteAsync [](start = 24, length = 10)

this is write back to the service so no need?

JialinXin: I think it writes to clients, isn't it?

---
In reply to: [306695578](https://github.com/Azure/azure-signalr/pull/612#discussion_r306695578) [](ancestors = 306695578)

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
```diff
@@ -66,15 +67,20 @@ protected override Task DisposeConnection()
...
                 if (_connectionIds.Count == 0)
                 {
                     return;
                 }
-                await Task.WhenAll(_connectionIds.Select(s => PerformDisconnectAsyncCore(s.Key, false)));
+                var connectionIds = _connectionIds.Select(s => s.Key);
+                if (!string.IsNullOrEmpty(instanceId))
+                {
+                    connectionIds = _connectionIds.Where(s => s.Value == instanceId).Select(s => s.Key);
+                }
```
vicancy:
```
var connectionIds = _connectionIds;

if (!string.IsNullOrEmpty(instanceId)){
connectionIds = _connectionIds.Where(s=>s.Value==instanceId);
}

 await Task.WhenAll(connectionIds.Select(s => PerformDisconnectAsyncCore(s.Key, false)));
```

test/Microsoft.Azure.SignalR.Tests/ServiceConnectionFacts.cs
```diff
@@ -451,6 +452,52 @@ await proxy.WriteMessageAsync(new PingMessage()
...
+
+            // Server received instance offline ping on instanceId1 and trigger cleanup related client1
+            await proxy.WriteMessageAsync(new PingMessage()
+            {
+                Messages = new[] { "offline", instanceId1 }
+            });
+
+            // Validate client1 is closed and client2 is still connected
+            var disconnectTask = proxy.WaitForConnectionCloseAsync(connectionId1);
+            await disconnectTask.OrTimeout();
```
vicancy: >disconnectTask [](start = 18, length = 14)

any test to validate that disconnect task is handled?

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
JialinXin: Tried before, but `Where` filter will return `Generic Dictionary` and can not convert to `ConcurrentDictionary`.

test/Microsoft.Azure.SignalR.Tests/ServiceConnectionFacts.cs
JialinXin: current validation not enough? what else shall we check?

vicancy: I don't think this one checks the hub's disconnect event.

---
In reply to: [307564663](https://github.com/Azure/azure-signalr/pull/612#discussion_r307564663) [](ancestors = 307564663)

Issue #623, opened 2019-08-09
[Bug] Context.GetHttpContext().Connection.RemoteIpAddress returns null


ispysoftware:
It stopped working about 9 hours ago - it connects but doesn't relay any messages to the backend server.

I've restarted the service and even deployed a new instance of it with the same results. Something has crapped it's pants on Azure's side. This has been working for months.

ispysoftware:
seems to only be an issue on Central US

vicancy:
Resolved in 1.0.12

ispysoftware:
This bug cost us a lot of money - in lost customers and site downtime. What kind of assurances do we have that it won't happen again?  Is there a way to specify an azure signalr server version to use? 

I'm extremely nervous that our platform will go down at any moment due to a botched push from your side that we have no control over and no notification that it's being "upgraded".

PR #625, opened 2019-08-12
add e2e test get remote ip

ccic:
Add a E2E test case for issue #623


test/Microsoft.Azure.SignalR.E2ETests/SignalR/TestHub.cs
```diff
@@ -31,6 +31,15 @@ public override Task OnDisconnectedAsync(Exception exception)
             return Task.CompletedTask;
         }
 
+        // Verify whether 'get client IP' is working or not
+        public void TestClientIPEcho(string message)
+        {
+            if (!string.IsNullOrEmpty(Context.GetHttpContext().Connection.RemoteIpAddress.ToString()))
```
vicancy: shall we also test headers and query strings are not missing?

PR #630, opened 2019-08-14
fix for tcs's RunContinuationsAsynchronously

realwanpengli:


vicancy:
Adding a ut for this.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -32,7 +32,7 @@ internal abstract class ServiceConnectionBase : IServiceConnection
 
         private readonly SemaphoreSlim _serviceConnectionLock = new SemaphoreSlim(1, 1);
 
-        private readonly TaskCompletionSource<bool> _serviceConnectionStartTcs = new TaskCompletionSource<bool>(TaskContinuationOptions.RunContinuationsAsynchronously);
+        private readonly TaskCompletionSource<bool> _serviceConnectionStartTcs = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
```
scale-tone: So, where are unit tests for this?

As we can now see, the code here is extremely fragile (confusing a ctor, that caused a critical hidden bug is a strong evidence of that). How can it **still** not be covered by tests?

realwanpengli:
appveyor always fail, wierd.
see #639

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
realwanpengli: Added. this bug only occurs in console app.

PR #658, opened 2019-09-11
Message order fix

KKhurin:
This change fixes SignalR core message order issues on SDK side.

Each new client connection would establish an async local context for the duration of the client connection. This async local context flows through every user hub method invocation and ensures that the same service connection will be used to send all the messages back to the service _throughout the duration of this client connection_. If for some reason this service connection is no longer available, all future calls to the service originated from this client connection will be assigned a new service connection. 

We provide a new utility class ServiceConnectionScope to be able to track these connection changes (via ConnectionChanged) and react to them if the application logic requires so. This ServiceConnectionScope class can also be used to provide "stickiness" to the same service connection when sending to clients occur outside of the Hub call async context (on a separate thread or a task that didn't flow the execution context from any hub method).


src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -288,40 +289,73 @@ protected virtual ServiceConnectionStatus GetStatus()
...
+                {
+                    try
+                    {
+                        // still possible the connection is not valid
+                        await connection.WriteAsync(serviceMessage);
+                        return connection;
+                    }
+                    catch (ServiceConnectionNotActiveException)
+                    {
+                        if (retry == maxRetry - 1)
```
chenkennt: should be `retry == maxRetry`?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -288,40 +289,73 @@ protected virtual ServiceConnectionStatus GetStatus()
...
+            {
+                // indicate invalid usage
+                Debug.Assert(connectionHolder != null);
+            }
+        }
+
+        private async Task<IServiceConnection> WriteWithRetry(ServiceMessage serviceMessage, IServiceConnection connection)
+        {
+            // go through all the connections, it can be useful when one of the remote service instances is down
+            var initial = StaticRandom.Next(-FixedConnectionCount, FixedConnectionCount);
```
chenkennt: should include on demand connection as well?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionScopeInternal.cs
```diff
@@ -0,0 +1,57 @@
...
+
+using System;
+using System.Diagnostics;
+using System.Threading;
+
+namespace Microsoft.Azure.SignalR
+{
+    internal class ServiceConnectionScopeHolder
+    {
+        private static AsyncLocal<ServiceConnectionScopeHolder> s_AsyncLocalSCC = new AsyncLocal<ServiceConnectionScopeHolder>();
```
davidfowl: 😢

yanweizh:
Add some ut?

vicancy:
Add some ut?

src/Microsoft.Azure.SignalR/ServiceConnectionScope.cs
```diff
@@ -0,0 +1,74 @@
...
+
+using System;
+
+namespace Microsoft.Azure.SignalR
+{
+    /// <summary>
+    /// When using Azure SignalR Service ServiceConnectionScope class tries to ensure that all the messages being sent to clients 
+    /// within its scope are using the same connection from App Server to the Service thus helping to ensure the order of messages.
+    /// </summary>
+    public class ServiceConnectionScope : IDisposable
```
davidfowl: We expect people to use this in their hubs? I think that's a non starter and it ties your Hubs to the service logic.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -288,40 +289,73 @@ protected virtual ServiceConnectionStatus GetStatus()
...
+            var connectionHolder = ServiceConnectionScopeInternal.Holder;
+            var connection = connectionHolder?.ServiceConnection;
+            var connectionUsed = await WriteWithRetry(serviceMessage, connection);
+
+            // Try to persist the connection choice for the subsequent calls within the same async flow
+            if (connectionHolder != null)
+            {
+                if (connectionUsed != connection)
+                {
+                    connectionHolder.ServiceConnection = connectionUsed;
```
chenkennt: this should be set immediately after a connection is selected?

KKhurin:
@chenkennt
@vicancy
@davidfowl
Here is the plan. 
Please provide feedback and/or questions if any.

1. Abandon this PR in its current shape.

KKhurin:
2. Do some ground work for connection handling:
- git rid of the semaphore used to serialize writes to the pipe; replace it with an explicit queue
- merge fixed and on demand connection pools
- better handle underlying socket disconnects and errors
- reassign messages in the explicit queues to another connection to avoid losing messages when the connection gets broken
- make the connection selection extensible to allow future improvements (better load balancing, better routing, etc)
These changes won’t fix the message ordering, but they will provide necessary API surface to handle MO properly.

KKhurin:
3. Re-introduce the current fix (using AsyncLocal) with the following changes:
- No [ServiceConnectionScopeHolder](https://github.com/Azure/azure-signalr/pull/658#discussion_r324549352) (fall back to the solution described in #4 below for the cases when we can’t use async local)
- Add [unit tests](https://github.com/Azure/azure-signalr/pull/658#issuecomment-531676919)
- Take advantage of the changes in [#2 above](https://github.com/Azure/azure-signalr/pull/658#issuecomment-544778544) to [better handle connection selection]( https://github.com/Azure/azure-signalr/pull/658#discussion_r323096830)

KKhurin:
4. Implement the fallback for persistent connection selection for the scenarios when #2 is not applicable.
The fallback would use an extensibility point (introduced in [#2 above](https://github.com/Azure/azure-signalr/pull/658#issuecomment-544778544)) to select (and persist the selection) the appropriate connection & queue.
The current plan is to use cache with a sliding expiration to store the associations between the connection and the recipient.

KKhurin:
5. Further optimize the service connection selection – for example for p2p scenarios take into account which clients are connected to each RT to minimize in-service routing. I expect these additional optimizations (when applicable) to provide the best overall optimization, while [#2 above](https://github.com/Azure/azure-signalr/pull/658#issuecomment-544778544) to provide better scalability and performance than [#4 above](https://github.com/Azure/azure-signalr/pull/658#issuecomment-544779195). While being far from ideal (use additional memory to store explicit associations per client/group/user/list), the #4 is going to be a good enough fallback mechanism when no better optimizations exist or possible.

PR #659, opened 2019-09-11
Prepare 3.0 release

JialinXin:
1. Upgrade build version to 3.0.100
2. Deprecate PackageIconUrl: https://aka.ms/deprecateIconUrl 
3. Upgrade blazor app sample to latest template 
4. Fix tests

2-4 relate to 1.


src/Microsoft.Azure.SignalR.Management/Microsoft.Azure.SignalR.Management.csproj
```diff
@@ -46,4 +43,11 @@
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0'">
     <PackageReference Include="Microsoft.AspNetCore.SignalR" Version="$(MicrosoftAspNetCoreSignalRPackageVersion)" />
   </ItemGroup>
+
+  <!-- Specifying signing for multiple files in a package. -->
+  <ItemGroup>
```
vicancy: move to `directory.Build.props`?

test/Microsoft.Azure.SignalR.Tests/NegotiateHandlerFacts.cs
```diff
@@ -357,6 +358,8 @@ public override ServiceEndpoint GetNegotiateEndpoint(HttpContext context, IEnume
                 {
                     context.Response.StatusCode = 400;
                     var response = Encoding.UTF8.GetBytes("Invalid request");
+                    // In latest DefaultHttpContext, response body is set to null
+                    context.Response.Body = new MemoryStream();
```
vicancy: check if it is null first?

JialinXin: Directly access Body to check null will return ObjectNullReference exception.

JialinXin: Looks in AspNetCore tests they just new MemoryStream() for the DefaultHttpContext(). Let's follow it.
https://github.com/aspnet/AspNetCore/blob/master/src/SignalR/common/Http.Connections/test/ServerSentEventsTests.cs#L119

src/Microsoft.Azure.SignalR.Management/Microsoft.Azure.SignalR.Management.csproj
JialinXin: I thought different packages have different dlls. Looks move the common Directory.Build.props also works. Nice catch.

PR #665, opened 2019-09-20
[WIP] Add fields to HandshakeRequest & ConnectionData

terencefan:
1. Add `MigratableStatus` field to `HandshakeRequestMessage` to indicate if the client connection mapped to the designated server connection can be migrated to another server connection.
2. Add `SequenceId` field to `ConnectionDataMessage` to help the server-side SDK to guarantee the message order (cause it could be sent from different `signalr-runtime` instance in the future)

10/12/2019 updates:

1. Gives the ability that migrates the current CC to another SC whenever the associated SC has been dropped accidentally.
2. Gives the ability to gracefully restart the server without the
current established CC to be dropped.

Changes:
1. Add an `asrs-migrated-from` header to `OpenConnectionMessage` to indicate the
client connection is migrated from another server.
2. Ignore the `HandshakeResponse` receive from the target server after a
migration process.
3. Send the last received SeqID to ASRS in `CloseConnectionMessage` so
the ASRS could replay the current unfinished message to the target
server.


src/Microsoft.Azure.SignalR.Protocols/ServiceMessage.cs
```diff
@@ -30,6 +30,14 @@ public class HandshakeRequestMessage : ServiceMessage
         /// </summary>
         public int Version { get; set; }
 
+        public const int TypeDefault = 0;
+        public const int TypeOnDemand = 0;
+        public const int TypeWeak = 0;
```
JialinXin: Change to enum type like https://github.com/Azure/azure-signalr/blob/dev/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServerConnectionType.cs ?

src/Microsoft.Azure.SignalR.Protocols/ServiceProtocol.cs
```diff
@@ -570,12 +572,17 @@ private static CloseConnectionMessage CreateCloseConnectionMessage(byte[] input,
...
+        private static ConnectionDataMessage CreateConnectionDataMessage(byte[] input, ref int offset, int arrayLength)
         {
             var connectionId = ReadString(input, ref offset, "connectionId");
             var payload = ReadBytes(input, ref offset, "payload");
 
-            return new ConnectionDataMessage(connectionId, payload);
+            var message = new ConnectionDataMessage(connectionId, payload);
+            if (arrayLength >= 4)
+            {
+                message.SequenceId = ReadInt32(input, ref offset, "sequenceId");
```
JialinXin: Int64 for long?

src/Microsoft.Azure.SignalR.Protocols/ServiceProtocolConstants.cs
```diff
@@ -26,4 +26,11 @@ public static class ServiceProtocolConstants
         public const int LeaveGroupWithAckMessageType = 19;
         public const int AckMessageType = 20;
     }
+
+    public enum ServerMigratableStatus
+    {
+        Off = 0,
+        Limited = 1,
+        Any = 2,
+    }
```
JialinXin: Put this to a separate file under Common project.

src/Microsoft.Azure.SignalR.Protocols/ServiceProtocol.cs
```diff
@@ -227,20 +227,20 @@ private static void WriteMessageCore(ServiceMessage message, Stream packer)
...
-            {
-                MessagePackBinary.WriteArrayHeader(packer, 4);
-                MessagePackBinary.WriteInt32(packer, ServiceProtocolConstants.HandshakeRequestType);
-                MessagePackBinary.WriteInt32(packer, message.Version);
+            MessagePackBinary.WriteArrayHeader(packer, 5);
+            MessagePackBinary.WriteInt32(packer, ServiceProtocolConstants.HandshakeRequestType);
+            MessagePackBinary.WriteInt32(packer, message.Version);
+
+            if (message.ConnectionType == 0) { // add paddings
+                MessagePackBinary.WriteInt32(packer, 0);
```
vicancy: can be merged with `else`?

terencefan: Sure, I've done it.

davidfowl:
We shouldn't be updating the types without also updating the spec. We need to treat the protocol itself like part of the product. Changes need to be documented and backwards compatible (since there's no versioning scheme).

PR #672, opened 2019-09-29
Improve ServiceConnectionBase logic 

vicancy:
to remove most dependencies of the semaphore lock

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
+++ b/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
@@ -30,7 +30,7 @@ namespace Microsoft.Azure.SignalR
         private readonly ReadOnlyMemory<byte> _cachedPingBytes;
         private readonly HandshakeRequestMessage _handshakeRequest;
 
-        private readonly SemaphoreSlim _serviceConnectionLock = new SemaphoreSlim(1, 1);
+        private readonly SemaphoreSlim _writeLock = new SemaphoreSlim(1, 1);
 
         private readonly TaskCompletionSource<bool> _serviceConnectionStartTcs = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
 
@@ -39,13 +39,17 @@ namespace Microsoft.Azure.SignalR
         private readonly IServiceMessageHandler _serviceMessageHandler;
         private readonly object _statusLock = new object();
 
+        private volatile string _errorMessage;
+
         // Check service timeout
         private long _lastReceiveTimestamp;
 
         // Keep-alive tick
         private long _lastSendTimestamp;
 
-        private volatile ServiceConnectionStatus _status;
+        private ServiceConnectionStatus _status;
+
+        private int _started;
 
         protected HubServiceEndpoint HubEndpoint { get; }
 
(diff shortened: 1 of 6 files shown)


src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -111,33 +112,65 @@ public async Task StartAsync(string target = null)
...
+                }
+                finally
+                {
+                    // wait until all the connections are cleaned up to close the outgoing pipe
+                    // mark the status as Disconnected so that no one will write to this connection anymore
+                    await _writeLock.WaitAsync();
+                    try
+                    {
+                        // Don't allow write anymore when the connection is disconnected
+                        Status = ServiceConnectionStatus.Disconnected;
```
KKhurin: I wonder if we should change the status before we get the lock - otherwise other writers may pick this connection, wait for their turn to take the lock only to find that the connection is gone. Without any logic to switch the connection at this point it would mean lost messages (while they could have picked a different connection or report an error earlier).

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -111,33 +112,65 @@ public async Task StartAsync(string target = null)
...
+                    finally
+                    {
+                        _writeLock.Release();
+                    }
+                }
             }
-            catch (Exception ex)
+            else
             {
+                _serviceConnectionStartTcs.TrySetResult(false);
```
KKhurin: Do we even use the boolean flag anywhere?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -149,42 +179,39 @@ public Task StopAsync()
...
-            if (ConnectionContext == null)
+            if (Status != ServiceConnectionStatus.Connected)
             {
-                _serviceConnectionLock.Release();
                 throw new ServiceConnectionNotActiveException();
             }
 
+            // We have to lock around outgoing sends since the pipe is single writer.
+            // The lock is per serviceConnection
+            await _writeLock.WaitAsync();
```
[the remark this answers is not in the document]
vicancy: yes, it is in this PR. This PR is focused on cleaning up the logic

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
vicancy: nice catch

---
In reply to: [334720815](https://github.com/Azure/azure-signalr/pull/672#discussion_r334720815) [](ancestors = 334720815)

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
vicancy: no, but I feel OK to keep it this way?

---
In reply to: [334723761](https://github.com/Azure/azure-signalr/pull/672#discussion_r334723761) [](ancestors = 334723761)

PR #679, opened 2019-10-11
fix #677: Management SDK reconnect right after IServiceHubContext dispose

### LINKED ISSUE #677: Hub connection re-opens automatically after dispose called
I'm using the management SDK to call my hub from a functions project. The code effectively looks like this with the appropriate connection string and hub name.

```
public static class MyFunction
    {
        [FunctionName("MyFunction")]
        public static async Task Run([TimerTrigger("0 */10 * * * *")]TimerInfo myTimer, ILogger log, ExecutionContext executionContext)
        {
            var connectionString = "";

            var serviceManager = new ServiceManagerBuilder()
                .WithOptions(option =>
                {
                    option.ServiceTransportType = ServiceTransportType.Persistent;
                    option.ConnectionString = connectionString;
                })
                .Build();

            var hub = await serviceManager.CreateHubContextAsync("myhub");

            await hub.DisposeAsync();
        }
    }
```

When I use the persistent transport type, I can see server connections leaking on Azure. When I drill into an individual transaction I see that immediately after the dispose is called and the transport stopped, it opens again,

image

I also see this behaviour locally with the above code.

realwanpengli:
due to reconnect policy


src/Microsoft.Azure.SignalR.Common/ServiceConnections/ManagementServiceConnectionContainer.cs
```diff
@@ -0,0 +1,45 @@
...
+
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.SignalR.Common.ServiceConnections
+{
+    internal class ManagementServiceConnectionContainer : WeakServiceConnectionContainer
+    {
+        private bool _manuallyStop;
```
zackliu: `volatile` needed? `StartCoreAsync` and `StopAsync` may be invoked in different threads?

test/Microsoft.Azure.SignalR.Tests.Common/TaskExtensions.cs
```diff
@@ -1,10 +1,13 @@
...
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
+using System.Threading;
+using System.Threading.Tasks;
 
-namespace System.Threading.Tasks
+namespace Microsoft.Azure.SignalR.Tests.Common
```
zackliu: Why changing namespace?

test/Microsoft.Azure.SignalR.E2ETests/Management/ServiceHubContextE2EFacts.cs
```diff
@@ -171,9 +171,9 @@ internal async Task SendToConnectionTest(ServiceTransportType serviceTransportTy
             try
             {
                 await RunTestCore(clientEndpoint, clientAccessTokens,
-                    async () => 
+                    async () =>
                     {
-                        var connectionId = await task.OrTimeout();
```
realwanpengli: since conflict @

test/Microsoft.Azure.SignalR.Tests.Common/TaskExtensions.cs
realwanpengli: since https://github.com/Azure/azure-signalr/pull/679/files#r333869040

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ManagementServiceConnectionContainer.cs
realwanpengli: it can be modified in `StopAsync` only, and read by `StartCoreAsync` only

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ManagementServiceConnectionContainer.cs
```diff
@@ -0,0 +1,45 @@
...
+            finally
+            {
+                if (!_manuallyStop)
+                {
+                    await OnConnectionComplete(connection);
+                }
+            }
+        }
+
+        public override Task StopAsync()
```
vicancy: Shall we rethink our current design? I do feel with explicit StopAsync, the container, no matter which, should stop the connection. Cc @zackliu

realwanpengli: Shall we move the reconnect part to connection level instead of container level? Single connection also has this issue.

Add a bool to mark the explicit stop? Or throw a special exception to skip reconnect?

PR #683, opened 2019-10-12
Add helpful fields & Update specs.

terencefan:
Changes are the same as the previous PR and **spec** updates are included in this one.

Changes:

1. Add `MigratableStatus` field to `HandshakeRequestMessage` to indicate if the client connection assigned to this server connection can be migrated to another server connection.
2. Add `SequenceId` field to `ConnectionDataMessage` to help the server-side SDK to guarantee the message order (cause it could be sent from a different signalr-runtime instance in the future)
3. Add `LastSequenceId` field to `CloseConnectionMessage` to acknowledge the **last acked** `SequenceId`, which could help the ASRS to replay the message to the target server.
4. Add `Asrs-Migrated-From` header to `OpenConnectionMessage` to indicate a new client connection is migrated from another server.


specs/ServiceProtocol.md
```diff
@@ -67,15 +67,20 @@ Server will initiate a physical connection to Service, using WebSocket transport
 
 When a new client is connected to Service, a `OpenConnection` message will be sent by Service to Server.
 
+### Client migrated from another server
+
+When a client is migrated from another server, a `OpenConnection` message will be sent by Service to Server, with an `Asrs-Migrated-From` header given.
```
vicancy: >Asrs [](start = 118, length = 4)

`X-ASRS-`?

specs/ServiceProtocol.md
```diff
@@ -67,15 +67,20 @@ Server will initiate a physical connection to Service, using WebSocket transport
...
+
 ### Client Disconnect
 
 - When a client is disconnected from Service, a `CloseConnection` message will be sent by Service to Server.
 - When Server wants to disconnect a client, a `CloseConnection` message will be sent by Server to Service. Then Service will disconnect the phyical connection with the target client.
+	- In case of the `MigratableStatus` of the server connection is `Any`, the service will try to migrate the associated client to another server before dropping the connection.
```
vicancy: >In case of [](start = 3, length = 10)

Move to another section `### Client migrated`, also mention that  this one is transparent to clients but there might be message loss.

specs/ServiceProtocol.md
```diff
@@ -100,6 +105,8 @@ MessagePack uses different formats to encode values. Refer to the [MessagePack F
 ```
 - 1 - Message Type, indicating this is a `HandshakeRequest` message.
 - Version - A `Int32` encoding number of the protocol version.
+- ConnectionType - A `Int32` encoding number of the connection type.
```
vicancy: >encoding [](start = 29, length = 8)

also describe `1,2,3` values in detail

vicancy:
> [5, ConnectionId, ErrorMessage]

also update this

---
Refers to: specs/ServiceProtocol.md:146 in f93858b. [](commit_id = f93858ba95f9c791202b16fe477fc69f1d46681b, deletion_comment = False)

specs/ServiceProtocol.md
```diff
@@ -141,6 +148,7 @@ MessagePack uses different formats to encode values. Refer to the [MessagePack F
 - 5 - Message Type, indicating this is a `CloseConnection` message.
 - ConnectionId - A `String` encoding unique Id of the connection.
 - ErrorMessage - Optional `String` encoding error message.
+- LastSequenceId - A `Int64` encoding number indicates the last acked `ConnectionDataMessage` sequence id.
```
vicancy: >LastSequenceId [](start = 2, length = 14)

naming `LastAckId`?

specs/ServiceProtocol.md
JialinXin: >a [](start = 47, length = 1)

an, and update a few others like an Int32.

specs/ServiceProtocol.md
terencefan: Actually we don't have an `Ack` action.
What we could ensure is we have delivered the message to the application, not `Ack`

specs/ServiceProtocol.md
terencefan: image

PR #689, opened 2019-10-17
Graceful shutdown support

terencefan:
The target of this PR is to ensure all incoming and outgoing packages will be processed before a scheduled shutdown. 

**UPDATE 11/13**
> 1. Add test cases for `HubDispatcher` and `MultiEndpointContainer`.

**UPDATE 11/12**
> 1. Lift the `ShutdownAsync` method up to the `HubDispatcher` level.

**UPDATE 11/06**
> 1. Lift the `ShutdownAsync` method up to the `MultiEndpointServiceConnectionContainer` level.
> 2. Add `OfflineAsync` method to `IServiceConnectionContainer`.

**UPDATE 11/05**
> 1. Move `CloseAsync` method to `ServiceConnectionContainer` level.

**UPDATE 11/01**
> 1. Replace **CloseConnectionMessage** with **PingMessage**.
> 2. Wait on TCS instead of connection **outgoing tasks**.
> 3. Prevent rebalanced (ondemand) connection from starting in the shutdown process. 

**UPDATE 10/31**

> Add test cases for corner cases.

**UPDATE 10/29**

> Add test cases for common scenarios.

**UPDATE 10/28**

> Reimplement `CloseAsync` process by using `TaskCompletionSource` which makes the code more readable.

**UPDATE 10/24**

>After discussed thoroughly with @KKhurin  and @vicancy, we decided **NOT** do anything with the messages during the shutdown process in this scenario. We just remove server connections from the routing table to **prevent new client connections to be assigned to this server** and wait until either client or server to close client connections **by themselves**.

> Since the server-side may not have the ability to notify the client to reconnect, maybe we will introduce a mechanism between server and ASRS to let the server has this ability in the future.

> And there will also a TIMEOUT (by default it's 30s). 
> We simply drop every connection after reaching the time limit.

The `CloseAsync` (for each server connection) contains the following 6 steps:

1. Try acquiring `closeConnectionLock` to ensure at most 1 `closeAsync` in progress.
2. Send a `CloseConnectionMessage` with a `FIN` to `ASRS`
    - Once `ASRS` receives a `FIN`, it will remove the server connection immediately from its route table (to prevent new client connections from assigning to this server connection)
3. Wait until `ASRS`  has sent back a `CloseConnectionMessage` with a `FINACK`.
4. Wait until all `client connections` have stopped.
    - Our `ASRS` will handle the client connections and find a proper time to close them, or in some cases we just let our customer close them by themselves. So we intentionally do nothing on the SDK side during this step.
5. Complete `incoming pipe`
6. Wait until the connection `(StartAsync)` has stopped safely.

image

### DIFF
--- a/src/Microsoft.Azure.SignalR.AspNet/ServerConnections/ServiceConnection.cs
+++ b/src/Microsoft.Azure.SignalR.AspNet/ServerConnections/ServiceConnection.cs
@@ -80,7 +80,7 @@ namespace Microsoft.Azure.SignalR.AspNet
             }
             else
             {
-                // the manager still contains this connectionId, probably this connection is not yet cleaned up 
+                // the manager still contains this connectionId, probably this connection is not yet cleaned up
                 Log.DuplicateConnectionId(Logger, connectionId, null);
                 return WriteAsync(
                     new CloseConnectionMessage(connectionId, $"Duplicate connection ID {connectionId}"));
@@ -319,7 +319,7 @@ namespace Microsoft.Azure.SignalR.AspNet
             public string InstanceId { get; }
 
             public ChannelReader<ServiceMessage> Input { get; }
-            
+
             public ChannelWriter<ServiceMessage> Output { get; }
 
             public IServiceTransport Transport { get; set; }
--- a/src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
+++ b/src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
@@ -18,20 +18,17 @@ namespace Microsoft.Azure.SignalR
         private readonly IMessageRouter _router;
         private readonly ILogger _logger;
         private readonly IServiceConnectionContainer _inner;
-        private readonly IClientConnectionLifetimeManager _clientLifetime;
 
         private IReadOnlyList<HubServiceEndpoint> _endpoints;
 
         public Dictionary<ServiceEndpoint, IServiceConnectionContainer> Connections { get; }
 
-        public MultiEndpointServiceConnectionContainer(string hub, Func<HubServiceEndpoint, IServiceConnectionContainer> generator, IServiceEndpointManager endpointManager, IMessageRouter router, IClientConnectionLifetimeManager lifetime, ILoggerFactory loggerFactory)
+        public MultiEndpointServiceConnectionContainer(string hub, Func<HubServiceEndpoint, IServiceConnectionContainer> generator, IServiceEndpointManager endpointManager, IMessageRouter router, ILoggerFactory loggerFactory)
         {
             if (generator == null)
             {
                 throw new ArgumentNullException(nameof(generator));
             }
-
-            _clientLifetime = lifetime;
 
             _logger = loggerFactory?.CreateLogger<MultiEndpointServiceConnectionContainer>() ?? throw new ArgumentNullException(nameof(loggerFactory));
 
@@ -57,14 +54,12 @@ namespace Microsoft.Azure.SignalR
             IServiceEndpointManager endpointManager,
             IMessageRouter router,
             IServerNameProvider nameProvider,
-            IClientConnectionLifetimeManager lifetime,
             ILoggerFactory loggerFactory
             ) : this(
                 hub,
                 endpoint => CreateContainer(serviceConnectionFactory, endpoint, count, loggerFactory),
                 endpointManager,
                 router,
-                lifetime,
                 loggerFactory
                 )
         {
@@ -131,16 +126,16 @@ namespace Microsoft.Azure.SignalR
             }));
         }
 
-        public async Task ShutdownAsync(TimeSpan timeout)
+        public Task OfflineAsync()
         {
-            // TODOS
-
-            // 1. write FIN to every server connection of every connection container.
-
-            // 2. wait until all client connections have been closed (either by server/client side)
-
-            // 3. stop every container.
-            await StopAsync();
+            if (_inner != null)
+            {
+                return _inner.OfflineAsync();
+            }
+            else
+            {
+                return Task.WhenAll(Connections.Select(c => c.Value.OfflineAsync()));
+            }
         }
 
         public Task WriteAsync(ServiceMessage serviceMessage)
--- a/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
+++ b/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
@@ -183,7 +183,6 @@ namespace Microsoft.Azure.SignalR
             {
                 Log.UnexpectedExceptionInStop(Logger, ConnectionId, ex);
             }
-
             return Task.CompletedTask;
         }
 
@@ -395,7 +394,6 @@ namespace Microsoft.Azure.SignalR
                             {
                                 Log.HandshakeError(Logger, handshakeResponse.ErrorMessage, ConnectionId);
                             }
-
                             return false;
                         }
                     }
@@ -416,6 +414,7 @@ namespace Microsoft.Azure.SignalR
         private async Task ProcessIncomingAsync(ConnectionContext connection)
         {
             var keepAliveTimer = StartKeepAliveTimer();
+
             try
             {
                 while (true)
@@ -468,6 +467,7 @@ namespace Microsoft.Azure.SignalR
             finally
             {
                 keepAliveTimer.Stop();
+                _serviceConnectionOfflineTcs.TrySetResult(true);
             }
         }
 
@@ -625,7 +625,6 @@ namespace Microsoft.Azure.SignalR
 
             private static readonly Action<ILogger, string, Exception> _receivedInstanceOfflinePing =
                 LoggerMessage.Define<string>(LogLevel.Information, new EventId(31, "ReceivedInstanceOfflinePing"), "Received instance offline service ping: {InstanceId}");
-
 
             public static void FailedToWrite(ILogger logger, string serviceConnectionId, Exception exception)
             {
--- a/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
+++ b/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
@@ -18,6 +18,8 @@ namespace Microsoft.Azure.SignalR
         private static TimeSpan ReconnectInterval =>
             TimeSpan.FromMilliseconds(StaticRandom.Next(MaxReconnectBackOffInternalInMilliseconds));
 
+        private static TimeSpan RemoveFromServiceTimeout = TimeSpan.FromSeconds(3);
+
         private readonly BackOffPolicy _backOffPolicy = new BackOffPolicy();
 
         private readonly object _lock = new object();
@@ -31,6 +33,11 @@ namespace Microsoft.Azure.SignalR
         private volatile ServiceConnectionStatus _status;
 
         private volatile bool _terminated = false;
+
+        private static readonly PingMessage _shutdownFinMessage = new PingMessage()
+        {
+            Messages = new string[2] { Constants.ServicePingMessageKey.ShutdownKey, Constants.ServicePingMessageValue.ShutdownFin }
+        };
 
         protected ILogger Logger { get; }
 
@@ -115,12 +122,6 @@ namespace Microsoft.Azure.SignalR
         {
             _terminated = true;
             return Task.WhenAll(FixedServiceConnections.Select(c => c.StopAsync()));
-        }
-
-        public virtual Task ShutdownAsync(TimeSpan timeout)
-        {
-            _terminated = true;
-            return Task.CompletedTask;
         }
 
         /// <summary>
@@ -347,6 +348,30 @@ namespace Microsoft.Azure.SignalR
             {
                 yield return CreateServiceConnectionCore(InitialConnectionType);
             }
+        }
+
+        protected async Task WriteFinAsync(IServiceConnection c)
+        {
+            await c.WriteAsync(_shutdownFinMessage);
+        }
+
+        protected async Task RemoveConnectionFromService(IServiceConnection c)
+        {
+            _ = WriteFinAsync(c);
+
+            var source = new CancellationTokenSource();
+            var task = await Task.WhenAny(c.ConnectionOfflineTask, Task.Delay(RemoveFromServiceTimeout, source.Token));
+            source.Cancel();
+
+            if (task != c.ConnectionOfflineTask)
+            {
+                // log
+            }
+        }
+
+        public virtual Task OfflineAsync()
+        {
+            return Task.WhenAll(FixedServiceConnections.Select(c => RemoveConnectionFromService(c)));
         }
 
         private static class Log
--- a/src/Microsoft.Azure.SignalR.Common/ServiceConnections/StrongServiceConnectionContainer.cs
+++ b/src/Microsoft.Azure.SignalR.Common/ServiceConnections/StrongServiceConnectionContainer.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.SignalR.Protocol;
 using Microsoft.Extensions.Logging;
@@ -42,13 +43,11 @@ namespace Microsoft.Azure.SignalR
             );
         }
 
-        public override Task ShutdownAsync(TimeSpan timeout)
+        public override Task OfflineAsync()
         {
-            var task = base.StopAsync();
-            return Task.WhenAll(
-                task,
-                Task.CompletedTask // TODO
-            );
+            var task1 = base.OfflineAsync();
+            var task2 = Task.WhenAll(_onDemandServiceConnections.Select(c => RemoveConnectionFromService(c)));
+            return Task.WhenAll(task1, task2);
         }
 
         protected override ServiceConnectionStatus GetStatus()
--- a/src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
+++ b/src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Connections;
 using Microsoft.AspNetCore.Http;
@@ -65,9 +66,28 @@ namespace Microsoft.Azure.SignalR
             _ = _serviceConnectionManager.StartAsync();
         }
 
-        public Task Shutdown(TimeSpan timeout)
+        public async Task ShutdownAsync(TimeSpan timeout)
         {
-            return _serviceConnectionManager.ShutdownAsync(timeout);
+            using CancellationTokenSource source = new CancellationTokenSource();
+
+            var expected = OfflineAndWaitForCompletedAsync();
+            var actual = await Task.WhenAny(
+                Task.Delay(timeout, source.Token), expected
+            );
+
+            if (actual != expected)
+            {
+                // TODO log timeout.
+            }
+
+            source.Cancel();
+            await _serviceConnectionManager.StopAsync();
+        }
+
+        private async Task OfflineAndWaitForCompletedAsync()
+        {
+            await _serviceConnectionManager.OfflineAsync();
+            await _clientConnectionManager.WhenAllCompleted();
         }
 
         private IServiceConnectionContainer GetMultiEndpointServiceConnectionContainer(string hub, ConnectionDelegate connectionDelegate, Action<HttpContext> contextConfig = null)
@@ -77,12 +97,11 @@ namespace Microsoft.Azure.SignalR
             serviceConnectionFactory.ConfigureContext = contextConfig;
 
             var factory = new ServiceConnectionContainerFactory(
-                serviceConnectionFactory, 
+                serviceConnectionFactory,
                 _serviceEndpointManager,
-                _router, 
-                _options, 
-                _nameProvider, 
-                _clientConnectionManager,
+                _router,
+                _options,
+                _nameProvider,
                 _loggerFactory
             );
             return factory.Create(hub);
--- a/src/Microsoft.Azure.SignalR/ServerConnections/IServiceConnectionManager.cs
+++ b/src/Microsoft.Azure.SignalR/ServerConnections/IServiceConnectionManager.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.SignalR;
@@ -17,7 +16,7 @@ namespace Microsoft.Azure.SignalR
 
         Task StopAsync();
 
-        Task ShutdownAsync(TimeSpan timeout);
+        Task OfflineAsync();
 
         Task WriteAsync(ServiceMessage seviceMessage);
 
--- a/src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnectionManager.cs
+++ b/src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnectionManager.cs
@@ -2,8 +2,6 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
-using System.Collections.Generic;
-using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.SignalR;
@@ -31,9 +29,9 @@ namespace Microsoft.Azure.SignalR
             return _serviceConnection.StopAsync();
         }
 
-        public Task ShutdownAsync(TimeSpan timeout)
+        public async Task OfflineAsync()
         {
-            return _serviceConnection.ShutdownAsync(timeout);
+            await _serviceConnection.OfflineAsync();
         }
 
         public Task WriteAsync(ServiceMessage serviceMessage)
--- a/src/Microsoft.Azure.SignalR/ServiceOptionsSetup.cs
+++ b/src/Microsoft.Azure.SignalR/ServiceOptionsSetup.cs
@@ -16,6 +16,9 @@ namespace Microsoft.Azure.SignalR
         private readonly string _connectionString;
         private readonly ServiceEndpoint[] _endpoints;
 
+        private readonly bool _gracefulShutdownEnabled = false;
+        private readonly TimeSpan _shutdownTimeout = TimeSpan.FromSeconds(Constants.DefaultShutdownTimeoutInSeconds);
+
         public ServiceOptionsSetup(IConfiguration configuration)
         {
             _appName = configuration[Constants.ApplicationNameDefaultKeyPrefix];
@@ -44,6 +47,8 @@ namespace Microsoft.Azure.SignalR
             options.Endpoints = _endpoints;
             options.ApplicationName = _appName;
             options.ServerStickyMode = _serverStickyMode;
+            options.EnableGracefulShutdown = _gracefulShutdownEnabled;
+            options.ServerShutdownTimeout = _shutdownTimeout;
         }
 
         private static (string, ServiceEndpoint[]) GetEndpoint(IConfiguration configuration, string key)
--- a/src/Microsoft.Azure.SignalR/ServiceRouteBuilder.cs
+++ b/src/Microsoft.Azure.SignalR/ServiceRouteBuilder.cs
@@ -7,6 +7,7 @@ using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Routing;
 using Microsoft.AspNetCore.SignalR;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
 
 namespace Microsoft.Azure.SignalR
 {
@@ -58,6 +59,18 @@ namespace Microsoft.Azure.SignalR
 
             var dispatcher = _serviceProvider.GetRequiredService<ServiceHubDispatcher<THub>>();
             dispatcher.Start(app);
+
+#if NETCOREAPP
+            var lifetime = _serviceProvider.GetService<IHostApplicationLifetime>();
+#elif NETSTANDARD
+            var lifetime = _serviceProvider.GetService<IApplicationLifetime>();
+#else
+            var lifetime = null;
+#endif
+            if (lifetime != null)
+            {
+                // lifetime.ApplicationStopping.Register(() => dispatcher.ShutdownAsync(TimeSpan.FromSeconds(30)).Wait());
+            }
         }
     }
 }
--- a/test/Microsoft.Azure.SignalR.Tests.Common/TestClasses/TestServiceConnection.cs
+++ b/test/Microsoft.Azure.SignalR.Tests.Common/TestClasses/TestServiceConnection.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.IO.Pipelines;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Connections;
@@ -10,13 +11,28 @@ using Microsoft.Extensions.Logging.Abstractions;
 
 namespace Microsoft.Azure.SignalR.Tests.Common
 {
-    internal sealed class TestServiceConnection : ServiceConnectionBase
+    internal class TestServiceConnection : ServiceConnectionBase
     {
         private readonly bool _throws;
+
         private ServiceConnectionStatus _expectedStatus;
+
         private ConnectionContext _connection;
 
-        public TestServiceConnection(ServiceConnectionStatus status = ServiceConnectionStatus.Connected, bool throws = false) : base(null, null, null, null, ServerConnectionType.Default, NullLogger.Instance)
+        public IDuplexPipe Application { get; private set; }
+
+        private TaskCompletionSource<object> _created = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
+
+        public Task ConnectionCreated => _created.Task;
+
+        public TestServiceConnection(ServiceConnectionStatus status = ServiceConnectionStatus.Connected, bool throws = false) : base(
+            new ServiceProtocol(),
+            Guid.NewGuid().ToString(),
+            new HubServiceEndpoint(),
+            null, // TODO replace it with a NullMessageHandler
+            ServerConnectionType.Default,
+            NullLogger.Instance
+        )
         {
             _expectedStatus = status;
             _throws = throws;
@@ -37,6 +53,10 @@ namespace Microsoft.Azure.SignalR.Tests.Common
         {
             var pipeOptions = new PipeOptions();
             var duplex = DuplexPipe.CreateConnectionPair(pipeOptions, pipeOptions);
+
+            Application = duplex.Application;
+            _created.SetResult(null);
+
             return Task.FromResult<ConnectionContext>(new DefaultConnectionContext()
             {
                 Application = duplex.Application,
@@ -69,6 +89,11 @@ namespace Microsoft.Azure.SignalR.Tests.Common
         protected override Task OnMessageAsync(ConnectionDataMessage connectionDataMessage)
         {
             return Task.CompletedTask;
+        }
+
+        protected Task WriteAsyncBase(ServiceMessage serviceMessage)
+        {
+            return base.WriteAsync(serviceMessage);
         }
 
         public override Task WriteAsync(ServiceMessage serviceMessage)
--- a/.gitignore
+++ b/.gitignore
@@ -292,8 +292,8 @@ __pycache__/
 
 .publish/
 
-# vim
-*.swp
-
 # docker
 .docker/
+
+# vim
+*.swp
--- a/src/Microsoft.Azure.SignalR.AspNet/DispatcherHelper.cs
+++ b/src/Microsoft.Azure.SignalR.AspNet/DispatcherHelper.cs
@@ -104,7 +104,7 @@ namespace Microsoft.Azure.SignalR.AspNet
                 configuration.Resolver.Register(typeof(IServiceConnectionFactory), () => scf);
             }
 
-            var sccf = new ServiceConnectionContainerFactory(scf, endpoint, router, options, serverNameProvider, ccm, loggerFactory);
+            var sccf = new ServiceConnectionContainerFactory(scf, endpoint, router, options, serverNameProvider, loggerFactory);
 
             if (hubs?.Count > 0)
             {
--- a/src/Microsoft.Azure.SignalR.AspNet/ServerConnections/ServiceConnectionManager.cs
+++ b/src/Microsoft.Azure.SignalR.AspNet/ServerConnections/ServiceConnectionManager.cs
@@ -82,9 +82,9 @@ namespace Microsoft.Azure.SignalR.AspNet
             return Task.WhenAll(GetConnections().Select(s => s.StopAsync()));
         }
 
-        public Task ShutdownAsync(TimeSpan timeout)
+        public Task OfflineAsync()
         {
-            return Task.WhenAll(GetConnections().Select(s => s.ShutdownAsync(timeout)));
+            return Task.WhenAll(GetConnections().Select(s => s.OfflineAsync()));
         }
 
         public IServiceConnectionContainer WithHub(string hubName)
--- a/src/Microsoft.Azure.SignalR.Common/Interfaces/IServiceConnectionContainer.cs
+++ b/src/Microsoft.Azure.SignalR.Common/Interfaces/IServiceConnectionContainer.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
-using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.SignalR.Protocol;
@@ -15,7 +14,7 @@ namespace Microsoft.Azure.SignalR
 
         Task StopAsync();
 
-        Task ShutdownAsync(TimeSpan timeout);
+        Task OfflineAsync();
 
         Task WriteAsync(ServiceMessage serviceMessage);
 
--- a/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerFactory.cs
+++ b/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerFactory.cs
@@ -13,7 +13,6 @@ namespace Microsoft.Azure.SignalR
         private readonly IServiceEndpointManager _serviceEndpointManager;
         private readonly IMessageRouter _router;
         private readonly IServerNameProvider _nameProvider;
-        private readonly IClientConnectionLifetimeManager _lifetime;
         private readonly IServiceConnectionFactory _serviceConnectionFactory;
 
         public ServiceConnectionContainerFactory(
@@ -22,7 +21,6 @@ namespace Microsoft.Azure.SignalR
         IMessageRouter router,
         IServiceEndpointOptions options,
         IServerNameProvider nameProvider,
-        IClientConnectionLifetimeManager lifetime,
         ILoggerFactory loggerFactory)
         {
             _serviceConnectionFactory = serviceConnectionFactory;
@@ -30,13 +28,12 @@ namespace Microsoft.Azure.SignalR
             _router = router ?? throw new ArgumentNullException(nameof(router));
             _options = options;
             _nameProvider = nameProvider;
-            _lifetime = lifetime;
             _loggerFactory = loggerFactory;
         }
 
         public IServiceConnectionContainer Create(string hub)
         {
-            return new MultiEndpointServiceConnectionContainer(_serviceConnectionFactory, hub, _options.ConnectionCount, _serviceEndpointManager, _router, _nameProvider, _lifetime, _loggerFactory);
+            return new MultiEndpointServiceConnectionContainer(_serviceConnectionFactory, hub, _options.ConnectionCount, _serviceEndpointManager, _router, _nameProvider, _loggerFactory);
         }
     }
 }
--- a/src/Microsoft.Azure.SignalR.Common/ServiceConnections/WeakServiceConnectionContainer.cs
+++ b/src/Microsoft.Azure.SignalR.Common/ServiceConnections/WeakServiceConnectionContainer.cs
@@ -47,8 +47,6 @@ namespace Microsoft.Azure.SignalR.Common.ServiceConnections
             return Task.CompletedTask;
         }
 
-        public override Task ShutdownAsync(TimeSpan timeout) => StopAsync();
-
         public override Task WriteAsync(ServiceMessage serviceMessage)
         {
             if (!_active && !(serviceMessage is PingMessage))
@@ -59,6 +57,11 @@ namespace Microsoft.Azure.SignalR.Common.ServiceConnections
             }
 
             return base.WriteAsync(serviceMessage);
+        }
+
+        public override Task OfflineAsync()
+        {
+            return Task.CompletedTask;
         }
 
         internal bool GetServiceStatus(bool active, int checkWindow, TimeSpan checkTimeSpan)
--- a/test/Microsoft.Azure.SignalR.AspNet.Tests/MultiEndpointServiceConnectionContainerTests.cs
+++ b/test/Microsoft.Azure.SignalR.AspNet.Tests/MultiEndpointServiceConnectionContainerTests.cs
@@ -26,7 +26,7 @@ namespace Microsoft.Azure.SignalR.AspNet.Tests
                                                           IEndpointRouter router,
                                                           ILoggerFactory loggerFactory
 
-                ) : base(hub, generator, endpoint, router, null, loggerFactory)
+                ) : base(hub, generator, endpoint, router, loggerFactory)
             {
             }
         }
--- a/test/Microsoft.Azure.SignalR.AspNet.Tests/TestClasses/TestClientConnectionManager.cs
+++ b/test/Microsoft.Azure.SignalR.AspNet.Tests/TestClasses/TestClientConnectionManager.cs
@@ -1,9 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
 using System.Collections.Concurrent;
-using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Connections;
 using Microsoft.Azure.SignalR.Protocol;
--- a/test/Microsoft.Azure.SignalR.Tests.Common/TestClasses/TestBaseServiceConnectionContainer.cs
+++ b/test/Microsoft.Azure.SignalR.Tests.Common/TestClasses/TestBaseServiceConnectionContainer.cs
@@ -19,10 +19,5 @@ namespace Microsoft.Azure.SignalR.Tests.Common
         {
             return Task.CompletedTask;
         }
-
-        protected override Task OnConnectionComplete(IServiceConnection connection)
-        {
-            return Task.CompletedTask;
-        }
     }
 }
--- a/test/Microsoft.Azure.SignalR.Tests.Common/TestClasses/TestServiceConnectionContainer.cs
+++ b/test/Microsoft.Azure.SignalR.Tests.Common/TestClasses/TestServiceConnectionContainer.cs
@@ -62,7 +62,7 @@ namespace Microsoft.Azure.SignalR.Tests.Common
             return Task.CompletedTask;
         }
 
-        public Task ShutdownAsync(TimeSpan timeout)
+        public Task OfflineAsync()
         {
             return Task.CompletedTask;
         }
--- a/test/Microsoft.Azure.SignalR.Tests/Infrastructure/TestServiceConnectionManager.cs
+++ b/test/Microsoft.Azure.SignalR.Tests/Infrastructure/TestServiceConnectionManager.cs
@@ -57,9 +57,6 @@ namespace Microsoft.Azure.SignalR.Tests
             return Task.CompletedTask;
         }
 
-        public Task ShutdownAsync(TimeSpan timeout)
-        {
-            return Task.CompletedTask;
-        }
+        public Task OfflineAsync() => Task.CompletedTask;
     }
 }
--- a/test/Microsoft.Azure.SignalR.Tests/MultiEndpointServiceConnectionContainerTests.cs
+++ b/test/Microsoft.Azure.SignalR.Tests/MultiEndpointServiceConnectionContainerTests.cs
@@ -28,7 +28,7 @@ namespace Microsoft.Azure.SignalR.Tests
                                                           IEndpointRouter router,
                                                           ILoggerFactory loggerFactory
 
-                ) : base(hub, generator, endpoint, router, null, loggerFactory)
+                ) : base(hub, generator, endpoint, router, loggerFactory)
             {
             }
         }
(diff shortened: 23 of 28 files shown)


src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -454,28 +511,46 @@ private async Task ProcessIncomingAsync(ConnectionContext connection)
...
-                    return OnDisconnectedAsync(closeConnectionMessage);
-                case ConnectionDataMessage connectionDataMessage:
-                    return OnMessageAsync(connectionDataMessage);
-                case ServiceErrorMessage serviceErrorMessage:
-                    return OnServiceErrorAsync(serviceErrorMessage);
-                case PingMessage pingMessage:
-                    return OnPingMessageAsync(pingMessage);
-                case AckMessage ackMessage:
-                    return OnAckMessageAsync(ackMessage);
+            _clientConnectionLock.WaitAsync();
```
vwxyzh: looks useless.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -180,6 +189,43 @@ public Task StopAsync()
             return Task.CompletedTask;
         }
 
+        public async Task CloseAsync()
+        {
+            if (_closeIncomingLock.Wait(0))
+            {
+                // incoming task hasn't started yet so we don't need to wait.
+                return;
```
vwxyzh: release before return, and is it necessary?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -430,7 +480,14 @@ private async Task ProcessIncomingAsync(ConnectionContext connection)
 
                             while (ServiceProtocol.TryParseMessage(ref buffer, out var message))
                             {
-                                _ = DispatchMessageAsync(message);
+                                if (message is CloseConnectionMessage closeConnectionMessage && closeConnectionMessage.ConnectionId == Constants.ConnectionFinAck)
+                                {
+                                    inLoop = false;
```
vwxyzh: `return`?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
terencefan: Yes, the lock should be held by the incoming task and if we got this here, it means the incoming task hasn't started so we could simply return.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
terencefan: You're right, it was once useful and can be dropped now.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
terencefan: I'm not sure what else should be done after the while loop so I can't simply return here.

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
```diff
@@ -60,6 +60,19 @@ protected override Task<ConnectionContext> CreateConnection(string target = null
             return _connectionFactory.ConnectAsync(HubEndpoint, TransferFormat.Binary, ConnectionId, target, headers: CustomHeader);
         }
 
+        protected override Task InnerCloseAsync()
+        {
+            foreach (var key in _clientConnectionManager.ClientConnections.Keys)
```
davidfowl: Just loop over the connections and access the key. Using Keys or Values will make a copy of the list while enumerator the collection directly doesn't (but it takes a snapshot).

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -30,7 +31,11 @@ internal abstract class ServiceConnectionBase : IServiceConnection
         private readonly ReadOnlyMemory<byte> _cachedPingBytes;
         private readonly HandshakeRequestMessage _handshakeRequest;
 
-        private readonly SemaphoreSlim _writeLock = new SemaphoreSlim(1, 1);
+        private readonly SemaphoreSlim _closeConnectionLock = new SemaphoreSlim(1, 1);
```
davidfowl: Why do you need 4 locks? They seem unnecessary.

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
```diff
@@ -60,6 +60,19 @@ protected override Task<ConnectionContext> CreateConnection(string target = null
             return _connectionFactory.ConnectAsync(HubEndpoint, TransferFormat.Binary, ConnectionId, target, headers: CustomHeader);
         }
 
+        protected override Task InnerCloseAsync()
+        {
+            foreach (var key in _clientConnectionManager.ClientConnections.Keys)
+            {
+                var connection = _clientConnectionManager.RemoveClientConnection(key);
+                if (null == connection)
```
vwxyzh: useless?

src/Microsoft.Azure.SignalR/ServiceRouteBuilder.cs
```diff
@@ -58,6 +71,20 @@ public ServiceRouteBuilder(RouteBuilder routes)
 
             var dispatcher = _serviceProvider.GetRequiredService<ServiceHubDispatcher<THub>>();
             dispatcher.Start(app);
+            _shutdownTasks.Add(new Task(() =>
```
vwxyzh: `_shutdownTasks.Add(dispatcher.Shutdown())`?

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
terencefan: That's exactly what I want, I mean to get a copy.

I will remove the connection from the Dict in the loop, and it's usually not a good practice doing this while iterating the original dataset I guess.

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
```diff
@@ -122,6 +137,9 @@ private async Task ProcessOutgoingMessagesAsync(ServiceConnectionContext connect
                         {
                             Log.ErrorSendingMessage(Logger, ex);
                         }
+                    } else if (_outgoingStopped)
```
vwxyzh: code style

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
terencefan: `_closeConnectionLock` to ensure at most one `CloseAsync` process for each connection.
`_incomingLock` and `_outcoming` lock are used for synchronizing the `CloseAsync` process.
`_serverConnectionLock` is used for avoiding collisions during the `WriteAsync` method and it is already there.

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
terencefan: I'll delete these.

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
terencefan: Sorry, this method is no longer useful.
We will send a message to Service with `lastAckId` for each `clientConnection` in the previous proposal and it's not necessary now.

src/Microsoft.Azure.SignalR/ServiceRouteBuilder.cs
```diff
@@ -58,6 +71,20 @@ public ServiceRouteBuilder(RouteBuilder routes)
 
             var dispatcher = _serviceProvider.GetRequiredService<ServiceHubDispatcher<THub>>();
             dispatcher.Start(app);
+            _shutdownTasks.Add(new Task(() =>
+            {
+                Task task = dispatcher.Shutdown();
+                task.Wait();
+            }));
+        }
+
```
vwxyzh: Something like following?
```cs
private readonly List<Action<Task>> _shutdownFuncs = new List<Action<Task>>();
// ...
_shutdownFuncs.Add(dispatcher.Shutdown);
// ...
public void ShutdownAll() => Task.WhenAll(from f in _shutdownFuncs select f()).Wait();
```

davidfowl:
Spoke to @terencefan offline:

- There should be no changes to the route API. Instead there should be a TimeSpan that can be set on [`ServiceOptions`](https://github.com/Azure/azure-signalr/blob/dev/src/Microsoft.Azure.SignalR/ServiceOptions.cs) that let the user control how long to wait for shutdown.
- By default we should Inject the IApplicationLifetime and handle this ourselves.
- There should be no locks, CloseAsync on ServiceConnectionBase should Write the Fin and wait on task returned from `ProcessIncomingAsync`.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
terencefan: Turned the locks into awaiting Tasks after discussion.

KKhurin:
Shouldn't we drain the outgoing pipe as the first step of the whole shutdown process or at least after we dispose the hub and close client connections but before we send the fin?
The reason is the messages queued in the pipe (and the implicit queue inside the semaphore lock) could require acks and it feels wrong to even have a possibility to send any messages after sending fin.

terencefan:
@KKhurin Actually we won't block the outgoing pipe even after sending the `FIN`. 

Once our `ASRS Runtime` receives a `FIN`, it will block the upcoming messages from the **client**, not the **server**. All pending messages in the outgoing pipe would be sent in my proposal.

src/Microsoft.Azure.SignalR/ServiceOptionsSetup.cs
```diff
@@ -35,6 +38,12 @@ public ServiceOptionsSetup(IConfiguration configuration)
 
             _connectionString = connectionString;
             _endpoints = endpoints;
+
+            _gracefulShutdownEnabled = configuration[Constants.ServerGracefulShutdownKey] == "true" ? true : false;
+
+            if (Int32.TryParse(configuration[Constants.ServerGracefulShutdownTimeoutKey], out int timeout)) {
+                _shutdownTimeoutInSeconds = timeout;
```
terencefan: validate the value given by user.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -30,6 +31,11 @@ internal abstract class ServiceConnectionBase : IServiceConnection
         private readonly ReadOnlyMemory<byte> _cachedPingBytes;
         private readonly HandshakeRequestMessage _handshakeRequest;
 
+        private readonly SemaphoreSlim _closeConnectionLock = new SemaphoreSlim(1, 1);
+
+        protected volatile Task _incomingTask = Task.CompletedTask;
```
davidfowl: Is there any change these could be assigned *after* shutting down?

src/Microsoft.Azure.SignalR/ServiceOptions.cs
```diff
@@ -52,6 +52,18 @@ public class ServiceOptions : IServiceEndpointOptions
         /// </summary>
         public ServerStickyMode ServerStickyMode { get; set; } = ServerStickyMode.Disabled;
 
+        /// <summary>
+        /// Specifies if server will shutdown gracefully. 
+        /// Default value is false.
+        /// </summary>
+        public bool EnableGracefulShutdown { get; set; } = false;
```
chenkennt: Shall we make this by default enabled?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -180,6 +190,32 @@ public Task StopAsync()
             return Task.CompletedTask;
         }
 
+        public async Task CloseAsync()
+        {
+            if (!_closeConnectionLock.Wait(0))
```
davidfowl: We never call release?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -180,6 +190,32 @@ public Task StopAsync()
...
+            {
+                // only one close async process can exist at a time.
+                return;
+            }
+
+            try
+            {
+                Console.WriteLine($"{HubEndpoint}: sending fin request.");
```
davidfowl: Remove Console.WriteLine calls and use logging with the appropriate level.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -180,6 +190,32 @@ public Task StopAsync()
...
+                await WriteAsync(new CloseConnectionMessage(Constants.Fin));
+                await _incomingTask;
+                Console.WriteLine($"{HubEndpoint}: fin ack received.");
+            } catch (Exception)
+            {
+                // simply skip the awaiting incoming task step if any error occured.
```
davidfowl: Log the exception as debug.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -430,7 +468,14 @@ private async Task ProcessIncomingAsync(ConnectionContext connection)
 
                             while (ServiceProtocol.TryParseMessage(ref buffer, out var message))
                             {
-                                _ = DispatchMessageAsync(message);
+                                if (message is CloseConnectionMessage closeConnectionMessage && closeConnectionMessage.ConnectionId == Constants.FinAck)
+                                {
+                                    inLoop = false;
+                                    break;
+                                } else
```
davidfowl: nit: Fix the formatting here.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -430,7 +468,14 @@ private async Task ProcessIncomingAsync(ConnectionContext connection)
 
                             while (ServiceProtocol.TryParseMessage(ref buffer, out var message))
                             {
-                                _ = DispatchMessageAsync(message);
+                                if (message is CloseConnectionMessage closeConnectionMessage && closeConnectionMessage.ConnectionId == Constants.FinAck)
+                                {
+                                    inLoop = false;
```
davidfowl: Does this mean the derived type never gets to see the close message? Does that break anything?

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
```diff
@@ -122,6 +122,9 @@ private async Task ProcessOutgoingMessagesAsync(ServiceConnectionContext connect
                         {
                             Log.ErrorSendingMessage(Logger, ex);
                         }
+                    } else if (_shutdown)
```
davidfowl: nit: Fix formatting

src/Microsoft.Azure.SignalR/ServiceOptionsSetup.cs
```diff
@@ -35,6 +38,12 @@ public ServiceOptionsSetup(IConfiguration configuration)
 
             _connectionString = connectionString;
             _endpoints = endpoints;
+
+            _gracefulShutdownEnabled = configuration[Constants.ServerGracefulShutdownKey] == "true" ? true : false;
+
+            if (Int32.TryParse(configuration[Constants.ServerGracefulShutdownTimeoutKey], out int timeout)) {
```
davidfowl: nit: Fix formatting, the format is allman style braces.

src/Microsoft.Azure.SignalR/ServiceOptions.cs
```diff
@@ -52,6 +52,18 @@ public class ServiceOptions : IServiceEndpointOptions
...
+        /// Specifies if server will shutdown gracefully. 
+        /// Default value is false.
+        /// </summary>
+        public bool EnableGracefulShutdown { get; set; } = false;
+
+        /// <summary>
+        /// Specifies the timeout of a graceful shutdown process (in seconds). 
+        /// Default value is 30 seconds.
+        /// </summary>
+        public int ServerShutdownTimeout { get; set; } = 30;
```
davidfowl: This should be a TimeSpan, not an integer.

src/Microsoft.Azure.SignalR/ServiceRouteBuilder.cs
```diff
@@ -57,7 +61,23 @@ public ServiceRouteBuilder(RouteBuilder routes)
...
+#if NETCOREAPP
+            var lifetime = _serviceProvider.GetService<IHostApplicationLifetime>();
+#elif NETSTANDARD
+            var lifetime = _serviceProvider.GetService<IApplicationLifetime>();
+#endif
+
+            if (lifetime != null && dispatcher.EnableGracefulShutdown)
+            {
+                lifetime.ApplicationStopping.Register(() =>
+                    Task.WhenAny(
```
davidfowl: If this doesn't block, it'll end immediately. this should be WaitAny

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
terencefan: We don't need to, The `closeAsync` will be called only once.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
terencefan: There is a double check `ConnectionId` == "finack", we only handle `CloseConnectionMessage` with this feature.

terencefan: It's kinda tricky but we don't want to introduce new fields.
What about your opinion? Shall we introduce a new field like `Type`?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
terencefan: Sadly it would happen.

Considering if a connection hadn't started (running `StartAsync`) while the matching `CloseAsync` was being called. The `WriteAsync` would throw an exception and we have no chance to gracefully stop the `incoming` Task. 

But the odds are very low.

If we want to fully prevent this from happening, we have to introduce a lock to avoid the `StartAsync` and `CloseAsync` run at the same time, which could make our codes even more complex.

Should we do it?

davidfowl: You don’t have to. There are other things that could be done:
1. Do notching, assume it’s rare enough and leave it. I’d the process is being torn down it’s not a big deal but if not, we could leak the tasks.

2. Trigger a cancellation token in close that process input would respect. It means that even if the ProcessIncoming started, it would eventually be cancelled by the token.

3. Create a TaskCompletionSource and trigger it based on the completion of ProcessIncoming and ProcessOutgoing. This way you always have the same reference which just changes state.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
terencefan: I Logged this as a warning, similar to `Start` and `Stop`.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -175,11 +185,34 @@ public Task StopAsync()
...
-            }
-            
+            }
+
             return Task.CompletedTask;
         }
 
+        public async Task CloseAsync()
+        {
+            if (!_closeConnectionLock.Wait(0))
```
BrennanConroy: If it's only ever called once, you don't need the lock

samples/ChatSample/ChatSample/ChatSample.csproj
```diff
@@ -1,11 +1,15 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk.Web">
   <PropertyGroup>
-    <TargetFramework>netcoreapp2.1</TargetFramework>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
     <UserSecretsId>chatsample</UserSecretsId>
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore" Version="$(MicrosoftAspNetCorePackageVersion)" />
```
davidfowl: These added dependencies are unnecessary. Remove them all except for MessagePack

docs/use-signalr-service.md
```diff
@@ -102,6 +102,19 @@ They can be accessed at [`Hub.Context.User`](https://github.com/aspnet/SignalR/b
     - When `Preferred`, Azure SignalR tries to find the app server which the client first negotiates with in a way that no additional cost or global routing is needed. This one can be useful when your scenario is *send to connection*. *Send to connection* can have better performance and lower latency when the sender and the receiver are routed to the same app server.
     - When `Required`, Azure SignalR always tries to find the app server which the client first negotiates with. This options can be useful when some client context is fetched from `negotiate` step and stored in memory, and then to be used inside `Hub`s. However, this option may have performance drawbacks because it requires Azure SignalR to take additional efforts to find this particular app server globally, and to keep globally routing traffics between client and server.
 
+#### `GracefulShutdown`
+- Default value is `Diabled`
```
vicancy: typo

samples/ChatSample/ChatSample/.gitignore
```diff
@@ -0,0 +1 @@
+appsettings.json
```
vicancy: appsettings should not be ignored

docs/use-signalr-service.md
```diff
@@ -102,6 +102,19 @@ They can be accessed at [`Hub.Context.User`](https://github.com/aspnet/SignalR/b
...
+- This option specifies if the server will shutdown gracefully
+- A graceful shutdown is defined by the following rules:
+	- Our SignalR Service will receive a notification and upcoming messages from **client/browser** will be blocked.
```
vicancy: `blocked` -> `ignored`?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
```diff
@@ -116,6 +116,23 @@ public Task StopAsync()
...
+            if (_inner != null)
+            {
+                return _inner.ShutdownAsync();
+            }
+            else
+            {
+                return Task.WhenAll(Connections.Select(s =>
+                {
+                    // Closing connection.
+                    Log.StoppingConnection(_logger, s.Key.Endpoint);
```
vicancy: typo, `ClosingConnection`?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -175,11 +185,34 @@ public Task StopAsync()
...
+            {
+                // only one close async process can exist at a time.
+                return;
```
vicancy: line 124 uses `Interlocked.CompareExchange` to make sure `StartAsync` is called only once, do similar to `CloseAsync`?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -404,9 +437,11 @@ private async Task<bool> ReceiveHandshakeResponseAsync(PipeReader input, Cancell
         private async Task ProcessIncomingAsync(ConnectionContext connection)
         {
             var keepAliveTimer = StartKeepAliveTimer();
+            var inLoop = true;
+
             try
             {
-                while (true)
+                while (inLoop)
```
vicancy: `break` is enough, we don't need `inLoop`?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
vicancy: This is tricky... I'd prefer a new field though...

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -454,27 +497,34 @@ private async Task ProcessIncomingAsync(ConnectionContext connection)
                 }
             }
             finally
-            {
-                keepAliveTimer.Stop();
-            }
+            {
+                keepAliveTimer.Stop();
+                Status = ServiceConnectionStatus.Disconnected;
+                await DisposeConnection(connection);
```
vicancy: This should be removed. It is what I cleaned up with my last PR https://github.com/Azure/azure-signalr/pull/672

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -175,11 +185,34 @@ public Task StopAsync()
...
+                await _incomingTask;
+            }
+            catch (Exception ex)
+            {
+                Log.UnexpectedExceptionInClose(Logger, ConnectionId, ex);
+            }
+
+            // wait until outcoming pipe has been drained.
+            _shutdown = true;
+            await _outgoingTask;
```
vicancy: Here is a good example of how to drain `sending` and `receiving`: https://github.com/aspnet/AspNetCore/blob/master/src/SignalR/clients/csharp/Http.Connections.Client/src/Internal/WebSocketsTransport.cs#L157

src/Microsoft.Azure.SignalR/ServiceOptionsSetup.cs
```diff
@@ -35,6 +38,21 @@ public ServiceOptionsSetup(IConfiguration configuration)
 
             _connectionString = connectionString;
             _endpoints = endpoints;
+
+            _gracefulShutdownEnabled = configuration[Constants.ServerGracefulShutdownKey] == "Enabled" ? true : false;
```
vicancy: How about separating the PR into several, for example, one for adding and reading options, one for adding interface, one for the drain....

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
terencefan: I'll control the process by using `InterLock.Exchange`, I've already learned this is an equivalent of `CAS` in C#.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
terencefan: Is it? So introducing a `type` field might seem good in this senario I guess?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
terencefan: Already replaced by 'Interlock', I just want to prevent it from being called twice or more.

KKhurin: So it is either called once and then we don't need any kind of lock. Or it could be called multiple times and then do the callers always wait for the returned task? If they do I'd like to see us returning the same task for each caller.

terencefan: Since `Interlock.Exchange` is lock-free, I guess it's enough?

src/Microsoft.Azure.SignalR/ServiceOptionsSetup.cs
terencefan: I'd like to remain the compactness of this PR since this feature doesn't have any potential risk I guess?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -175,11 +185,44 @@ public Task StopAsync()
...
 
+        public async Task CloseAsync()
+        {
+            if (Interlocked.CompareExchange(ref _closeStage, ClosingSendingFin1, ClosingDefault) != ClosingDefault)
+            {
+                // CloseAsync is in progress or done.
+                return;
+            }
+
+            while (Interlocked.CompareExchange(ref _closeStage, ClosingWaitingApplicationTasks, ClosingReceivedFinAck1) != ClosingReceivedFinAck1)
```
davidfowl: Why are we doing this instead of waiting on the task?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -180,11 +190,44 @@ public Task StopAsync()
...
+            {
+                await Task.Delay(TimeSpan.FromMilliseconds(50));
+            }
+
+            if (clientOutgoingTasks.Values.Count > 0)
+            {
+                await Task.WhenAll(clientOutgoingTasks.Values);
+            }
+
+            if (_connectionContext != null)
```
davidfowl: What is this for?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
terencefan: Because we won't stop `incomingTask` in the newest proposal (we let our user doing this by themselves, or close connections on the `ASRS` side after connection migration feature has been implemented)

davidfowl: This code seems unnecessarily complex. What's the new flow?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -180,11 +190,44 @@ public Task StopAsync()
...
+            {
+                await _connectionContext.Transport.Output.FlushAsync();
+                _connectionContext.Transport.Output.Complete();
+            }
+        }
+
+        public async Task CloseAsync(TimeSpan timeout)
+        {
+            await Task.WhenAny(
+                Task.Delay(timeout),
```
vicancy: >Delay [](start = 21, length = 5)

pass in cancallationToken

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -180,11 +190,44 @@ public Task StopAsync()
...
+            }
+
+            if (_connectionContext != null)
+            {
+                await _connectionContext.Transport.Output.FlushAsync();
+                _connectionContext.Transport.Output.Complete();
+            }
+        }
+
+        public async Task CloseAsync(TimeSpan timeout)
```
vicancy: >CloseAsync [](start = 26, length = 10)

this one as public is enough?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
terencefan: We're not sure if the buffer is empty even after all outgoing tasks have done.

So we call the `Flush` method explicitly to guarantee all messages can be sent to `ASRS`.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -180,11 +190,44 @@ public Task StopAsync()
...
+            }
+
+            while (Interlocked.CompareExchange(ref _closeStage, ClosingWaitingApplicationTasks, ClosingReceivedFinAck1) != ClosingReceivedFinAck1)
+            {
+                await Task.Delay(TimeSpan.FromMilliseconds(50));
+            }
+
+            if (clientOutgoingTasks.Values.Count > 0)
+            {
+                await Task.WhenAll(clientOutgoingTasks.Values);
```
vicancy: >clientOutgoingTasks [](start = 35, length = 19)

what if it never ends?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
davidfowl: This doesn't look right. We should have a single place that flushes and this should be awaiting that place.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
terencefan: This is the endless version.
There is another method with a timeout `CloseAsync(TimeSpan timeout)` implemented by this method.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
terencefan: You're right. 
The `incomingPipe` is no longer valid after all outgoing tasks are done. 
So I can just Complete the `incomingPipe` and wait the connection stopped automatically.

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnectionContext.cs
```diff
@@ -35,6 +35,13 @@ internal class ServiceConnectionContext : ConnectionContext,
             readerScheduler: PipeScheduler.ThreadPool,
             useSynchronizationContext: false);
 
+        private readonly TaskCompletionSource<int> _connectionEndTcs = new TaskCompletionSource<int>(TaskContinuationOptions.RunContinuationsAsynchronously);
```
davidfowl: What is this and who triggers it?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -459,27 +541,28 @@ private async Task ProcessIncomingAsync(ConnectionContext connection)
...
-                    return OnDisconnectedAsync(closeConnectionMessage);
-                case ConnectionDataMessage connectionDataMessage:
-                    return OnMessageAsync(connectionDataMessage);
-                case ServiceErrorMessage serviceErrorMessage:
-                    return OnServiceErrorAsync(serviceErrorMessage);
-                case PingMessage pingMessage:
-                    return OnPingMessageAsync(pingMessage);
-                case AckMessage ackMessage:
-                    return OnAckMessageAsync(ackMessage);
+        {
```
davidfowl: Is this just a formatting change?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -406,13 +474,27 @@ private async Task<bool> ReceiveHandshakeResponseAsync(PipeReader input, Cancell
             }
         }
 
+        protected virtual void ProcessMessage(ServiceMessage message)
+        {
+            if (message is PingMessage ping && ping.TryGetValue(Constants.ShutdownState, out string val))
+            {
+                if (val == Constants.FinAck)
+                {
+                    _serviceConnectionRemoveTcs.TrySetResult(true);
```
davidfowl: Why does this trigger the task? Shouldn't it just end the loop?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
terencefan: I'm afraid so, I haven't touched these codes.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
terencefan: No, we still have to wait for our `ASRS` to close the connection in a proper time.
What we can know from receiving a "FINACK" is no more new client connections will be assigned to this connection (Because our `ASRS` will remove this connection from its routing table).

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
terencefan: I'm afraid so, I haven't touched these.

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnectionContext.cs
terencefan: Sorry, I was working on another PR split from this one these days and failed to review this PR though I updated it several times.

I've updated the code, it will be set after the client connection completed.

src/Microsoft.Azure.SignalR/ServiceOptionsSetup.cs
```diff
@@ -35,6 +38,24 @@ public ServiceOptionsSetup(IConfiguration configuration)
 
             _connectionString = connectionString;
             _endpoints = endpoints;
+
+            if (configuration[Constants.ServerGracefulShutdownKey] == "Enabled")
+            {
+                _gracefulShutdownEnabled = true;
+            }
```
JialinXin: Get section value to check whether enabled?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -406,13 +476,25 @@ private async Task<bool> ReceiveHandshakeResponseAsync(PipeReader input, Cancell
             }
         }
 
+        protected virtual void ProcessMessage(ServiceMessage message)
+        {
+            if (message is PingMessage ping && ping.TryGetValue(Constants.ServicePingMessageKey.ShutdownKey, out string val) && val == Constants.ServicePingMessageVal.ShutdownFinAck)
+            {
+                _serviceConnectionRemoveTcs.TrySetResult(true);
+            } else
```
JialinXin: > else [](start = 13, length = 5)

new line

src/Microsoft.Azure.SignalR.AspNet/ServerConnections/ServiceConnection.cs
```diff
@@ -290,8 +290,13 @@ private string GetInstanceId(IDictionary<string, StringValues> header)
                 return instanceId;
             }
             return null;
-        }
-
+        }
+
+        protected override Task ClientConnectionsCompletedTask()
+        {
+            throw new NotImplementedException();
```
terencefan: WIP

samples/ChatSample/ChatSample.CoreApp3/appsettings.json
```diff
@@ -1,8 +1,7 @@
 {
   "Azure": {
     "SignalR": {
-      "Enabled": true,
```
vicancy: >Enabled [](start = 7, length = 7)

remove unnecessary changes

docs/use-signalr-service.md
```diff
@@ -102,6 +102,19 @@ They can be accessed at [`Hub.Context.User`](https://github.com/aspnet/SignalR/b
     - When `Preferred`, Azure SignalR tries to find the app server which the client first negotiates with in a way that no additional cost or global routing is needed. This one can be useful when your scenario is *send to connection*. *Send to connection* can have better performance and lower latency when the sender and the receiver are routed to the same app server.
     - When `Required`, Azure SignalR always tries to find the app server which the client first negotiates with. This options can be useful when some client context is fetched from `negotiate` step and stored in memory, and then to be used inside `Hub`s. However, this option may have performance drawbacks because it requires Azure SignalR to take additional efforts to find this particular app server globally, and to keep globally routing traffics between client and server.
 
+#### `GracefulShutdown`
```
vicancy: >GracefulShutdown [](start = 6, length = 16)

move to next PR when the feature is ready

src/Microsoft.Azure.SignalR/ServiceRouteBuilder.cs
```diff
@@ -58,6 +59,17 @@ public ServiceRouteBuilder(RouteBuilder routes)
 
             var dispatcher = _serviceProvider.GetRequiredService<ServiceHubDispatcher<THub>>();
             dispatcher.Start(app);
+
+#if NETCOREAPP
+            var lifetime = _serviceProvider.GetService<IHostApplicationLifetime>();
+#elif NETSTANDARD
+            var lifetime = _serviceProvider.GetService<IApplicationLifetime>();
+#endif
```
vicancy: >endif [](start = 1, length = 5)

otherwise throw?

src/Microsoft.Azure.SignalR/ServiceRouteBuilder.cs
```diff
@@ -58,6 +59,17 @@ public ServiceRouteBuilder(RouteBuilder routes)
...
             var dispatcher = _serviceProvider.GetRequiredService<ServiceHubDispatcher<THub>>();
             dispatcher.Start(app);
+
+#if NETCOREAPP
+            var lifetime = _serviceProvider.GetService<IHostApplicationLifetime>();
+#elif NETSTANDARD
+            var lifetime = _serviceProvider.GetService<IApplicationLifetime>();
+#endif
+
+            if (lifetime != null && true)
```
vicancy: >true [](start = 36, length = 4)

?

src/Microsoft.Azure.SignalR/ServiceRouteBuilder.cs
```diff
@@ -58,6 +59,17 @@ public ServiceRouteBuilder(RouteBuilder routes)
...
+            {
+                lifetime.ApplicationStopping.Register(() => dispatcher.Shutdown(TimeSpan.FromSeconds(30)).Wait());
```
vicancy: >Shutdown [](start = 71, length = 8)

pass in the timeout option

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
```diff
@@ -299,6 +301,11 @@ private string GetInstanceId(IDictionary<string, StringValues> header)
                 return instanceId;
             }
             return string.Empty;
-        }
+        }
+
+        protected override Task ClientConnectionsCompletedTask()
+        {
+            return Task.WhenAll(_clientConnectionManager.ClientConnections.Values.Select(c => c.CompleteTask));
```
vicancy: >_clientConnectionManager [](start = 32, length = 24)

this is for connections clients in this app server not for this service connection, `_connectionIds` lives connections for this service connection

terencefan: So it's a global manager? Singleton?

Well, seems like it's an inherent barrier we've talked in the previous PR.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -180,8 +186,7 @@ public Task StopAsync()
             {
                 Log.UnexpectedExceptionInStop(Logger, ConnectionId, ex);
             }
-
-            return Task.CompletedTask;
+            await _serviceConnectionEndTcs.Task;
```
terencefan: Add a case to test if `StopAsync` could quit successfully when it failed to start.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
```diff
@@ -131,25 +131,23 @@ public Task StopAsync()
...
 
-        public async Task ShutdownAsync(TimeSpan timeout)
+        public Task OfflineAsync()
         {
-            // TODOS
+            if (_inner != null)
+            {
+                return _inner.OfflineAsync();
+            } else
+            {
```
vicancy: new line

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
```diff
@@ -131,25 +131,23 @@ public Task StopAsync()
...
+                return Task.WhenAll(Connections.Values.Select(c => c.OfflineAsync()));
```
vicancy: 'Connections.Select(c=>c.Value......)' instead, 'Values' have a global lock

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
```diff
@@ -131,25 +131,23 @@ public Task StopAsync()
...
-        }
-
         public Task WriteAsync(ServiceMessage serviceMessage)
         {
             if (_inner != null)
             {
                 return _inner.WriteAsync(serviceMessage);
             }
-
             return WriteMultiEndpointMessageAsync(serviceMessage, connection => connection.WriteAsync(serviceMessage));
```
vicancy: keep the line. as the convention, leave an extra line after the closing }

src/Microsoft.Azure.SignalR.Common/ServiceConnections/StrongServiceConnectionContainer.cs
```diff
@@ -33,22 +33,20 @@ public override Task HandlePingAsync(PingMessage pingMessage)
             return Task.CompletedTask;
         }
 
-        public override Task StopAsync()
+        public override async Task StopAsync()
         {
             var task = base.StopAsync();
-            return Task.WhenAll(
+            await Task.WhenAll(
```
vicancy: return is enough, await creates more

src/Microsoft.Azure.SignalR.Common/ServiceConnections/StrongServiceConnectionContainer.cs
```diff
@@ -33,22 +33,20 @@ public override Task HandlePingAsync(PingMessage pingMessage)
...
 
-        public override Task ShutdownAsync(TimeSpan timeout)
+        public override Task OfflineAsync()
         {
-            var task = base.StopAsync();
-            return Task.WhenAll(
-                task,
-                Task.CompletedTask // TODO
-            );
+            var task1 = base.OfflineAsync();
```
vicancy: more meaningful naming

src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
```diff
@@ -65,9 +65,21 @@ public void Start(ConnectionDelegate connectionDelegate, Action<HttpContext> con
             _ = _serviceConnectionManager.StartAsync();
         }
 
-        public Task Shutdown(TimeSpan timeout)
+        private async Task OfflineAndWaitForCompletedAsync()
         {
-            return _serviceConnectionManager.ShutdownAsync(timeout);
+            await Task.WhenAny(
+                _serviceConnectionManager.OfflineAsync(),
+                Task.Delay(TimeSpan.FromSeconds(3))
```
vicancy: provide a cancellation token for the delay

src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
```diff
@@ -65,9 +65,21 @@ public void Start(ConnectionDelegate connectionDelegate, Action<HttpContext> con
...
         {
-            return _serviceConnectionManager.ShutdownAsync(timeout);
+            await Task.WhenAny(
+                _serviceConnectionManager.OfflineAsync(),
+                Task.Delay(TimeSpan.FromSeconds(3))
+            );
+            await _clientConnectionManager.WhenAllCompleted();
+        }
+
+        public async Task ShutdownAsync(TimeSpan timeout)
```
vicancy: public methods above private ones

src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
vicancy: and move the timeout check logic into OfflineAsync?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/StrongServiceConnectionContainer.cs
terencefan: I think everyone could understand that I just want to combine these 2 tasks and waiting for all of them to be completed.

src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
terencefan: Don't feel like to add a `timeout` param for the `OfflineAsync` method.

src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
```diff
@@ -65,9 +65,18 @@ public void Start(ConnectionDelegate connectionDelegate, Action<HttpContext> con
             _ = _serviceConnectionManager.StartAsync();
         }
 
-        public Task Shutdown(TimeSpan timeout)
+        public async Task ShutdownAsync(TimeSpan timeout)
         {
-            return _serviceConnectionManager.ShutdownAsync(timeout);
+            await Task.WhenAny(
+                Task.Delay(timeout), OfflineAndWaitForCompletedAsync()
```
vicancy: same here, cancellation token for `Task.Delay`

vicancy: same here, cancellation token for Task.Delay

vicancy: Also add log when timeout reaches

src/Microsoft.Azure.SignalR/ServiceRouteBuilder.cs
```diff
@@ -58,6 +59,18 @@ public ServiceRouteBuilder(RouteBuilder routes)
...
+#if NETCOREAPP
+            var lifetime = _serviceProvider.GetService<IHostApplicationLifetime>();
+#elif NETSTANDARD
+            var lifetime = _serviceProvider.GetService<IApplicationLifetime>();
+#else
+            var lifetime = null;
+#endif
+            if (lifetime != null)
+            {
+                lifetime.ApplicationStopping.Register(() => dispatcher.ShutdownAsync(TimeSpan.FromSeconds(30)).Wait());
```
vicancy: >TimeSpan.FromSeconds(30) [](start = 85, length = 24)

timeout option is not respected?

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnectionManager.cs
```diff
@@ -31,9 +29,11 @@ public Task StopAsync()
             return _serviceConnection.StopAsync();
         }
 
-        public Task ShutdownAsync(TimeSpan timeout)
+        public async Task OfflineAsync()
         {
-            return _serviceConnection.ShutdownAsync(timeout);
+            using CancellationTokenSource source = new CancellationTokenSource();
+            source.CancelAfter(TimeSpan.FromSeconds(3));
```
vicancy: move to a readonly static field.
Also we need logs when timeout

test/Microsoft.Azure.SignalR.Tests.Common/TestClasses/TestServiceConnection.cs
```diff
@@ -10,13 +11,31 @@
...
         private readonly bool _throws;
+
         private ServiceConnectionStatus _expectedStatus;
+
         private ConnectionContext _connection;
 
-        public TestServiceConnection(ServiceConnectionStatus status = ServiceConnectionStatus.Connected, bool throws = false) : base(null, null, null, null, ServerConnectionType.Default, NullLogger.Instance)
+        public IDuplexPipe Application { get; private set; }
+
+        private TaskCompletionSource<int> _created = new TaskCompletionSource<int>(TaskContinuationOptions.RunContinuationsAsynchronously);
```
zackliu: TaskCreationOptions.RunContinuationsAsynchronously

src/Microsoft.Azure.SignalR/ServiceRouteBuilder.cs
terencefan: `docs` and `options` will be in the final PR

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -347,8 +341,36 @@ private IEnumerable<IServiceConnection> CreateFixedServiceConnection(int count)
...
+            };
+            await c.WriteAsync(ping);
+        }
+
+        protected async Task RemoveConnectionFromService(IServiceConnection c)
+        {
+            _ = WriteFinAsync(c);
+
+            var source = new CancellationTokenSource();
+            var task = await Task.WhenAny(c.ConnectionOfflineTask, Task.Delay(TimeSpan.FromSeconds(3), source.Token));
```
vicancy: move `TimeSpan.FromSeconds(3)` to a local field

src/Microsoft.Azure.SignalR/ServiceRouteBuilder.cs
vicancy: then remove this line. Otherwise when your PR is in, the SDK contains this feature?

test/Microsoft.Azure.SignalR.Tests.Common/TestClasses/TestServiceConnection.cs
```diff
@@ -10,13 +11,31 @@
...
         private readonly bool _throws;
+
         private ServiceConnectionStatus _expectedStatus;
+
         private ConnectionContext _connection;
 
-        public TestServiceConnection(ServiceConnectionStatus status = ServiceConnectionStatus.Connected, bool throws = false) : base(null, null, null, null, ServerConnectionType.Default, NullLogger.Instance)
+        public IDuplexPipe Application { get; private set; }
+
+        private TaskCompletionSource<int> _created = new TaskCompletionSource<int>(TaskCreationOptions.RunContinuationsAsynchronously);
```
vicancy: `TaskCompletionSource<object>` and `SetResult(null)` is a common practice

src/Microsoft.Azure.SignalR/ServiceRouteBuilder.cs
terencefan: simply disable the feature could be better?

src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
```diff
@@ -65,9 +66,28 @@ public void Start(ConnectionDelegate connectionDelegate, Action<HttpContext> con
...
 
-        public Task Shutdown(TimeSpan timeout)
+        public async Task ShutdownAsync(TimeSpan timeout)
         {
-            return _serviceConnectionManager.ShutdownAsync(timeout);
+            using CancellationTokenSource source = new CancellationTokenSource();
+
+            var expected = OfflineAndWaitForCompletedAsync();
+            var actual = await Task.WhenAny(
+                Task.Delay(timeout, source.Token), OfflineAndWaitForCompletedAsync()
```
vicancy: replace with `expected`

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -347,8 +343,36 @@ private IEnumerable<IServiceConnection> CreateFixedServiceConnection(int count)
             {
                 yield return CreateServiceConnectionCore(InitialConnectionType);
             }
-        }
-
+        }
+
+        protected async Task WriteFinAsync(IServiceConnection c)
+        {
+            var ping = new PingMessage()
```
vicancy: move to local field?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -347,8 +348,32 @@ private IEnumerable<IServiceConnection> CreateFixedServiceConnection(int count)
...
+        {
+            _ = WriteFinAsync(c);
+
+            var source = new CancellationTokenSource();
+            var task = await Task.WhenAny(c.ConnectionOfflineTask, Task.Delay(RemoveFromServiceTimeout, source.Token));
+            source.Cancel();
+
+            if (task != c.ConnectionOfflineTask)
+            {
+                // log
```
davidfowl: Why didn't this get fixed in the PR?

src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
```diff
@@ -65,9 +66,28 @@ public void Start(ConnectionDelegate connectionDelegate, Action<HttpContext> con
...
+            using CancellationTokenSource source = new CancellationTokenSource();
+
+            var expected = OfflineAndWaitForCompletedAsync();
+            var actual = await Task.WhenAny(
+                Task.Delay(timeout, source.Token), expected
+            );
+
+            if (actual != expected)
+            {
+                // TODO log timeout.
```
davidfowl: Here as well.

src/Microsoft.Azure.SignalR/ServerConnections/IServiceConnectionManager.cs
```diff
@@ -17,7 +16,7 @@ internal interface IServiceConnectionManager<THub> where THub : Hub
 
         Task StopAsync();
 
-        Task ShutdownAsync(TimeSpan timeout);
+        Task OfflineAsync();
```
davidfowl: OfflineAsync isn't a verb. MarkOfflineAsync? MakeOfflineAsync?

PR #702, opened 2019-10-28
Add docker files for ChatSample

terencefan:
And add some non-important files and options.

Split from #689


JialinXin:
There's a chat sample targeting .Net Core 3.0 app. https://github.com/Azure/azure-signalr/tree/dev/samples/ChatSample/ChatSample.CoreApp3. We may leave this for targeting 2.0 app or remove one if both targeting 3.0.

terencefan:
@JialinXin Ok I finally understand what the `CoreApp3` is for...

samples/ChatSample/ChatSample.CoreApp3/Chat.cs
```diff
@@ -1,6 +1,10 @@
-﻿using Microsoft.AspNetCore.SignalR;
-using System;
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using Microsoft.AspNetCore.SignalR;
```
vicancy: system usings move to top

samples/ChatSample/ChatSample.CoreApp3/Properties/launchSettings.json
```diff
@@ -23,6 +23,13 @@
         "ASPNETCORE_ENVIRONMENT": "Development",
         "ASPNETCORE_HOSTINGSTARTUPASSEMBLIES": "Microsoft.Azure.SignalR"
```
JialinXin: >"ASPNETCORE_HOSTINGSTARTUPASSEMBLIES": "Microsoft.Azure.SignalR" [](start = 8, length = 64)

You can add this configuration in below to skip the call `AddAzureSignalR()` in Startup.cs. Otherwise, it will use local SignalR. It's available in NetCoreApp3.0 only.

PR #707, opened 2019-10-29
fix auto reconnect after stop bug

realwanpengli:

PR #709, opened 2019-10-29
Graceful shutdown controlling process.

terencefan:
Split from #689 

Contain only `ControllingProcess` for graceful shutdown, a few new `Constants` and 2 new `Options`

**UPDATE 11/1**

@vicancy @KKhurin @vwxyzh 

After discussed with @chenkennt, we found a **barrier** between **all client connection closed** and **close server connections** is necessary.

Considering if our SDK connected with multiple `Endpoints`, and it would like to broadcast to all `Endpoints` after receiving a message. In my current workflow, I shutdown the `Endpoints` respectively, which means if all client connections from `Endpoint A`  have been dropped, the relevant server connection will be dropped immediately. And if our SDK received a message from `Endpoint B`, it wouldn't have any chance to broadcast messages to `Endpoint A`.

So maybe I have to introduce a **barrier** to handle this situation, and the workflow may like this:

image

By doing this, we can make sure our server connections to each endpoint will stay alive until even our SDK has received the last message from its last client connection. 

**Is this useful? Please give me your advice.**

And here is the overall process:

image

The `CloseAsync` method will be implemented in the main PR.


src/Microsoft.Azure.SignalR.Common/Interfaces/IServiceConnection.cs
```diff
@@ -15,6 +15,10 @@ internal interface IServiceConnection
 
         Task StopAsync();
 
+        Task CloseAsync();
```
vicancy: >CloseAsync [](start = 13, length = 10)

have the below one is enough?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
```diff
@@ -115,6 +115,22 @@ public Task StopAsync()
...
+        {
+            if (_inner != null)
+            {
+                return _inner.ShutdownAsync(timeout);
+            }
+            else
+            {
+                return Task.WhenAll(Connections.Select(s =>
+                {
+                    Log.StoppingConnection(_logger, s.Key.Endpoint);
```
vicancy: >StoppingConnection [](start = 24, length = 18)

ClosingConnection?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/StrongServiceConnectionContainer.cs
```diff
@@ -35,8 +35,17 @@ public override Task HandlePingAsync(PingMessage pingMessage)
...
         {
-            var stopOndemandConnectionsTask =  Task.WhenAll(_onDemandServiceConnections.Select(c => c.StopAsync()));
-            return Task.WhenAll(stopOndemandConnectionsTask, base.StopAsync());
+            var tasksForOndemandConnections = Task.WhenAll(_onDemandServiceConnections.Select(c => c.StopAsync()));
+            return Task.WhenAll(tasksForOndemandConnections, base.StopAsync());
+        }
+
+        public override Task ShutdownAsync(TimeSpan timeout)
+        {
+            _shutdown = true;
```
vicancy: >_shutdown [](start = 12, length = 9)

only base class to set this, maybe for base class to `virtual ShutdownCore` instead

src/Microsoft.Azure.SignalR.Common/ServiceConnections/WeakServiceConnectionContainer.cs
```diff
@@ -47,6 +47,8 @@ public override Task HandlePingAsync(PingMessage pingMessage)
             return Task.CompletedTask;
         }
 
+        public override Task ShutdownAsync(TimeSpan timeout) => StopAsync();
```
vicancy: >StopAsync [](start = 64, length = 9)

I am a little confused about the naming now....

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -150,6 +155,11 @@ protected IServiceConnection CreateServiceConnectionCore(ServerConnectionType ty
 
         protected virtual async Task OnConnectionComplete(IServiceConnection serviceConnection)
         {
+            if (_shutdown)
```
vicancy: >_shutdown [](start = 16, length = 9)

Consider adding test for this

src/Microsoft.Azure.SignalR/ServiceOptions.cs
```diff
@@ -52,6 +52,18 @@ public class ServiceOptions : IServiceEndpointOptions
         /// </summary>
         public ServerStickyMode ServerStickyMode { get; set; } = ServerStickyMode.Disabled;
 
+        /// <summary>
+        /// Specifies if server will shutdown gracefully. 
+        /// Default value is false.
```
davidfowl: ?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/WeakServiceConnectionContainer.cs
terencefan: There is nothing to do with the `WeekConnection` while shutting down it (no active client connection attached to it), so we simply reuse the `StopAsync`.

src/Microsoft.Azure.SignalR/ServiceOptions.cs
terencefan: fixed

src/Microsoft.Azure.SignalR.Common/Interfaces/IServiceConnection.cs
```diff
@@ -15,6 +15,8 @@ internal interface IServiceConnection
 
         Task StopAsync();
 
+        Task CloseAsync(TimeSpan timeout);
```
vicancy: Is it Container level decision to `Shutdown`? and for the inner connection, it can just provide a method to `wait for all complete`. Looks like it is not neccessory for the inner connection to have a `Close` logic

vwxyzh:
In my options:
* `ShutdownAsync` should be `StopAsync`.
* Merge with #707

src/Microsoft.Azure.SignalR.Common/Constants.cs
```diff
@@ -11,6 +11,14 @@ internal static class Constants
         public const string ConnectionStringDefaultKey = "Azure:SignalR:ConnectionString";
         public const string ApplicationNameDefaultKey = "Azure:SignalR:ApplicationName";
 
+        public const string ServerGracefulShutdownKey = "Azure:SignalR:GracefulShutdown";
+        public const string ServerGracefulShutdownTimeoutKey = "Azure:SignalR:GracefulShutdownTimeout";
```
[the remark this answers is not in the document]
terencefan: But our user cannot change `ServiceOptions` directly,
They could only give us `json` config file and we need to parse options from it.
And I don't feel like to write a "magic string" in our codes.

src/Microsoft.Azure.SignalR.Common/Constants.cs
```diff
@@ -11,6 +11,14 @@ internal static class Constants
...
+
+        public const int DefaultShutdownTimeoutInSeconds = 30;
+
+        public const string Fin = "fin";
```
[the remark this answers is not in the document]
terencefan: In the future PR, even if we switched to `PingMessage` instead of `CloseMessage`, we still need it.

src/Microsoft.Azure.SignalR.Common/Constants.cs
```diff
@@ -11,6 +11,14 @@ internal static class Constants
         public const string ConnectionStringDefaultKey = "Azure:SignalR:ConnectionString";
         public const string ApplicationNameDefaultKey = "Azure:SignalR:ApplicationName";
 
+        public const string ServerGracefulShutdownKey = "Azure:SignalR:GracefulShutdown";
```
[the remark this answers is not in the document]
terencefan: Maybe we can enable it in the future PR since this PR is just a part of our entire solution?

src/Microsoft.Azure.SignalR.Common/Interfaces/IServiceConnection.cs
terencefan: The problem is we hold `IServiceConnection` in our `ServiceConnectionContainer` so I have no other choice except for adding a method to the interface. Reflection/Conversion could also achieve this but it's not a good choice I guess.

terencefan:
@vwxyzh 
There are still small differences between these 2 methods, 
For `StopAsync`, we cancel the pending messages in the pipe.
For `CloseAsync`, we simply send a message to our runtime, but nothing to do with our pipes.

src/Microsoft.Azure.SignalR.Common/Constants.cs
```diff
@@ -66,6 +68,13 @@ public static class ServicePingMessageKey
         {
             public const string RebalanceKey = "target";
             public const string OfflineKey = "offline";
+            public const string ShutdownKey = "_shutdown";
```
vicancy: follow the same naming pattern as others?

src/Microsoft.Azure.SignalR.Common/Constants.cs
```diff
@@ -66,6 +68,13 @@ public static class ServicePingMessageKey
...
+        }
+
+        public static class ServicePingMessageVal
```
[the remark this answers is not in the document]
terencefan: I think it's not necessary to create `class` for every header of `ServicePingMessage`, we can differentiate them by prefixes.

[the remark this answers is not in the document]
vicancy: full name `Value`

src/Microsoft.Azure.SignalR.Common/ServiceConnections/StrongServiceConnectionContainer.cs
```diff
@@ -35,10 +35,17 @@ public override Task HandlePingAsync(PingMessage pingMessage)
...
         public override Task StopAsync()
         {
-            var stopOndemandConnectionsTask =  Task.WhenAll(_onDemandServiceConnections.Select(c => c.StopAsync()));
+            var stopOndemandConnectionsTask = Task.WhenAll(_onDemandServiceConnections.Select(c => c.StopAsync()));
             return Task.WhenAll(stopOndemandConnectionsTask, base.StopAsync());
         }
 
+        public override Task ShutdownAsync(TimeSpan timeout)
+        {
+            StartShutdown();
```
vicancy: no need to expose `StartShutdown`? call `base.ShutdownAsync` is enough?

src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
```diff
@@ -27,7 +28,18 @@ internal class ServiceHubDispatcher<THub> where THub : Hub
         private readonly string _hubName;
         private readonly IServerNameProvider _nameProvider;
 
-        public ServiceHubDispatcher(IServiceProtocol serviceProtocol,
+        internal bool EnableGracefulShutdown
```
vicancy: why adding properties to the Dispatcher? Is it possible to fetch options from `_options` directly when needed?

test/Microsoft.Azure.SignalR.Tests/TestServiceConnectionContainer.cs
```diff
@@ -0,0 +1,78 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Microsoft.Azure.SignalR.Protocol;
+using Microsoft.Extensions.Logging.Abstractions;
+
+namespace Microsoft.Azure.SignalR.Tests
+{
+    internal sealed class TestServiceConnectionContainer : ServiceConnectionContainerBase
```
vicancy: We have `TestServiceConnectionContainer` in test common project, can that be leveraged?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/StrongServiceConnectionContainer.cs
terencefan: `StartShutdown` is used by both `Shutdown` and `StopAsync`.

src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
terencefan: No, we can't, pls check how I use these properties in the related PR.

test/Microsoft.Azure.SignalR.Tests/TestServiceConnectionContainer.cs
terencefan: I'm afraid they are for different scenarios.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -31,7 +30,7 @@ internal abstract class ServiceConnectionContainerBase : IServiceConnectionConta
 
         private volatile ServiceConnectionStatus _status;
 
-        private volatile bool _terminated = false;
+        protected volatile bool _terminated = false;
```
vicancy: >protected [](start = 8, length = 9)

private is enough?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -110,23 +109,32 @@ private set
...
+
         /// <summary>
         /// Start and manage the whole connection lifetime
         /// </summary>
         /// <returns></returns>
         protected async Task StartCoreAsync(IServiceConnection connection, string target = null)
         {
+            if (_terminated)
+            {
+                await Task.CompletedTask;
```
vicancy: no need

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -162,11 +170,6 @@ protected virtual async Task OnConnectionComplete(IServiceConnection serviceConn
                 throw new ArgumentNullException(nameof(serviceConnection));
             }
 
-            if (_terminated)
```
vicancy: >_terminated [](start = 16, length = 11)

why remove?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
terencefan: removed

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
terencefan: No, we can't simply call `base.ShutdownAsync` since `StrongContainer` has its own `on-demand` connections and needs to be stopped after the `_terminated` flag to be set to `true`.

test/Microsoft.Azure.SignalR.Tests/ServiceConnectionContainerBaseTests.cs
```diff
@@ -0,0 +1,70 @@
...
+            {
+                return new SimpleTestServiceConnection();
+            }
+        }
+
+        private sealed class SimpleTestServiceConnection : IServiceConnection
+        {
+            public ServiceConnectionStatus Status { get; set; }
+
+            public Task ConnectionInitializedTask => Task.Delay(TimeSpan.FromSeconds(1));
```
vicancy: Delay 1 sec is too long, maybe 10 milliseconds?

terencefan: Actually we don't have to wait in the current implementation since our `StartCoreAsync` doesn't truly await anything before calling inner `StartAsync`. 

But what if it does something that needs to be awaited, and, won't cost too long in future implementation? (like write a log to file, cause real await but very short)

That's why I introduced a 1s delay in this test case. **1s for 1 test case is fair**, to strike a balance between robust and time-costs.

PR #712, opened 2019-10-30
fix e2e test config in csproj

realwanpengli:


test/Microsoft.Azure.SignalR.E2ETests/Microsoft.Azure.SignalR.E2ETests.csproj
```diff
@@ -25,4 +25,9 @@
     <ProjectReference Include="..\..\src\Microsoft.Azure.SignalR.Management\Microsoft.Azure.SignalR.Management.csproj" />
     <ProjectReference Include="..\..\src\Microsoft.Azure.SignalR.Common\Microsoft.Azure.SignalR.Common.csproj" />
   </ItemGroup>
+
+  <ItemGroup Condition=" '$(AzureSignalRSDKE2ETest)' == 'true' ">
```
vicancy: What is this for? It already have `ProjectReference` for Management above

realwanpengli: Only for nightly e2e test. Do you remember we want to e2e test including restore packages from myget?

vicancy: Then remove the one from above for this condition?

PR #722, opened 2019-11-06
Rename methods

vicancy:
Add `Client` to all client related methods


terencefan:
I think this PR can be merged? Since my major PR has already been merged.

PR #723, opened 2019-11-06
Some refactor for `graceful shutdown` feature

terencefan:
Fix potential capacity problems, for `ASP.NET` and `ASP.NET Core` respectively.

1. Add `IClientConnectionLifetime` interface.
    - This interface can be used to let us know whether our client connections have already stopped.
    - Will be implemented In the future PR #689 , nothing in it now.
2. Use `ServiceConnectionContainerFactory` to create `MultiEndpointConnectionContainer`.

Here is a simple picture explained what I did in this PR.

- Blue: existing `class`
- Yellow: existing `interface`
- Green: new `class` or `interface`

image


src/Microsoft.Azure.SignalR.Common/Interfaces/IClientConnectionLifetime.cs
```diff
@@ -0,0 +1,8 @@
+﻿using System.Threading.Tasks;
+
+namespace Microsoft.Azure.SignalR
+{
+    interface IClientConnectionLifetime
```
vicancy: add `internal` explicitly. nit naming: `IClientConnectionLifetimeManager`?

Issue #535, opened 2019-05-17
AzureSignalR Proxy support for ServiceEndpoint connection


tebeco:
This issue has been created to address Hub connectivity throught a proxy,
I've been wondering about it before so i tried to provide a PR #534  

## Context :
* The goal of this pull request is to run an application hosting a `Hub` and `AddAzureSignalR` 
* As part of a company that still refuse to use transparent proxy we have to explicitly "point" at the proxy for outgoing request
* This is done under the hood by this Sdk when the application start.

## No goal
Other idea was suggested before this PR but are not really possible :
* Targeting `netfx`
  ==> NO GO "because it has magic code that interact with windows"
  ==> Require to setup the proxy in IE
  ==> Will not work if not on windows
* using `HTTP_PROXY`
  ==> Pretty bad idea when your application communicate mostly in the LAN 
  ==> Especially when the only place you need a `Proxy` is for the connection to `Azure Signalr`
* using `NO_PROXY`
  ==> would be too complexe to filter out internal network
* `netcoreapp3.0` 
  ==> `netcoreapp3.0` add `HTTP_SUPPORT` (see previous point)
  ==> We arleady have a running app on `netcoreapp2.2` waiting for `netcoreapp3.0` would require much more change only for this

## Target
The PR is targeting "master" because it "look like" the current version that is already published on nuget.org and is the one targeting `netstandard2.0` so we could use it while still targetting `netcoreapp2.x`

## A bit of details

If we take a look as the CSharp client for signalr, it seems that the class [`HttpConnectionOptions`](https://github.com/aspnet/AspNetCore/blob/master/src/SignalR/clients/csharp/Http.Connections.Client/src/HttpConnectionOptions.cs) contains code for :
* [`public IWebProxy Proxy { get; set; }`](https://github.com/aspnet/AspNetCore/blob/master/src/SignalR/clients/csharp/Http.Connections.Client/src/HttpConnectionOptions.cs#L119)
* [`public Func<HttpMessageHandler, HttpMessageHandler> HttpMessageHandlerFactory { get; set; }`](https://github.com/aspnet/AspNetCore/blob/master/src/SignalR/clients/csharp/Http.Connections.Client/src/HttpConnectionOptions.cs#L39)
[...]

PR #727, opened 2019-11-07
Add proxy support for the ServiceManager from ServiceManagerOptions

tebeco:
The idea of this PR is the same as the already Merged PR #534 

This issue has been created to address `ServiceManager` connectivity throught a proxy,
i tried to provide a PR #727 

### Investigation
The `ServiceManager` rely on the `ServiceEndpointProvider` for connectivity
the `ServiceEndpointProvider` already `ServiceOptions` which already has Proxy support (see #534 / #535)

### Proposed fix
The idea is just to add an `IWebProxy` property to the `ServiceManagerOptions` so that is can be forwarded to the `ServiceEndpointProvider` through the `ServiceOptions` :

```
            _endpointProvider = new ServiceEndpointProvider(_endpoint, Options.Create(new ServiceOptions
            {
                // ....
                Proxy = serviceManagerOptions.Proxy
            }));
```

### DIFF
--- a/src/Microsoft.Azure.SignalR.Management/ServiceManager.cs
+++ b/src/Microsoft.Azure.SignalR.Management/ServiceManager.cs
@@ -32,7 +32,8 @@ namespace Microsoft.Azure.SignalR.Management
             _endpoint = new ServiceEndpoint(_serviceManagerOptions.ConnectionString, EndpointType.Secondary);
             _endpointProvider = new ServiceEndpointProvider(_endpoint, Options.Create(new ServiceOptions
             {
-                ApplicationName = _serviceManagerOptions.ApplicationName
+                ApplicationName = _serviceManagerOptions.ApplicationName,
+                Proxy = serviceManagerOptions.Proxy
             }));
             _serverNameProvider = new DefaultServerNameProvider();
             _productInfo = productInfo;
--- a/src/Microsoft.Azure.SignalR.Management/ServiceManagerOptions.cs
+++ b/src/Microsoft.Azure.SignalR.Management/ServiceManagerOptions.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Net;
 using Microsoft.Azure.SignalR;
 
 namespace Microsoft.Azure.SignalR.Management
@@ -30,6 +31,11 @@ namespace Microsoft.Azure.SignalR.Management
         /// Gets or sets the total number of connections from SDK to Azure SignalR Service. Default value is 1.
         /// </summary>
         public int ConnectionCount { get; set; } = 1;
+
+        /// <summary>
+        /// Gets or sets the proxy used when ServiceManager will attempt to connect to Azure SignalR Service.
+        /// </summary>
+        public IWebProxy Proxy { get; set; }
 
         internal void ValidateOptions()
         {
--- a/src/Microsoft.Azure.SignalR/ServiceOptions.cs
+++ b/src/Microsoft.Azure.SignalR/ServiceOptions.cs
@@ -65,7 +65,7 @@ namespace Microsoft.Azure.SignalR
         internal TimeSpan ServerShutdownTimeout { get; set; } = TimeSpan.FromSeconds(Constants.DefaultShutdownTimeoutInSeconds);
 
         /// <summary>
-        /// Gets or sets the proxy used when ServiceEndpoint will attempt to connect to Azure SignalR.
+        /// Gets or sets the proxy used when ServiceEndpoint will attempt to connect to Azure SignalR Service.
         /// </summary>
         public IWebProxy Proxy { get; set; }
     }


realwanpengli:
Could you add some tests for it?

tebeco:
On my way
Just to be sure how do you want the test to be designed ?
Checking that
* the IOptions forward the value to the `ServiceManager`
* to the underlying `ServicepointProvider` ?
* lots of test are already KO because of connection string I believe test on the repo are like end to end and I need to create an ASR ? (Did I miss some docs about the setup to be able to run the test ?)
* can you point me to a similar existing test because if I have to pull out an E2E lookalike I would love to use the same pattern as the existing one

tebeco:
ok found it here :
https://github.com/Azure/azure-signalr/blob/dev/test/appsettings.Test.json#L2

Is it normal that the doc for the current branch is explaining about how to build using `dotnet/arcade` ?
(Probably way i skipped the build doc)

If i understand if correctly, arcade is going to land in this repo but not yet ?

tebeco:
@realwanpengli 
I've try to add something like this :
```csharp
   public class ServiceManagerFacts
    {
        private static readonly IWebProxy[] _proxies = new IWebProxy[] { null };

        public static IEnumerable<object[]> TestGenerateProxyData => from appName in _appNames
                                                                     from proxy in _proxies
                                                                     select new object[] { appName, proxy };

    }

        [Theory]
        [MemberData(nameof(TestGenerateProxyData))]
        internal async Task CreateServiceManagerProxyTest(string appName, IWebProxy proxy)
        {
            var serviceManager = new ServiceManager(new ServiceManagerOptions
            {
                ApplicationName = appName,
                Proxy = proxy
            }, null);


            Assert.Equal(serviceManager.??????????????)
        }
```

but here is the API surface of the `ServiceManager` for the test point of view :
image

As you can see there there is no way to reach either the `ServiceManagerOptions` or the `ServiceEndpointProvider` that are declared like this :

```csharp
    {
        private readonly ServiceManagerOptions _serviceManagerOptions;
        private readonly ServiceEndpointProvider _endpointProvider;
        private readonly ServiceEndpoint _endpoint;
...
    }
```

I used to test this kind of scenario by resolving the existing `IOptions<>` from the `IServiceProvider`.  
It seems that this lib does not rely on the same paradigm, so i have no `IOptions<>` to resolves  
The imediate effect is that the code in this repo does not rely on `DependencyInjection` so it's very very very hard for me to understand how should i do unit test these class.  

As of today, to test proxy settings, it feels like you would have to create your own Infra with a proxy and spin it.  

Do you know if there is an planned refacto so that this lib relies on DependencyInjection in order to be able to have unit testing made easier ?

For example i attempted to work on the `Transient` part but i found this :  
https://github.com/Azure/azure-signalr/blob/dev/src/Microsoft.Azure.SignalR.Management/HttpClientFactory.cs

realwanpengli:
> On my way
> Just to be sure how do you want the test to be designed ?
> Checking that
> 
> * the IOptions forward the value to the `ServiceManager`
> * to the underlying `ServicepointProvider` ?
> * can you point me to a similar existing test because if I have to pull out an E2E lookalike I would love to use the same pattern as the existing one

I think you can make a E2E lookalike test. The existing one which is similar is [StopServiceHubContextTest](https://github.com/Azure/azure-signalr/blob/dev/test/Microsoft.Azure.SignalR.E2ETests/Management/ServiceHubContextE2EFacts.cs#L222)



> Should this be moved to the Docs regarding the `Build this repo` ?
> https://github.com/Azure/azure-signalr/blob/dev/test/appsettings.Test.json#L2
> 
> Is it normal that the doc for the current branch is explaining about how to build using `dotnet/arcade` ?
> (Probably way i skipped the build doc)
> 
> If i understand if correctly, arcade is going to land in this repo but not yet ?

[appsettings.Test.json](https://github.com/Azure/azure-signalr/blob/dev/test/appsettings.Test.json) is only for E2E test. So it isn't in the doc.
We are still using Korebuild.

> Do you know if there is an planned refacto so that this lib relies on DependencyInjection in order to be able to have unit testing made easier ?
The src repo can be mark as `InternalsVisibleTo` to test repo, so anything that is internal can be used by test repo. [sample](https://github.com/Azure/azure-signalr/blob/dev/test/Microsoft.Azure.SignalR.E2ETests/Management/ServiceHubContextE2EFacts.cs#L235)

I make a PR #731 to make test easier.

tebeco:
Thx a lot for all that info and the change.
Just to be sure this will be tested through an E2E ? No unit test for this one.
I began writing something like :
```
public class MockWebProxy : IWebProxy
{
  public bool WasCalled {get;set;} = false;

  public Uri GetProxy(Uri uri)
  {
    WasCalled = true;
    return uri;
  }
  ...
}
```

Then asserting that `WasCalled` was properly set to true but I’m not sure of the idea
(As you suggested I kinda had the same idea by re-using `TestConfiguration.Instance...)

I’ll take a look at the test you pointed

Is there any thing else you have in mind ?

tebeco:
@realwanpengli 
Thx, just saw the commit en `dev` branch, will take a look at it again.

Had some trouble while trying `E2E` as `IWebProxy` only allow `http` scheme => no `https` or `wss`
I may have done something wrong

(rebasing my PR)

realwanpengli:
@tebeco Sorry for the late response.
The schema is hard code [here](https://github.com/tebeco/azure-signalr/blob/service-manager-proxy-support/test/Microsoft.Azure.SignalR.Tests.Common/E2ETest/TestServerBase.cs#L52). You can change the schema here.

tebeco:
@realwanpengli 
not sure i get your point, the errors is that, if i understand correctly, `WebProxy` won't allow `wss` scheme


Speaking about that ... totally unrelated
```
        private static string GetRandomPortUrl()
        {
            return $"http://localhost:{StaticRandom.Next(49152, 65535)}";
        }
```
why not Bind the port on port `0` ?

this maybe bad, but it gives a random free port :

```
var host = WebHost.CreateDefaultBuilder()
                        .UseUrls("http://*:0")
                        .ConfigureServices(services => { })
                        .Configure(app => { })
                        .Build();

await host.StartAsync().ConfigureAwait(false);

var serverFeature = host.ServerFeatures.Select(kvp => kvp.Value).OfType<IServerAddressesFeature>().Single();
var address = serverFeature.Addresses.Select(address => new Uri(address)).Single(); //Can use sonething else that `Single` if `UseUrls` ...

var baseUri = new Uri($"http://localhost:{address.Port}");
```

tebeco:
@vicancy @realwanpengli 

Is there any test on the repo that did tried to test the `Proxy` or the connectivity with a proxy ?
I'm really willing to make these tests works but i'm not sure :
* with current design, the `Value` is only set during connection process, so i can't just run `IsEquals()` unless the connection is attempted
* what to test ?
  * Connection failure with wrong proxy ? looks possible
  * Connection Success with `null` => no proxy (should be te easiest test i guess)
  * Connection success with a proxy ? => this does really need an actually running proxy
    * We use `Squid` at work, running in `AKS` to proxify the access from the `LAN` to either
      * `Hub` (previous PR)
      * `ASR` with `ServiceManagment`, this PR

I'm in a dead end here, going in holiday in few day and will be back at the end of January

I attempted multiple times to trick `.Net` `Signalr` and `WebProxy` around. I deleted few test since the rebase with the PR @realwanpengli made to access internal field but it leaves with other "End To End" issue

this is enabled :
image
so if you think you found to way to test it feel free to either merge the PR as/is or update it and merge it.

As this repo is on your side, you might have more visibility to decide how you want an E2E test to run :
* Real e2e (you need to create a real proxy somehow, or add a `docker run` with a proxy in the CI)
* `IWebProxy` (could not make it work but it's probably me here TBH)
* ...

tebeco:
(i'm rebasing now the branch on Dev branch so that less quirk on merge)

vicancy:
How about merging this PR for the upcoming release? I think we can help to improve the e2e test from our side with more infra-layer enhancement along. @realwanpengli Any concerns?

realwanpengli:
No, how about merge it first, then we add the tests later

tebeco:
Well, you’re the bosses here
I don’t mind one choice or another ;)

But to be honest I’m probably going to need some help if the test does need an actual connection to work to properly work

If it needs :
* one Host with the sdk 
* another host as an aspnetcore app acting as a proxy for http(s) and ws(s)
Why not, I’m just not sure what to use
Maybe something like https://github.com/damianh/ProxyKit
 that would makes a proper E2E => no infra related and all the code is here

tebeco:
Thx a lot @wanlwanl and @vicancy 
Is there anyway I could help for the test ?
(I’ll be back at work mid January if you want me to add test over it)

I’ll just need more direction or choice on your side since test over a would need connectivity :
* another webapi/kestrel running something like proxyKit
* spinning something like `squid` as a container
* should it be isolated to be sure the proxy does the dns resolution ?
* only testing failing ? And null proxy ?

Also : 
How to make the test works for contributor that have no proxy locally ?
It should probably be skipped based on a detection or a setting ?

wanlwanl:
@tebeco Thanks for offering help! I'm going on vacation and will be back in early January. I found a [blog](https://marcin-chwedczuk.github.io/dotnet-core-http-client-proxy-and-localhost) about proxy, it might be helpful. Looks like we can do a basic test locally first in this way. As for other discussions, how about we discuss them when you come back. Have a nice vacation!

PR #730, opened 2019-11-11
Add Interfaces for `Graceful shutdown`

terencefan:
Related to #689 
1. Add `ConnectionOfflineTask` for `IServiceConnection`
    - Use for waiting for a server connection to be offlined.
2. Add `AllConnectionsCompleted` for `IClientConnectionLifetimeManager`
    - Use for waiting for existing client connections to be stopped.

### DIFF
--- a/src/Microsoft.Azure.SignalR.AspNet/ClientConnections/ClientConnectionManager.cs
+++ b/src/Microsoft.Azure.SignalR.AspNet/ClientConnections/ClientConnectionManager.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.AspNet.SignalR;
 using Microsoft.AspNet.SignalR.Hosting;
@@ -120,6 +121,8 @@ namespace Microsoft.Azure.SignalR.AspNet
                 return reader.ReadToEnd();
             }
         }
+
+        public Task WhenAllCompleted() => Task.CompletedTask;
 
         private sealed class ClientConnectionHubDispatcher : HubDispatcher
         {
--- a/src/Microsoft.Azure.SignalR.Common/Interfaces/IClientConnectionLifetimeManager.cs
+++ b/src/Microsoft.Azure.SignalR.Common/Interfaces/IClientConnectionLifetimeManager.cs
@@ -1,6 +1,9 @@
-﻿namespace Microsoft.Azure.SignalR
+﻿using System.Threading.Tasks;
+
+namespace Microsoft.Azure.SignalR
 {
     interface IClientConnectionLifetimeManager
     {
+        Task WhenAllCompleted();
     }
 }
--- a/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
+++ b/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
@@ -32,6 +32,7 @@ namespace Microsoft.Azure.SignalR
         private readonly SemaphoreSlim _writeLock = new SemaphoreSlim(1, 1);
 
         private readonly TaskCompletionSource<bool> _serviceConnectionStartTcs = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
+        private readonly TaskCompletionSource<object> _serviceConnectionOfflineTcs = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
 
         private readonly ServerConnectionType _connectionType;
 
@@ -83,6 +84,8 @@ namespace Microsoft.Azure.SignalR
         }
 
         public Task ConnectionInitializedTask => _serviceConnectionStartTcs.Task;
+
+        public Task ConnectionOfflineTask => _serviceConnectionOfflineTcs.Task;
 
         protected ServiceConnectionBase(IServiceProtocol serviceProtocol, string connectionId,
             HubServiceEndpoint endpoint, IServiceMessageHandler serviceMessageHandler, ServerConnectionType connectionType, ILogger logger)
@@ -259,6 +262,11 @@ namespace Microsoft.Azure.SignalR
             {
                 Log.ReceivedInstanceOfflinePing(Logger, instanceId);
                 return CleanupConnections(instanceId);
+            } 
+            if (pingMessage.TryGetValue(Constants.ServicePingMessageKey.ShutdownKey, out string val) && val == Constants.ServicePingMessageValue.ShutdownFinAck)
+            {
+                _serviceConnectionOfflineTcs.TrySetResult(null);
+                return Task.CompletedTask;
             }
             return _serviceMessageHandler.HandlePingAsync(pingMessage);
         }
--- a/src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnectionContext.cs
+++ b/src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnectionContext.cs
@@ -35,6 +35,10 @@ namespace Microsoft.Azure.SignalR
             readerScheduler: PipeScheduler.ThreadPool,
             useSynchronizationContext: false);
 
+        private readonly TaskCompletionSource<object> _connectionEndTcs = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
+
+        public Task CompleteTask => _connectionEndTcs.Task;
+
         private readonly object _heartbeatLock = new object();
         private List<(Action<object> handler, object state)> _heartbeatHandlers;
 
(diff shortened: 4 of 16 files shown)


src/Microsoft.Azure.SignalR.AspNet/ClientConnections/ClientConnectionManager.cs
```diff
@@ -60,6 +62,8 @@ public ClientConnectionManager(HubConfiguration configuration, ILoggerFactory lo
 
         public bool TryAdd(string connectionId, IServiceConnection serviceConnection)
         {
+            TaskCompletionSource<bool> tcs = new TaskCompletionSource<bool>(TaskContinuationOptions.RunContinuationsAsynchronously);
```
vicancy: Should be `TaskCreationOptions`

vicancy: Why not add the tcs as a property of the Client Connection?

src/Microsoft.Azure.SignalR.AspNet/ClientConnections/ClientConnectionManager.cs
```diff
@@ -119,8 +127,18 @@ internal static string GetContentAndDispose(MemoryStream stream)
...
+        }
+
+        public async Task AllConnectionsCompleted()
+        {
+            List<Task> tasks = new List<Task>();
+            foreach (var tcs in _clientCompletedTcs.Values)
+            {
+                tasks.Add(tcs.Task);
+            }
+            await Task.WhenAll(tasks);
```
vicancy: `return Task.WhenAll(_clientCompletedTcs.Select(s=>s.Value.Task))`

src/Microsoft.Azure.SignalR.Common/Interfaces/IClientConnectionLifetimeManager.cs
```diff
@@ -1,6 +1,9 @@
-﻿namespace Microsoft.Azure.SignalR
+﻿using System.Threading.Tasks;
+
+namespace Microsoft.Azure.SignalR
 {
     interface IClientConnectionLifetimeManager
     {
+        Task AllConnectionsCompleted();
```
vicancy: naming `WaitForCompletion`?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -32,6 +32,7 @@ internal abstract class ServiceConnectionBase : IServiceConnection
         private readonly SemaphoreSlim _writeLock = new SemaphoreSlim(1, 1);
 
         private readonly TaskCompletionSource<bool> _serviceConnectionStartTcs = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
+        private readonly TaskCompletionSource<bool> _serviceConnectionOfflineTcs = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
```
vicancy: `TaskCompletionSource<object>` is enough

src/Microsoft.Azure.SignalR.AspNet/ClientConnections/ClientConnectionManager.cs
terencefan: Because we don't need to, and the param here is an interface so I don't want to make things more complex...

terencefan: Actually `TaskContinuationOptions.RunContinuationsAsynchronously` and `TaskCreationOptions.RunContinuationsAsynchronously` share the same value.

But I have fixed it.

src/Microsoft.Azure.SignalR.AspNet/ClientConnections/ClientConnectionManager.cs
terencefan: I don't know why I can't do this in my `VS`?

I've been told that no definition for `Select`, for both `ConcurrentDictionary` and `ICollections` (value)

terencefan: Problem solved by adding "using System.Linq"

src/Microsoft.Azure.SignalR.AspNet/ClientConnections/ClientConnectionManager.cs
BrennanConroy: > Actually `TaskContinuationOptions.RunContinuationsAsynchronously` and `TaskCreationOptions.RunContinuationsAsynchronously` share the same value.


Just for future knowledge, just because they share the same value does not mean they are treated that same when creating a `TaskCompletionSource`. `TaskCompletionSource` has a constructor that takes an `object` and if you pass `TaskCompletionSource` in then it will use the `object` overload and end up being synchronous instead of the expected asynchronous behavior.

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnectionContext.cs
```diff
@@ -35,6 +35,10 @@ internal class ServiceConnectionContext : ConnectionContext,
             readerScheduler: PipeScheduler.ThreadPool,
             useSynchronizationContext: false);
 
+        private readonly TaskCompletionSource<bool> _connectionEndTcs = new TaskCompletionSource<bool>(TaskContinuationOptions.RunContinuationsAsynchronously);
```
vicancy: >TaskContinuationOptions [](start = 103, length = 23)

`TaskCreationOptions`, as pointed out by Brennan, they are different although the value is the same

test/Microsoft.Azure.SignalR.AspNet.Tests/ClientConnectionManagerTests.cs
```diff
@@ -93,6 +98,32 @@ public void TestGetUserPrincipal(string expectedAuthenticationType, string expec
...
         }
 
+        private async void RemoveConnection(IClientConnectionManager manager, string connectionId)
+        {
+            await Task.Delay(100);
+            manager.TryRemoveServiceConnection(connectionId, out _);
+        }
+
+        [Fact]
+        public async void TestAllClientConnectionsCompleted()
```
vicancy: >TestAllClientConnectionsCompleted [](start = 26, length = 33)

remove this test?

test/Microsoft.Azure.SignalR.AspNet.Tests/ClientConnectionManagerTests.cs
```diff
@@ -93,6 +98,32 @@ public void TestGetUserPrincipal(string expectedAuthenticationType, string expec
...
+            var token = "?connectionToken=invalidOne";
+
+            await manager.CreateConnection(new OpenConnectionMessage("foo", new Claim[0], null, token), null);
+            await manager.CreateConnection(new OpenConnectionMessage("bar", new Claim[0], null, token), null);
+
+            RemoveConnection(manager, "foo");
+            RemoveConnection(manager, "bar");
+
+            var expected = manager.WaitForCompletion();
+            var index = Task.WaitAny(
```
vwxyzh: WhenAny

src/Microsoft.Azure.SignalR.Common/Interfaces/IClientConnectionLifetimeManager.cs
```diff
@@ -1,6 +1,9 @@
-﻿namespace Microsoft.Azure.SignalR
+﻿using System.Threading.Tasks;
+
+namespace Microsoft.Azure.SignalR
 {
     interface IClientConnectionLifetimeManager
     {
+        Task WaitForCompletion();
```
vwxyzh: naming: When...

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -405,6 +408,18 @@ private async Task<bool> ReceiveHandshakeResponseAsync(PipeReader input, Cancell
             }
         }
 
+        private void ProcessMessage(ServiceMessage message)
+        {
+            if (message is PingMessage ping && ping.TryGetValue(Constants.ServicePingMessageKey.ShutdownKey, out string val) && val == Constants.ServicePingMessageValue.ShutdownFinAck)
```
vwxyzh: move to `OnPingMessageAsync`

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnectionContext.cs
```diff
@@ -35,6 +35,10 @@ internal class ServiceConnectionContext : ConnectionContext,
             readerScheduler: PipeScheduler.ThreadPool,
             useSynchronizationContext: false);
 
+        private readonly TaskCompletionSource<object> _connectionEndTcs = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
+
+        public Task CompleteTask { get => _connectionEndTcs.Task; }
```
vwxyzh: `public Task CompleteTask => _connectionEndTcs.Task;`

terencefan:
All fixed @vwxyzh @vicancy

PR #739, opened 2019-11-18
Server connection migration implementation

terencefan:
Related to #689 

The migration will start on the `Runtime` side after the graceful shutdown process begins.

And there is nothing to do with **the old server**. 
> we simply close each connection after receiving `CloseConnectionMessage`.

But we have to prevent `HandshakeResponse` from reaching our `Runtime` after the client connection has been migrated to **the new server** since the client connection has not been really dropped by our `Runtime`. It would be weird if we sent `HandshakeReponse` to our client again.

### DIFF
--- a/src/Microsoft.Azure.SignalR/ServerConnections/ClientConnectionContext.cs
+++ b/src/Microsoft.Azure.SignalR/ServerConnections/ClientConnectionContext.cs
@@ -15,7 +15,6 @@ using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Connections.Features;
 using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Http.Features.Authentication;
-using Microsoft.AspNetCore.SignalR;
 using Microsoft.AspNetCore.WebUtilities;
 using Microsoft.Azure.SignalR.Protocol;
 using Microsoft.Extensions.Primitives;
@@ -39,6 +38,8 @@ namespace Microsoft.Azure.SignalR
 
         public Task CompleteTask => _connectionEndTcs.Task;
 
+        public bool IsMigrated { get; }
+
         private readonly object _heartbeatLock = new object();
         private List<(Action<object> handler, object state)> _heartbeatHandlers;
 
@@ -46,6 +47,11 @@ namespace Microsoft.Azure.SignalR
         {
             ConnectionId = serviceMessage.ConnectionId;
             User = serviceMessage.GetUserPrincipal();
+
+            if (serviceMessage.Headers.TryGetValue(Constants.AsrsMigrateIn, out _))
+            {
+                IsMigrated = true;
+            }
 
             // Create the Duplix Pipeline for the virtual connection
             transportPipeOptions = transportPipeOptions ?? DefaultPipeOptions;
--- a/src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.Log.cs
+++ b/src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.Log.cs
@@ -47,6 +47,12 @@ namespace Microsoft.Azure.SignalR
             private static readonly Action<ILogger, Exception> _applicationTaskTimedOut =
                 LoggerMessage.Define(LogLevel.Error, new EventId(21, "ApplicationTaskTimedOut"), "Timed out waiting for the application task to complete.");
 
+            private static readonly Action<ILogger, string, Exception> _migrationStarting =
+                LoggerMessage.Define<string>(LogLevel.Debug, new EventId(22, "MigrationStarting"), "Connection {TransportConnectionId} migrated from another server.");
+
+            private static readonly Action<ILogger, string, Exception> _errorSkippingHandshakeResponse =
+                LoggerMessage.Define<string>(LogLevel.Error, new EventId(23, "ErrorSkippingHandshakeResponse"), "Error while skipping handshake response during migration, the connection will be dropped on the client-side. Error detail: {message}");
+
             public static void FailedToCleanupConnections(ILogger logger, Exception exception)
             {
                 _failedToCleanupConnections(logger, exception);
@@ -82,6 +88,11 @@ namespace Microsoft.Azure.SignalR
                 _connectedStarting(logger, connectionId, null);
             }
 
+            public static void MigrationStarting(ILogger logger, string connectionId)
+            {
+                _migrationStarting(logger, connectionId, null);
+            }
+
             public static void ConnectedEnding(ILogger logger, string connectionId)
             {
                 _connectedEnding(logger, connectionId, null);
@@ -100,6 +111,11 @@ namespace Microsoft.Azure.SignalR
             public static void ApplicationTaskTimedOut(ILogger logger)
             {
                 _applicationTaskTimedOut(logger, null);
+            }
+
+            public static void ErrorSkippingHandshakeResponse(ILogger logger, Exception ex)
+            {
+                _errorSkippingHandshakeResponse(logger, ex.Message, ex);
             }
         }
     }
--- a/src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
+++ b/src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
@@ -12,6 +12,7 @@ using Microsoft.AspNetCore.Http;
 using Microsoft.Azure.SignalR.Protocol;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Primitives;
+using SignalRProtocol = Microsoft.AspNetCore.SignalR.Protocol;
 
 namespace Microsoft.Azure.SignalR
 {
@@ -21,6 +22,8 @@ namespace Microsoft.Azure.SignalR
         // .NET Framework has restriction about reserved string as the header name like "User-Agent"
         private static readonly Dictionary<string, string> CustomHeader = new Dictionary<string, string> { { Constants.AsrsUserAgent, ProductInfo.GetProductInfo() } };
         private static readonly TimeSpan CloseTimeout = TimeSpan.FromSeconds(5);
+
+        private readonly bool _enableConnectionMigration;
 
         private const string ClientConnectionCountInHub = "#clientInHub";
         private const string ClientConnectionCountInServiceConnection = "#client";
@@ -53,6 +56,8 @@ namespace Microsoft.Azure.SignalR
             _connectionFactory = connectionFactory;
             _connectionDelegate = connectionDelegate;
             _clientConnectionFactory = clientConnectionFactory;
+
+            _enableConnectionMigration = false;
         }
 
         protected override Task<ConnectionContext> CreateConnection(string target = null)
@@ -102,8 +107,16 @@ namespace Microsoft.Azure.SignalR
         {
             var connection = _clientConnectionFactory.CreateConnection(message, ConfigureContext);
             connection.ServiceConnection = this;
-            AddClientConnection(connection, GetInstanceId(message.Headers));
-            Log.ConnectedStarting(Logger, connection.ConnectionId);
+            AddClientConnection(connection, message);
+
+            if (connection.IsMigrated)
+            {
+                Log.MigrationStarting(Logger, connection.ConnectionId);
+            }
+            else
+            {
+                Log.ConnectedStarting(Logger, connection.ConnectionId);
+            }
 
             // Execute the application code
             connection.ApplicationTask = _connectionDelegate(connection);
@@ -120,6 +133,17 @@ namespace Microsoft.Azure.SignalR
         protected override Task OnClientDisconnectedAsync(CloseConnectionMessage closeConnectionMessage)
         {
             var connectionId = closeConnectionMessage.ConnectionId;
+            if (_enableConnectionMigration && _clientConnectionManager.ClientConnections.TryGetValue(connectionId, out var context))
+            {
+                if (!context.HttpContext.Request.Headers.ContainsKey(Constants.AsrsMigrateOut))
+                {
+                    context.HttpContext.Request.Headers.Add(Constants.AsrsMigrateOut, "");
+                }
+                // We have to prevent SignalR `{type: 7}` (close message) from reaching our client while doing migration.
+                // Since all user-created messages will be sent to `ServiceConnection` directly.
+                // We can simply ignore all messages came from the application pipe.
+                context.Application.Input.CancelPendingRead();
+            }
             return PerformDisconnectAsyncCore(connectionId, false);
         }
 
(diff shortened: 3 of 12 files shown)


samples/ChatSample/ChatSample/Startup.cs
```diff
@@ -23,7 +23,7 @@ public void ConfigureServices(IServiceCollection services)
         }
 
         public void Configure(IApplicationBuilder app)
-        {
```
vicancy: please prevent such empty changes

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
```diff
@@ -113,6 +114,17 @@ private async Task ProcessOutgoingMessagesAsync(ServiceConnectionContext connect
                     var buffer = result.Buffer;
                     if (!buffer.IsEmpty)
                     {
+                        // We assume the first response message would be a HandshakeResponse.
+                        if (connection.IsMigrated)
+                        {
+                            if (SignalRProtocol.HandshakeProtocol.TryParseResponseMessage(ref buffer, out var message))
```
vwxyzh:
```cs
var handshakeBuffer = buffer;
if (SignalRProtocol.HandshakeProtocol.TryParseResponseMessage(ref handshakeBuffer, out var message))
{
  connection.IsMigrated = false;
  buffer = buffer.Slice(handshakeBuffer.Length);
  if (buffer.IsEmpty)
  {
    connection.Application.Input.AdvanceTo(buffer.End);
    continue;
  }
}
```

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.Log.cs
```diff
@@ -32,6 +32,9 @@ private static class Log
             private static readonly Action<ILogger, string, Exception> _connectedStarting =
                 LoggerMessage.Define<string>(LogLevel.Debug, new EventId(11, "ConnectedStarting"), "Connection {TransportConnectionId} started.");
 
+            private static readonly Action<ILogger, string, Exception> _migrationStarting =
+                LoggerMessage.Define<string>(LogLevel.Debug, new EventId(22, "MigrationStarting"), "Connection {TransportConnectionId} migrated from another server.");
```
vicancy: why event id is `22`?

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
```diff
@@ -7,6 +7,7 @@
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using SignalRProtocol = Microsoft.AspNetCore.SignalR.Protocol;
```
vicancy: move this using below others

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
```diff
@@ -113,6 +114,17 @@ private async Task ProcessOutgoingMessagesAsync(ServiceConnectionContext connect
                     var buffer = result.Buffer;
                     if (!buffer.IsEmpty)
                     {
+                        // We assume the first response message would be a HandshakeResponse.
+                        if (connection.IsMigrated)
+                        {
+                            if (SignalRProtocol.HandshakeProtocol.TryParseResponseMessage(ref buffer, out var message))
+                            {
+                                connection.IsMigrated = false;
+                                connection.Application.Input.AdvanceTo(buffer.End);
```
vicancy: should be `buffer.Start`

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
```diff
@@ -149,17 +161,27 @@ private async Task ProcessOutgoingMessagesAsync(ServiceConnectionContext connect
...
             _clientConnectionManager.AddClientConnection(connection);
             _connectionIds.TryAdd(connection.ConnectionId, instanceId);
         }
 
         protected override Task OnConnectedAsync(OpenConnectionMessage message)
         {
             var connection = _clientConnectionFactory.CreateConnection(message, ConfigureContext);
-            AddClientConnection(connection, GetInstanceId(message.Headers));
-            Log.ConnectedStarting(Logger, connection.ConnectionId);
+            AddClientConnection(connection, message);
```
vicancy: why change to pass in `message`?

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
vicancy: If this is expected to be the first message for Migrated connection, move this parse logic to the top of the function outside of the while loop? Then there is no need to reset the `IsMigrated` property and this property can be readonly

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
terencefan: Because we have to read `message.Headers` while creating the connection.
Our migration flag defined inside of it.

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
terencefan: Why? We just want to skip this message.

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.Log.cs
terencefan: 21 has already been occupied.
And since `connectedStarting` and `migrationStarting` are called in a similar scenario I think they should be put together?

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
terencefan: No, not all `HandshakeResponse` should be intercepted.

We only care about the connection that has been migrated from another server.

terencefan: Sounds great.
I'll try implementing this and see what will happen.

terencefan: But make this kind of change could have unprecedented impacts on our existing logic so maybe we have to reconsider it carefully.

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.Log.cs
vicancy: interesting, the previous pattern is order by event id though

---
In reply to: [347767989](https://github.com/Azure/azure-signalr/pull/739#discussion_r347767989) [](ancestors = 347767989)

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
vicancy: `buffer.End` skips all the content inside the pipe, for example "{handshake} {real message}"...

---
In reply to: [347766993](https://github.com/Azure/azure-signalr/pull/739#discussion_r347766993) [](ancestors = 347766993)

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
vicancy: I don't see related changes?

---
In reply to: [347766851](https://github.com/Azure/azure-signalr/pull/739#discussion_r347766851) [](ancestors = 347766851)

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.Log.cs
BrennanConroy: Would definitely recommend keeping the IDs in order, otherwise every time you add a new ID you need to check each log to make sure the ID isn't in use.

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
terencefan: I'm afraid our existing logic will face the same problem.
Maybe we have to fix them all if this is very important.
( since only ping and close message will in this pipe I don't think it is necessary

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.Log.cs
terencefan: Ok, I'll rearrange it ~

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
```diff
@@ -170,7 +219,13 @@ private async Task ProcessOutgoingMessagesAsync(ClientConnectionContext connecti
...
 
                     var buffer = result.Buffer;
-                    if (!buffer.IsEmpty)
+
+                    // Messages will be simply dropped after the client connection has been migrated.
+                    // To avoid PingMessage & CloseMessage reaching the client.
+                    // There is no need to worry about the user-created messages because they have been redirected to another target
+                    // in `ServiceLifetimeManager`.
+                    // We simply replaced the `HubLifetimeManager` injected by SignalR to ours to achieve this feature.
+                    if (!buffer.IsEmpty && !connection.HasBeenMigrated)
```
vicancy: >HasBeenMigrated [](start = 55, length = 15)

How about cancel connection.Application's read directly?

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
```diff
@@ -157,8 +174,40 @@ protected override async Task OnClientMessageAsync(ConnectionDataMessage connect
             }
         }
 
+        private async Task SkipHandshakeResponse(ClientConnectionContext connection)
+        {
+            while (true)
+            {
+                var result = await connection.Application.Input.ReadAsync();
+                if (result.IsCanceled || result.IsCompleted)
```
vicancy: >IsCompleted [](start = 48, length = 11)

why break when `result.IsCompleted`? Still need to process the buffer when it  is completed?

if `result.IsCanceled` is there any need to continue the process?

also adding `try...catch...` wrap

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
terencefan: Good idea

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
```diff
@@ -120,6 +133,10 @@ protected override Task OnClientConnectedAsync(OpenConnectionMessage message)
         protected override Task OnClientDisconnectedAsync(CloseConnectionMessage closeConnectionMessage)
         {
             var connectionId = closeConnectionMessage.ConnectionId;
+            if (_enableConnectionMigration && _clientConnectionManager.ClientConnections.TryGetValue(connectionId, out var context))
+            {
+                context.HasBeenMigrated = true;
```
vicancy: `CancelPendingRead` here?

src/Microsoft.Azure.SignalR/ServerConnections/ClientConnectionContext.cs
```diff
@@ -37,7 +37,9 @@ internal class ClientConnectionContext : ConnectionContext,
 
         private readonly TaskCompletionSource<object> _connectionEndTcs = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
 
-        public Task CompleteTask => _connectionEndTcs.Task;
+        public Task CompleteTask => _connectionEndTcs.Task;
+
+        public readonly bool IsMigrated;
```
vicancy: >IsMigrated [](start = 29, length = 10)

as a convention, use property instead of field for public ones 
`public bool IsMigrated {get;}`

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
```diff
@@ -120,6 +133,10 @@ protected override Task OnClientConnectedAsync(OpenConnectionMessage message)
         protected override Task OnClientDisconnectedAsync(CloseConnectionMessage closeConnectionMessage)
         {
             var connectionId = closeConnectionMessage.ConnectionId;
+            if (_enableConnectionMigration && _clientConnectionManager.ClientConnections.TryGetValue(connectionId, out var context))
+            {
+                context.Application.Input.CancelPendingRead();
```
vicancy: >CancelPendingRead [](start = 42, length = 17)

Add some explanation as comments?

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
```diff
@@ -158,20 +175,66 @@ protected override async Task OnClientMessageAsync(ConnectionDataMessage connect
...
+                        return;
+                    }
+
+                    var buffer = result.Buffer;
+                    if (buffer.IsEmpty)
+                    {
+                        continue;
+                    }
+
+                    if (SignalRProtocol.HandshakeProtocol.TryParseResponseMessage(ref buffer, out var message))
```
vicancy: >TryParseResponseMessage [](start = 58, length = 23)

shall we log something when the first response is not handshake and just stop the connection?

vicancy: So the `OnConnected` hub event will be triggered on this "migrated to" app server, right? will `OnDisconnected` hub event on the "migrated from" app server?

---
In reply to: [351961944](https://github.com/Azure/azure-signalr/pull/739#discussion_r351961944) [](ancestors = 351961944)

test/Microsoft.Azure.SignalR.Tests/Infrastructure/TestConnectionFactory.cs
```diff
@@ -31,13 +37,18 @@ public TestConnectionFactory(Func<TestConnection, Task> connectCallback)
...
+            Connections.Add(connection);
+
             // Start a task to process handshake request from the newly-created server connection.
             _ = HandshakeAsync(connection);
 
             // Do something for test purpose
             await AfterConnectedAsync(connection);
 
-            await _connectCallback(connection);
+            if (null != _connectCallback)
```
vicancy: value move to right

test/Microsoft.Azure.SignalR.Tests/ServiceConnectionTests.cs
```diff
@@ -196,6 +205,85 @@ public async Task TestServiceConnectionWithEndlessApplicationTask()
             }
         }
 
+        private class TestServiceConnection : ServiceConnection
```
vicancy: private class below

src/Microsoft.Azure.SignalR/ServerConnections/ClientConnectionContext.cs
```diff
@@ -37,7 +36,9 @@ internal class ClientConnectionContext : ConnectionContext,
 
         private readonly TaskCompletionSource<object> _connectionEndTcs = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
 
-        public Task CompleteTask => _connectionEndTcs.Task;
+        public Task CompleteTask => _connectionEndTcs.Task;
+
+        public bool IsMigrated { get; private set; }
```
vicancy: remove `private set` when it is only set in ctor

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
```diff
@@ -158,20 +182,72 @@ protected override async Task OnClientMessageAsync(ConnectionDataMessage connect
...
+                if (connection.IsMigrated)
+                {
+                    using var source = new CancellationTokenSource(DefaultHandshakeTimeout);
+
+                    // A handshake response is not expected to be given
+                    // if the connection was migrated from another server, 
+                    // since the connection hasn't been `dropped` from the client point of view.
+                    if (!await SkipHandshakeResponse(connection, source.Token))
+                    {
+                        return;
```
vicancy: >return [](start = 24, length = 6)

log handshake failed here?

terencefan: yep

PR #744, opened 2019-11-25
Redirect send2conn from sending to Hub to SvcConn

terencefan:
We almost have all kinds of messages sending to our `ServiceConnection` directly in the existing codes.

Only `sendToConnection` will be sent to `HubContext` and I redirected it to `ServiceConnection` in this PR to fit the rules we are following now.

- And also add an `EchoHub` for our `ChatSample`, which will let our customers doing their perf test easier.

### DIFF
--- a/samples/ChatSample/ChatSample/Startup.cs
+++ b/samples/ChatSample/ChatSample/Startup.cs
@@ -28,7 +28,7 @@ namespace ChatSample
             app.UseAzureSignalR(routes =>
             {
                 routes.MapHub<Chat>("/chat");
-                routes.MapHub<BenchHub>("/bench");
+                routes.MapHub<BenchHub>("/signalrbench");
             });
         }
     }
--- a/test/Microsoft.Azure.SignalR.Tests/ServiceLifetimeManagerFacts.cs
+++ b/test/Microsoft.Azure.SignalR.Tests/ServiceLifetimeManagerFacts.cs
@@ -4,12 +4,14 @@
 using System;
 using System.Buffers;
 using System.Collections.Generic;
+using System.Security.Claims;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Connections;
 using Microsoft.AspNetCore.SignalR;
 using Microsoft.AspNetCore.SignalR.Internal;
 using Microsoft.AspNetCore.SignalR.Protocol;
 using Microsoft.Azure.SignalR.Protocol;
+using Microsoft.Azure.SignalR.Tests.Common;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Abstractions;
 using Microsoft.Extensions.Options;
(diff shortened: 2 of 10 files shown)


samples/ChatSample/ChatSample/Startup.cs
```diff
@@ -28,7 +28,7 @@ public void Configure(IApplicationBuilder app)
             app.UseAzureSignalR(routes =>
             {
                 routes.MapHub<Chat>("/chat");
-                routes.MapHub<BenchHub>("/bench");
+                routes.MapHub<EchoHub>("/signalrbench");
```
vicancy: >EchoHub [](start = 30, length = 7)

why change?

terencefan: What's this used for？
Our azure-signalr-bench requires a Hub with simple Echo implementation.
And it needs to be registered as "/signalrbench".

vicancy: benchHub is used for `azure-signalr-bench`

---
In reply to: [350527793](https://github.com/Azure/azure-signalr/pull/744#discussion_r350527793) [](ancestors = 350527793)

test/Microsoft.Azure.SignalR.Tests/ServiceLifetimeManagerFacts.cs
```diff
@@ -189,6 +189,68 @@ public async void ServiceLifetimeManagerOnlyBlazorHubProtocolTest(string functio
             Assert.Equal(1, (serviceConnectionManager.ServiceMessage as MulticastDataMessage).Payloads.Count);
         }
 
+        private sealed class TestServiceConnectionPrivate : TestServiceConnection
```
vicancy: >class [](start = 23, length = 5)

move private class below

test/Microsoft.Azure.SignalR.Tests/ServiceLifetimeManagerFacts.cs
```diff
@@ -189,6 +189,68 @@ public async void ServiceLifetimeManagerOnlyBlazorHubProtocolTest(string functio
...
+            var manager = MockLifetimeManager(serviceConnectionManager, clientConnectionManager);
+
+            await manager.SendConnectionAsync("conn1", "foo", new object[] { 1, 2 });
+
+            Assert.NotNull(connection.last);
+            if (connection.last is MultiConnectionDataMessage m)
+            {
+                Assert.Equal("conn1", m.ConnectionList[0]);
+                Assert.Equal(1, m.Payloads.Count);
+                Assert.True(m.Payloads.ContainsKey(MockProtocol));
```
vicancy: >MockProtocol [](start = 51, length = 12)

payloads should not contain `blazor`

samples/ChatSample/ChatSample/Startup.cs
terencefan: Ok, I got it.
Since the URL has been updated to `/signalrbench` so I will just change this and reuse the `EchoHub`.

test/Microsoft.Azure.SignalR.Tests/ServiceLifetimeManagerFacts.cs
terencefan: Yes, it should.
Our resolver supports 3 different protocols (including a mocked protocol "blazor"),
and this mocked protocol has been designated in the lifetime manager creation options.

PR #747, opened 2019-11-28
Make shutdown for different Hubs run concurrently

terencefan:
Disable it as well since the feature isn't ready.

Related to #689

### DIFF
--- a/src/Microsoft.Azure.SignalR/ServiceRouteBuilder.cs
+++ b/src/Microsoft.Azure.SignalR/ServiceRouteBuilder.cs
@@ -2,6 +2,9 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using System.Linq;
 using Microsoft.AspNetCore.Connections;
 using Microsoft.AspNetCore.Http;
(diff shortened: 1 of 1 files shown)


src/Microsoft.Azure.SignalR/ServiceRouteBuilder.cs
```diff
@@ -18,6 +21,7 @@ public class ServiceRouteBuilder
     {
         private readonly IServiceProvider _serviceProvider;
         private readonly RouteBuilder _routes;
+        private readonly IList<Action<TimeSpan>> _shutdownHooks = new List<Action<TimeSpan>>();
```
vwxyzh: `IList<Func<TimeSpan, Task>>`

src/Microsoft.Azure.SignalR/ServiceRouteBuilder.cs
```diff
@@ -18,6 +21,7 @@ public class ServiceRouteBuilder
     {
         private readonly IServiceProvider _serviceProvider;
         private readonly RouteBuilder _routes;
+        private readonly IList<Func<TimeSpan, Task>> _shutdownHooks = new List<Func<TimeSpan, Task>>();
```
BrennanConroy: nit: Use the concrete List type instead of the interface for private/internal properties

PR #752, opened 2019-12-03
Add MigrationLevel to HandshakeRequest

terencefan:


specs/ServiceProtocol.md
```diff
@@ -100,6 +108,14 @@ MessagePack uses different formats to encode values. Refer to the [MessagePack F
 ```
 - 1 - Message Type, indicating this is a `HandshakeRequest` message.
 - Version - A `Int32` encoding number of the protocol version.
+- ConnectionType - A `Int32` encoding number of the connection type.
+	- 0, Default, it can carry clients, service runtime should always accept this kind of connection.
+	- 1, OnDemand, creating when service requested more connections, it can carry clients, but it may be rejected by service runtime.
+	- 2, Weak, it can not carry clients, but it can send message.
+- MigrationLevel - A `Int32` encoding number indicates if further client connections associated with this server connection could be migrated.
+	- 0, Off, a client connection can not be migrated to another server.
+	- 1, ShutdownOnly, a client connection can be migrated only if the pairing server was shutdown gracefully.
```
vicancy: are we providing too much options to user?

davidfowl: Which user, how does anyone get access to this?

terencefan: Our customer could enable this feature by the `ServiceOptions`, like this:
```C#
services.AddSignalR()
                    .AddAzureSignalR(option =>
                    {
                        option.ServerConnectionMigration = ServiceConnectionMigrationLevel.ShutdownOnly
                    });
```

KKhurin: Add details for the failed / non mainstream scenarios. E.g. we closed the client on the server and could not open on another server - will the client get disconnected? How many tries before we give up?  Or it took too long to connect to another server - how long does the service try? How much does the service buffer? What happens if the client sends more data than we buffer? What happens if during migration client disconnects - do we still migrate and deliver new messages including client disconnect?

src/Microsoft.Azure.SignalR.Protocols/ServiceMessage.cs
```diff
@@ -42,6 +42,18 @@ public class HandshakeRequestMessage : ServiceMessage
...
+        /// Gets or sets the migratable flag.
+        /// <value>
+        /// <list type="bullet">
+        /// <item>0, Off, a client connection can not be migrated to another server.</item>
+        /// <item>1, ShutdownOnly, a client connection can be migrated only if the pairing server was shutdown gracefully.</item>
+        /// <item>2, Any, a client connection can be migrated even if the pairing server connection was dropped accidentally. (may cause data loss)</item>
+        /// </list>
+        /// </value>
+        /// </summary>
+        public int MigrationLevel { get; set; }
```
halter73: Why not make this an enum?

E.g:

```C#
enum MigrationLevel
{
    Off,
    ShutdownOnly,
    Any,
}
```

halter73: I thought I was looking at ServiceOptions. My bad.

PR #759, opened 2019-12-09
Add a lifetime application track to service connection for aspnetcore

vicancy:
It's to solve the issue that the outgoing pipeline closes too early that context.Abort() might not working, we now wait for 5 seconds before shutdown the outgoing task


src/Microsoft.Azure.SignalR/ServerConnections/ClientConnectionContext.cs
```diff
@@ -115,7 +115,7 @@ public void TickHeartbeat()
 
         public ClaimsPrincipal User { get; set; }
 
-        public Task ApplicationTask { get; set; }
```
davidfowl: Tasks already support errors, why not just throw and have it be faulted?

src/Microsoft.Azure.SignalR/ServerConnections/ClientConnectionContext.cs
```diff
@@ -22,6 +22,16 @@
...
+    /// <summary>
+    /// Client Application Layer     Transport      Application
+    /// ========================     =========      ===========
+    /// |                      |     |        |     |         |
+    /// |      |-------|-------|-----|-Input--|-----|--Output |
+    /// |                      |     |        |     |         |
+    /// |      |-------|-------|-----|-Output-|-----|--Input  |
+    /// |                      |     |        |     |         |      
+    /// ========================     =========      ===========
+    /// </summary>
```
terencefan: A good comment, I like it :)

test/Microsoft.Azure.SignalR.Tests/TestHubs/AbortHub.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.SignalR;
+
+namespace Microsoft.Azure.SignalR.Tests
+{
+    public class AbortHub : Hub
```
terencefan: Where did we use this class?

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
```diff
@@ -180,19 +181,67 @@ protected override async Task OnClientMessageAsync(ConnectionDataMessage connect
...
+            else
+            {
+                // transport task ends first, no data will be dispatched out
+                Log.WaitingForApplication(Logger);
+
+                // Wait on the application task to complete
+                connection.CancelApplication(_closeTimeOutMilliseconds);
+                try
+                {
+                    await app;
```
terencefan: So here is the point, our `ApplicationTask` could throw `OperationCancelledException`?

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
```diff
@@ -180,19 +181,67 @@ protected override async Task OnClientMessageAsync(ConnectionDataMessage connect
...
+            // If we aren't already aborted, we send the abort message to the service
+            if (connection.AbortOnClose)
+            {
+                // Inform the Service that we will remove the client because SignalR told us it is disconnected.
+                var serviceMessage =
+                    new CloseConnectionMessage(connection.ConnectionId, errorMessage: exception?.Message);
+                await WriteAsync(serviceMessage);
+                Log.CloseConnection(Logger, connection.ConnectionId);
+            }
+
```
terencefan: There is a `connection.OnCompleted();` that has been called at the end of the `ProcessOutgoingMessagesAsync`.

I suppose that can be moved to here since here is the real end of a connection after refactor.

src/Microsoft.Azure.SignalR/ServerConnections/ClientConnectionContext.cs
KKhurin: Add `->` to show the direction of data flow?

vicancy: `<` has special meaning in triple-slash comments 😢

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
vicancy: yes nice catch

src/Microsoft.Azure.SignalR/ServerConnections/ClientConnectionContext.cs
KKhurin: move out of the ///?

src/Microsoft.Azure.SignalR/ServerConnections/ClientConnectionContext.cs
```diff
@@ -42,12 +52,19 @@ internal class ClientConnectionContext : ConnectionContext,
         private readonly object _heartbeatLock = new object();
         private List<(Action<object> handler, object state)> _heartbeatHandlers;
 
+        private volatile bool _abortOnClose = true;
+        private volatile Task _lifetimeTask;
```
davidfowl: Would a non volatile TCS be better?

src/Microsoft.Azure.SignalR/ServerConnections/ClientConnectionContext.cs
```diff
@@ -42,12 +52,19 @@ internal class ClientConnectionContext : ConnectionContext,
         private readonly object _heartbeatLock = new object();
         private List<(Action<object> handler, object state)> _heartbeatHandlers;
 
+        private volatile bool _abortOnClose = true;
```
davidfowl: Why change to this?

src/Microsoft.Azure.SignalR/ServerConnections/ClientConnectionContext.cs
```diff
@@ -42,12 +52,19 @@ internal class ClientConnectionContext : ConnectionContext,
...
+        private volatile Task _lifetimeTask;
+
         public Task CompleteTask => _connectionEndTcs.Task;
 
         public bool IsMigrated { get; }
 
-        // Send "Abort" to service on close except that Service asks SDK to close
-        public bool AbortOnClose { get; set; } = true;
+        // Send "Abort" to service on close except that Service asks SDK to close
+        public bool AbortOnClose
```
davidfowl: Why doesn't the auto property work?

KKhurin:
Logistically it is all about 3 things:
1. communicating the right message to the service
2. maintaining the list of client connections
3. completing transport pipes

Rather than "making it work" in each method that touches either 3 of the areas above, can we build a few abstractions around these 3 to make them thread safe and unified? So we won't have to worry about keeping all code paths dealing with these 3 in sync?

src/Microsoft.Azure.SignalR/ServerConnections/ClientConnectionContext.cs
vicancy: use emoji arrow....

src/Microsoft.Azure.SignalR/ServerConnections/ClientConnectionContext.cs
vicancy: it can be set from multi-threads

src/Microsoft.Azure.SignalR/ServerConnections/ClientConnectionContext.cs
vicancy: need to guarantee thread safety as the property can be set and read from multiple threads

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
```diff
@@ -418,8 +418,9 @@ private async Task PerformDisconnectAsyncCore(string connectionId)
                 connection.AbortOnClose = false;
 
                 // We're done writing to the application output
-                connection.Application.Output.CancelPendingFlush();
+                connection.Application.Output.Complete();
```
davidfowl: Errr not sure

vicancy: This happens to work because as we are using Unbounded pipeline, `WriteAsync` to this `Connection.Application.Output` is synchronous...... So this line must be triggered after all the `WriteAsync` action....

vicancy: I don't like it either....

vicancy:
Hi @KKhurin I think the issue the PR wants to resolve is one thing: to make the lifetime for a client connection context clean and the pipe completion is controlled by the single consumer. The 3 things you mentioned is out of the scope of this PR. I think we can discuss offline and make other PRs to resolve your concern if there is a need.

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
```diff
@@ -168,53 +167,116 @@ protected override async Task OnClientMessageAsync(ConnectionDataMessage connect
...
+                    // Inform the Service that we will remove the client because SignalR told us it is disconnected.
+                    var serviceMessage =
+                        new CloseConnectionMessage(connection.ConnectionId, errorMessage: exception?.Message);
+                    // when it fails, it means the underlying connection is dropped
+                    // service is responsible for closing the client connections in this case and there is no need to throw
+                    await SafeWriteAsync(serviceMessage);
+                    Log.CloseConnection(Logger, connection.ConnectionId);
+                }
+
+                connection.OnCompleted();
```
terencefan: Maybe we could call `OnComplete` in the finally block?

PR #767, opened 2019-12-13
Add a SafeWrite method as for client's close message

vicancy:
Add a SafeWrite method as for client's close message there is no need to throw, because when the connection is closed, service will be responsible for cleaning up the connected clients


src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -188,17 +188,18 @@ public Task StopAsync()
...
-                throw new ServiceConnectionNotActiveException(errorMessage);
+                throw new ServiceConnectionNotActiveException(_errorMessage);
             }
+        }
 
-            if (Status != ServiceConnectionStatus.Connected)
+        protected async Task<bool> SafeWriteAsync(ServiceMessage serviceMessage)
+        {
+            var errorMessage = _errorMessage;
+            if (!string.IsNullOrEmpty(errorMessage) || Status != ServiceConnectionStatus.Connected)
```
JialinXin: >errorMessage [](start = 38, length = 12)

could here errorMessage different from the one throws?

vicancy: maybe, the latest one matters, sounds make sense too

PR #768, opened 2019-12-13
Create GitHub Actions workflows

vicancy:


.github/workflows/osx.yml
```diff
@@ -0,0 +1,23 @@
...
+     - dev
+jobs:
+  build_osx:
+    runs-on: macOS-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Setup .NET Core
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: 2.2.108
```
JialinXin: shall we cover dotnet core 3.0+

PR #780, opened 2019-12-18
Move "migrated" headers to context items.

terencefan:
According to David's suggestion.

Relate to #739

### DIFF
--- a/src/Microsoft.Azure.SignalR/ServerConnections/ClientConnectionContext.cs
+++ b/src/Microsoft.Azure.SignalR/ServerConnections/ClientConnectionContext.cs
@@ -78,11 +78,6 @@ namespace Microsoft.Azure.SignalR
             ConnectionId = serviceMessage.ConnectionId;
             User = serviceMessage.GetUserPrincipal();
 
-            if (serviceMessage.Headers.TryGetValue(Constants.AsrsMigrateIn, out _))
-            {
-                IsMigrated = true;
-            }
-
             // Create the Duplix Pipeline for the virtual connection
             transportPipeOptions = transportPipeOptions ?? DefaultPipeOptions;
             appPipeOptions = appPipeOptions ?? DefaultPipeOptions;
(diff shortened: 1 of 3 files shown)


vicancy:
I think in `Header` is better as we'd like user to get an explicit idea from `HttpContext` from inside the `Hub` to know if the connection is a migrated one? Put it into `Items` sounds obscure to users.

davidfowl:
Lets change this to a feature. It's much cleaner and type safe. Proposal

```C#
public interface IConnectionMigrationFeature
{
    string MigratedFrom { get; }
}
```

We implement this on the `ClientConnectionContext` and the user can get the type to detect if this connection was migrated.

src/Microsoft.Azure.SignalR/ServerConnections/ClientConnectionContext.cs
```diff
@@ -92,6 +87,13 @@ public ClientConnectionContext(OpenConnectionMessage serviceMessage, Action<Http
             Application = pair.Transport;
 
             HttpContext = BuildHttpContext(serviceMessage);
+
+            if (serviceMessage.Headers.TryGetValue(Constants.AsrsMigrateIn, out var val))
+            {
+                IsMigrated = true;
+                HttpContext.Items.Add(Constants.AsrsMigrateIn, val);
```
halter73: @davidfowl Didn't you want this on ConnectionContext.Items not HttpContext.Items?

```suggestion
                Items.Add(Constants.AsrsMigrateIn, val);
```

davidfowl: Yea but now I want a feature instead.

vicancy: Feature looks good to me, maybe we make it `internal` first.

halter73: Why would we make the feature internal? Are you talking about the feature interface itself?

davidfowl: Right, it needs to be public.

vicancy: `internal` when it is not yet feature-ready, and switch to `public` when the feature is ready

davidfowl: Oh sure, we shouldn't ship an SDK with this feature being internal, that's all I'm saying.

PR #783, opened 2019-12-20
Move complete application.output into the client connection

vicancy:
There can be a race, when `ServiceConnection` tries to `CleanupClientConnections` and completes the client's `Application.Output` while client connection is writing messages to it.


davidfowl:
🎄

PR #788, opened 2019-12-27
Add migration feature interfaces.

terencefan:
Follow the previous discussion in #780

### DIFF
--- a/src/Microsoft.Azure.SignalR/ServerConnections/ClientConnectionContext.cs
+++ b/src/Microsoft.Azure.SignalR/ServerConnections/ClientConnectionContext.cs
@@ -20,6 +20,7 @@ using Microsoft.AspNetCore.Http.Features;
 using Microsoft.AspNetCore.Http.Features.Authentication;
 using Microsoft.AspNetCore.WebUtilities;
 using Microsoft.Azure.SignalR.Protocol;
+using Microsoft.Azure.SignalR.ServerConnections;
 using Microsoft.Extensions.Primitives;
 
 namespace Microsoft.Azure.SignalR
@@ -107,11 +108,6 @@ namespace Microsoft.Azure.SignalR
             ConnectionId = serviceMessage.ConnectionId;
             User = serviceMessage.GetUserPrincipal();
 
-            if (serviceMessage.Headers.TryGetValue(Constants.AsrsMigrateIn, out _))
-            {
-                IsMigrated = true;
-            }
-
             // Create the Duplix Pipeline for the virtual connection
             transportPipeOptions = transportPipeOptions ?? DefaultPipeOptions;
             appPipeOptions = appPipeOptions ?? DefaultPipeOptions;
@@ -124,6 +120,11 @@ namespace Microsoft.Azure.SignalR
             configureContext?.Invoke(HttpContext);
 
             Features = BuildFeatures();
+
+            if (serviceMessage.Headers.TryGetValue(Constants.AsrsMigrateFrom, out _))
+            {
+                IsMigrated = true;
+            }
         }
 
         public void CompleteIncoming()
--- a/src/Microsoft.Azure.SignalR/ServerConnections/ConnectionMigrationFeature.cs
+++ b/src/Microsoft.Azure.SignalR/ServerConnections/ConnectionMigrationFeature.cs
@@ -1,0 +1,15 @@
+﻿namespace Microsoft.Azure.SignalR.ServerConnections
+{
+    internal class ConnectionMigrationFeature : IConnectionMigrationFeature
+    {
+        public string MigrateTo { get; private set; }
+
+        public string MigrateFrom { get; private set; }
+
+        public ConnectionMigrationFeature(string from, string to)
+        {
+            MigrateFrom = from;
+            MigrateTo = to;
+        }
+    }
+}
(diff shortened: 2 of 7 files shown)


src/Microsoft.Azure.SignalR/ServerConnections/ConnectionMigrateInFeature.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace Microsoft.Azure.SignalR.ServerConnections
+{
+    class ConnectionMigrateInFeature : IConnectionMigrateInFeature
```
vicancy: nit: explicitly set `internal`

src/Microsoft.Azure.SignalR/ServerConnections/ClientConnectionContext.cs
```diff
@@ -223,6 +224,11 @@ private FeatureCollection BuildFeatures()
             features.Set<IConnectionIdFeature>(this);
             features.Set<IConnectionTransportFeature>(this);
             features.Set<IHttpContextFeature>(this);
+
+            if (IsMigrated)
+            {
+                features.Set<IConnectionMigrateInFeature>(new ConnectionMigrateInFeature(""));
```
vicancy: create a default ctor? How we get the migrated from?

src/Microsoft.Azure.SignalR/ServerConnections/IConnectionMigrateInFeature.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Microsoft.Azure.SignalR.ServerConnections
+{
+    interface IConnectionMigrateInFeature
```
vicancy: nit: `internal`

davidfowl: Why is it internal?

src/Microsoft.Azure.SignalR/ServerConnections/ConnectionMigrateOutFeature.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace Microsoft.Azure.SignalR.ServerConnections
+{
+    class ConnectionMigrateOutFeature : IConnectionMigrateOutFeature
```
davidfowl: What's this for?

terencefan: A connection can be either migrated-out or migrated-in, this feature will be added to a connection context when a connection is being migrated-out.

davidfowl: So this is for OnDisconnectedAsync and the other is for OnConnectedAsync? Does it need to be 2 features?

terencefan: Yes, these 2 features for `connected` and `disconnected` respectively. It can be merged into 1 feature but I think it could be misled to our user?

terencefan: Since connection migrate-out could happen regardless of if a connection has been migrated-in.

Both a normal connection or a migrated connection could be migrated-out. Maybe our user would like to distinguish them in `OnDisconnectedAsync`

davidfowl: Wouldn't that just mean 2 different properties?

terencefan: So we just leave the `MigrateFrom` to `null` when migration happened on this connection for the first time?

src/Microsoft.Azure.SignalR/ServerConnections/ConnectionMigrationFeature.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace Microsoft.Azure.SignalR.ServerConnections
+{
+    internal class ConnectionMigrationFeature : IConnectionMigrationFeature
+    {
+        public string MigrateTo { get; private set; }
```
vicancy: nit: ~~private set~~

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
```diff
@@ -136,10 +138,7 @@ protected override Task OnClientDisconnectedAsync(CloseConnectionMessage closeCo
             var connectionId = closeConnectionMessage.ConnectionId;
             if (_enableConnectionMigration && _clientConnectionManager.ClientConnections.TryGetValue(connectionId, out var context))
             {
-                if (!context.HttpContext.Request.Headers.ContainsKey(Constants.AsrsMigrateOut))
-                {
-                    context.HttpContext.Request.Headers.Add(Constants.AsrsMigrateOut, "");
-                }
+                context.Features.Set<IConnectionMigrationFeature>(new ConnectionMigrationFeature("", ""));
```
vicancy: what if this connection is migrated in? this one will override the previous one?

test/Microsoft.Azure.SignalR.Tests/ServiceConnectionTests.cs
```diff
@@ -456,8 +457,33 @@ await connection.OnClientConnectedAsyncForTest(new OpenConnectionMessage("foo",
...
-            // but the `migrated` status should remain False (readonly)
             Assert.True(context.IsMigrated);
+
+            var feature = context.Features.Get<IConnectionMigrationFeature>();
+            Assert.NotNull(feature);
+            Assert.Equal("another-server", feature.MigrateFrom);
+        }
+
+        [Fact]
+        public async void TestServiceConnectionForMigratedOut()
```
vicancy: `Task`

PR #796, opened 2020-01-14
fix NegotiateMatcherPolicy potential memory leak.

### LINKED ISSUE #794: NegotiateMatcherPolicy cache can keep growing unbounded
The design of `NegotiateMatcherPolicy` is pretty cool, but it's missing one detail. The collection of endpoints (and thus the endpoint instances used as keys in the cache) can change while the app is running. With the current design this can cause the matcher policy to cache an unbounded number of endpoints.

Some suggestions about how to improve this:

- Use the hub type from `HubMetadata` as the key - the set of `Type` objects doesn't change unless the user has written a plugin system
- Listen to the changes on `EndpointDataSource` and reset the cache when the set of endpoints change

Examples (can probably be simplified for your usage):

- https://github.com/dotnet/aspnetcore/blob/master/src/Http/Routing/src/DataSourceDependentCache.cs
- https://github.com/dotnet/aspnetcore/blob/0a1d68b8f375eb979c5722b8bedd57496749b1f8/src/Http/Routing/src/Matching/DataSourceDependentMatcher.cs#L23

Reach out to me or @davidfowl if you want more info

JialinXin:
Fix #794 

Considering get datasource and check `GetChangeToken().HasChanged` is a little heavy for each requests, and `HubType` could barely modified after app starts which should be able to cover most cases, use it as key should be good.


vicancy:
Is it possible to have a test covering the issue?

src/Microsoft.Azure.SignalR/Startup/NegotiateMatcherPolicy.cs
```diff
@@ -42,9 +43,10 @@ public Task ApplyAsync(HttpContext httpContext, CandidateSet candidates)
             {
                 ref var candidate = ref candidates[i];
                 // Only apply to RouteEndpoint
-                if (candidate.Endpoint is RouteEndpoint)
+                if (candidate.Endpoint is RouteEndpoint routeEndpoint)
                 {
-                    var newEndpoint = _negotiateEndpointCache.GetOrAdd(candidate.Endpoint, CreateNegotiateEndpoint);
+                    var hubType = routeEndpoint.Metadata.GetMetadata<HubMetadata>().HubType;
+                    var newEndpoint = _negotiateEndpointCache.GetOrAdd(hubType, CreateNegotiateEndpoint(routeEndpoint));
```
davidfowl: This change actually ends up making a new endpoint per negotiate request.

PR #798, opened 2020-01-16
[Part2][Live-scale] Status sync

JialinXin:
1. Move status ping to `base` class for all containers know whether there's active clients.
2. Add retry logic for `fin` ping to ensure ASRS received
3. Support `servers` ping to know global server connectivity between all endpoints to ensure new endpoint able to add
4. Sync `RuntimeServicePingMessage` class with ASRS
5. minor fix namespace


src/Microsoft.Azure.SignalR.AspNet/ServerConnections/ServiceConnectionManager.cs
```diff
@@ -117,6 +117,20 @@ public virtual Task<bool> WriteAckableMessageAsync(ServiceMessage serviceMessage
             return _appConnection.WriteAckableMessageAsync(serviceMessage, cancellationToken);
         }
 
+        public Task StartGetServersPingAsync()
```
vicancy: is an `Endpoint` level `Start` enough that there is no need to expose this method to the manager?

JialinXin: For a normal endpoint start, there's no need to get servers ping to sync status, do you suggest to make it common for all?

JialinXin:
Close the PR as separated to a few smaller ones as required.

Issue #799, opened 2020-01-16
 Microsoft.Azure.SignalR.Protocol MessagePack V2 breaking change


marcodemagalhaes:
Hi

When I upgraded messagepack to v2 the Microsoft.Azure.SignalR.Protocols package started to get error because it implements v1

There are plans to migrate to messagepack v2?

Thanks!

vicancy:
As https://github.com/dotnet/aspnetcore/pull/18133#issuecomment-570937428, we will follow the same practice to build MessagePack from source to prevent such issues in the future.

tebeco:
hello @vicancy

i've just tried to attempt the migration too here :
https://github.com/tebeco/azure-signalr/tree/msgpack-v2

* what should be the base branch ? since this SDK has a different lifetime than SignalR
* am i suppose to target `netstandard2.0` only ? or should i cross target `netcoreapp3.1` since aspnetcore3.x is only `netcoreapp3.x`
  also `netcoreapp3.x` might have extension method for `Memory/Span/...` that cannot exist in `netstandard` since these enhancement requires has been made in both runtime and GC as i understand it

current branch state : 
* build ok
* ~ 100+ test KO
* pretty sure i did not migrate the `WriteBytes()` properly (the 2 block with `// REVIEW`) i'm trying time see if i can do a `Span/Slice` and why the Loop+`TryGet`
* some tests seems to fails because of a connection Handshake failure, i'm not sure if the `ConnectionString` i use is supposed to target `Default|Classic|Serverless` or if i can inject multiple connection string

tebeco:
I think that thank to this commit :

I'm down to this :
![image](https://user-images.githubusercontent.com/2266487/72648761-2ce2a400-397c-11ea-84bf-271c0fe3df30.png)

All error here are present because this respository explicitly depends on `MessagePackHubProtocol` from `Microsoft.AspNetCore.SignalR.Protocols.MessagePack` (v3.0.0 <== this might be a good idea to update it, at least to 3.1.0 for LTS), which is `MessagePack 1.7.x`

34 Tests are KO are from `ServiceLifetimeManagerFacts` :
```
    System.TypeInitializationException : The type initializer for 'Microsoft.Azure.SignalR.Tests.ServiceLifetimeManagerFacts' threw an exception.
    ---- System.TypeInitializationException : The type initializer for 'SignalRResolver' threw an exception.
[...]

PR #803, opened 2020-01-17
Update AzureSignalR SDK MessagePack to v2 using SubModule

tebeco:
Addresses : https://github.com/Azure/azure-signalr/issues/799

# Summary of the chagnes :
This PR intends to update MessagePack (from 1.7 to 2.x)

# Reason 
This PR is the equivalent Update for MessagePack (from 1.7 to 2.x) binary serializer after this one :
dotnet/aspnetcore#18133
As this repository as an hard dependency on `dotnet/aspnetcore`, this will need to be addressed

current branch state : 
I'm not sure if the use of `MessagePack...Options` is applicable in this repo (especially for the `Formatter/Resolver`)

# Current thought :

Should these nuget target `netstandard2.0` only ? or should i cross target `netcoreapp3.1` since aspnetcore3.x is only `netcoreapp3.x` (and 3.1 is LTS)
Not sure it would helps here but few Memory Pritmitive might have extension method in `netcoreapp3.x` (that cannot exist in `netstandard2.x`) since these enhancement has been made in the runtime

I'm not sure if the use of `MessagePackSerializerOptions` is applicable in this repo (especially for the `Formatter/Resolver`)

# Need info :
@vicancy

[Can you enlighten me about the comment you linked here](https://github.com/Azure/azure-signalr/issues/799#issuecomment-575500945)
Are you planning to :
* Use MessagePack as a SubModule + build from source
* Update MessagePack as a Major / Breaking Change version ?

I have no idea what branch to use in the repo for the PR repo
The repo lifecycle in not the same as dotnet/aspnetcore
As of today i don't see an obvious solution so i used dev

## I think that thank to `MessagePack 2.x` there is now a way to simplify some code :
https://github.com/tebeco/azure-signalr/commit/2bab21d03270a38d24e7b8c58c488db6d24ac6b2
I truly need some feedback for this one, as, for example,Does it i break Backward compatibility like the parsing that happen inside the `Azure SignalR Service` cluster

I'm down to this :
image

## Hard dependency on `SignalR Core 3.x` and `MessagePack 1.7` in an issue here

All error here are present because this respository explicitly depends on `MessagePackHubProtocol` from `Microsoft.AspNetCore.SignalR.Protocols.MessagePack` (v3.0.0 <== this might be a good idea to update it, at least to 3.1.0 for LTS), which is `MessagePack 1.7.x`

34 Tests are KO are from `ServiceLifetimeManagerFacts` :
```
    System.TypeInitializationException : The type initializer for 'Microsoft.Azure.SignalR.Tests.ServiceLifetimeManagerFacts' threw an exception.
    ---- System.TypeInitializationException : The type initializer for 'SignalRResolver' threw an exception.
    -------- System.MissingFieldException : Field not found: 'MessagePack.Resolvers.DynamicEnumAsStringResolver.Instance'.
  Stack Trace: 
....
    MessagePackHubProtocol.CreateDefaultFormatterResolvers()
    MessagePackHubProtocolOptions.ctor()
    MessagePackHubProtocol.ctor() <============================ THIS
....
```

The 2 other errors are `UseAzureSignalRWithConnectionStringNotSpecified` / `UseAzureSignalRWithInvalidConnectionString`
And the Exception is the same (the `StackTrace` show another constructor, but end up the same way)


vicancy:
Really appreciate your contribution @tebeco. 
> Are you planning to :
      * Use MessagePack as a SubModule + build from source
      * Update MessagePack as a Major / Breaking Change version ?

SDK depends on MessagePack to resolve the service data protocol. My current preference is to `Use MessagePack as a SubModule + build from source` so that we don't need to maintain 2 major versions because of MessagePack.

tebeco:
> SDK depends on MessagePack to resolve the service data protocol. My current preference is to Use MessagePack as a SubModule + build from source so that we don't need to maintain 2 major versions because of MessagePack.

Makes sense ;)
That's the best solution for customer, though the repository will be a bit more complex to maintains / build / keep up to date

## Quick question
> Should i add csproj to the Solution ?
So far i've added `MessagePack.csproj` for simplicity, but it also requires `MessagePack.Annotation.csproj` to build

> The code from `AspNetCore` using the SubModule
 - is only for `Blazor`
 - they are cherry picking file with `*.cs` + `path/to/share/source`
 - `v2.0` changed the game and source files are generated by `.tt` code and are in the `Unity` project (Maybe for `CoreRt` <== i have no idea at all, just a weird guess)

tebeco:
(Sorry for all the `git push --force-with-lease`, i try to have the less possible commits with the change in all the `yml` files

tebeco:
@vicancy,

any idea for why i still hit the `MessagePack` conflict issue ?
The unit test failing are injection various `IHubProtocol` (`Json` and `MessagePack`)
As i referenced the `csproj` for the submodule it might be the reason, but i'm not sure here.
Should i instead use something like this instead ?
```
<... Include="path/to/submobe/messagepack-csharp/messagepack/*.cs" />
<... Include="path/to/submobe/messagepack-csharp/messagepack.unity/*.cs" /> <== because most of the `.cs` file are here now
```
like it's done here :
https://github.com/dotnet/aspnetcore/blob/master/src/Components/Server/src/Microsoft.AspNetCore.Components.Server.csproj#L55-L69

maybe @pranavkm have some insight here

tebeco:
(Rebased on `dev` branch)

tebeco:
@vicancy 
any "pre review" ?
I'm still wondering if doing like blazor (directly référence files) would solve the test issues, will check that Monday or Tuesday.

vicancy:
@tebeco sorry for the late response, the team was having a national festival and was just back to work. With a quick look, I think this PR is ready for review, thanks for all your efforts on this. I will go through this PR today.

tebeco:
@vicancy 

I think my biggest question is this one :
https://github.com/Azure/azure-signalr/pull/803#issuecomment-577618118

As `AspNetCore SignalR` won't be updated until `v5.0.0`, the current `MessagePackProtocol` pull down `MessagePack v1.7.x`, i think i there might be collision in either namespace or types somehow here.
As the goal of this PR is to avoid this specific collision, i wonder what i did wrong and now i suspect my C# knowledge of the type system is angry at me and my brain won't process it :D

vicancy:
I think the version conflict of UT failure is because even we use submodule and project reference, we don't change the assembly name of the `MessagePack` so as a result there is still conflict between the `1.7` version `MessagePack` from UT which we reference and the one we project referenced and build.... 

I agree that file reference can solve the issue, but effort taking.

tebeco:
That's what i feared too yeah.
I can take a look at it, but please triple check it as when i did "toyed" with the SubModule in `AspNetCore` (before reverting it), it was not a fun place.
It seems that the source code is generated by `tt` code and bunch of the `.cs` files are spitted out in the `Unity` project, though there's still 1-2 class in the `MessagePack` project.

I'll probably ping Aaron once done to avoid noise yet

tebeco:
Ok this is less fun than i expected ...

image

tebeco:
🥁 🥁 🥁 🥁 🥁 🥁 🥁

tebeco:
@vicancy 
Is there something wrong going on with the `Gated-Ubuntu` build ?
It seems to be stuck from ~18min, when the others seems to run in ~2 to 4min

src/Microsoft.Azure.SignalR.Protocols/SequenceOfT.cs
```diff
@@ -0,0 +1,465 @@
+﻿// Copyright (c) Andrew Arnott. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
```
AArnott: You'll either need to modify the LICENSE.txt reference to where you can *actually* find the license information, or update LICENSE.txt in the root of this repo to include the license info from the repo that you got this from.

tebeco:
thx for the fast answer,

Regarding point `1.` as you can see in the PR history, this was the first idea too here, but i'm not sure about the branch in this repository until `aspnetcore 5.x` is released.
Also it was still fresh from `dotnet/aspnetcore`, i tried to give a hand here. so that i could be merged soon or later.
Devil advocate :
* This package will depends on `Signalr` 
* A consumer use signalr but not `MessagePackHubProtocol`
* That consumer want `MessagePack v2` on his side of the code, before `November 2020` ... for reasons

Both design are fine for me, that's why i asked for helped and review ;)

src/Microsoft.Azure.SignalR.Protocols/SequenceOfT.cs
```diff
@@ -0,0 +1,465 @@
...
+
+    /// <summary>
+    /// Manages a sequence of elements, readily castable as a <see cref="ReadOnlySequence{T}"/>.
+    /// </summary>
+    /// <typeparam name="T">The type of element stored by the sequence.</typeparam>
+    /// <remarks>
+    /// Instance members are not thread-safe.
+    /// </remarks>
+    [DebuggerDisplay("{" + nameof(DebuggerDisplay) + ",nq}")]
+    public class Sequence<T> : IBufferWriter<T>, IDisposable
```
AArnott: This will need to be made internal if you stick with this build-from-source design.

tebeco:
> If you do remove the package/binary dependency and just compile source directly ala Blazor, you should also set the same DefineConstants that Blazor does. This makes all the public types become internal so that you're not 'leaking' public messagepack types that will collide with the public ones from the official messagepack package.

Thx i totally forgot that ... I had a question about this portion of the code for example :
https://github.com/neuecc/MessagePack-CSharp/blob/v2.0/src/MessagePack.UnityClient/Assets/Scripts/MessagePack/MessagePackReader.cs#L980-L991

```csharp
#if NETCOREAPP2_1
    initializedChars += decoder.GetChars(this.reader.UnreadSpan.Slice(0, bytesRead), charArray.AsSpan(initializedChars), flush);
#else
    unsafe
    {
        fixed (byte* pUnreadSpan = this.reader.UnreadSpan)
        fixed (char* pCharArray = &charArray[initializedChars])
        {
            initializedChars += decoder.GetChars(pUnreadSpan, bytesRead, pCharArray, charArray.Length - initializedChars, flush);
        }
    }
#endif
```

As this project is only `netstandard2.0` if will build the `unsafe {}` block, but it will also do the same if this repo move to `netcoreapp3.1` as the `DefinedConstant` would then be `NETCOREAPP3_1`.
If i remember correctly what i spotted on `dotnet/aspnetcore` if could be change by something like
```
<PropertyGroup Condition="'$(TargetFramworks)' == 'netcoreapp2.1' OR  $(TargetFramworks)' == 'netcoreapp2.2' OR $(TargetFramworks)' == 'netcoreapp3.'">
    <IsAtLeastNetCore21>true</IsAtLeastNetCore21>
<PropertyGroup>
<DefineContstants Condition="'$(IsAtLeastNetCore21)' == 'true'">$(DefineContstants);NETCOREAPP2_1_OR_MORE</DefineContstants>
```

```
#if NETCOREAPP2_1_OR_MORE
                initializedChars += decoder.GetChars(this.reader.UnreadSpan.Slice(0, bytesRead), charArray.AsSpan(initializedChars), flush);
#else
```

AArnott:
Yes, we'd probably take a code change to enhance that `#if` when you need it.

tebeco:
I don't think this repo need it yet as the project referencing it is just `netstandard2.0`, but i happened to spot this will doing it few hours ago and was curious ;)
I wonder how it build in Blazor though

src/Microsoft.Azure.SignalR.Protocols/SequenceOfT.cs
tebeco: I would not have guessed this one ^^

tebeco:
> And as I mentioned before, you're cutting off customers' ability to define their own serializable types using the messagepack attributes or write their own custom formatters/resolvers or even set their own serialization options.

@AArnott 
I totally did not thought about that ... Thank for that one.

Also @vicancy may help here, i'm not sure how the `IServiceProtocol` of `Azure SignalR Service` is "open" to either change or configuration.
I think it will only use `MessagePack` or `Json` based on the `SignalR` Protocol :
`Binary` => `MessagePack`
`Text` => `Json`

src/Microsoft.Azure.SignalR.Protocols/SequenceOfT.cs
tebeco: @vicancy, it's your call here

vicancy:
> And as I mentioned before, you're cutting off customers' ability to define their own serializable types using the messagepack attributes or write their own custom formatters/resolvers or even set their own serialization options.

No `IServiceProtocol` is not open to any customization. `messagepack` referenced by `azure-signalr` is used as an internal format to serialize the `ServiceProtocol` communicating between app server and the SignalR service. It is similar to Blazor that the serialization should be kept internal.

> If you do remove the package/binary dependency and just compile source directly ala Blazor, you should also set the same DefineConstants that Blazor does. This makes all the public types become internal so that you're not 'leaking' public messagepack types that will collide with the public ones from the official messagepack package.

Yes nice catch.

src/Microsoft.Azure.SignalR.Protocols/SequenceOfT.cs
vicancy: `LICENSE.txt` to `LICENSE`? (https://github.com/Azure/azure-signalr/blob/dev/LICENSE)

src/Microsoft.Azure.SignalR.Protocols/SequenceOfT.cs
AArnott: This one would conflict with the public type in the Nerdbank.Streams package.

src/Microsoft.Azure.SignalR.Protocols/SequenceOfT.cs
AArnott: And I guess we need to merge in a copyright/license from the source repository (Nerdbank.Streams in this case) or put this in 3rdpartynotices.txt or something.

src/Microsoft.Azure.SignalR.Protocols/SequenceOfT.cs
tebeco: Yeah, to be fair i'm not yet confident with `build from source` as this is something I never used before (for obvious reason).
It seems completely logic and obvious just like the `#if PRIVATE_BUILD` and `$(DefinedConstant)`
it's just not a reflex yet.

src/Microsoft.Azure.SignalR.Protocols/SequenceOfT.cs
tebeco: @AArnott, can you check the 2 last commits to be sure this is what you meant ?
@vicancy is these 2 ok for you ?

I understand there is a duplicate of information, is it too much or not ?

tebeco: I created a small PR here https://github.com/AArnott/Nerdbank.Streams/pull/159
as `NerdBank.Streams` contains `LICENSE` and not `LICENSE.txt` just in case someone wonder about that

tebeco:
What ???


## Net461 :

```
2020-02-04T17:54:33.0048064Z   [xUnit.net 00:01:48.87]     Microsoft.Azure.SignalR.AspNet.Tests.RunAzureSignalRTests.TestRequestsWithRunAzureSignalR(providerType: typeof(Microsoft.Azure.SignalR.AspNet.Tests.RunAzureSignalRTests+NullUserIdProvider), expectedUser: null) [FAIL]
2020-02-04T17:54:33.0051687Z     X Microsoft.Azure.SignalR.AspNet.Tests.RunAzureSignalRTests.TestRequestsWithRunAzureSignalR(providerType: typeof(Microsoft.Azure.SignalR.AspNet.Tests.RunAzureSignalRTests+NullUserIdProvider), expectedUser: null) [1m 40s]
2020-02-04T17:54:33.0055830Z     Error Message:
2020-02-04T17:54:33.0056298Z      System.Threading.Tasks.TaskCanceledException : A task was canceled.
```

## netcoreapp3.0

```
2020-02-04T17:54:42.1005583Z   [xUnit.net 00:00:02.12]     Microsoft.Azure.SignalR.Tests.ClientConnectionManagerTests.TestAllClientConnectionsCompleted [FAIL]
2020-02-04T17:54:42.1006692Z     X Microsoft.Azure.SignalR.Tests.ClientConnectionManagerTests.TestAllClientConnectionsCompleted [1s 13ms]
2020-02-04T17:54:42.1007669Z     Error Message:
2020-02-04T17:54:42.1008378Z      Assert.Equal() Failure
2020-02-04T17:54:42.1009412Z   Expected: 0
2020-02-04T17:54:42.1011364Z   Actual:   1
2020-02-04T17:54:42.1011605Z     Stack Trace:
2020-02-04T17:54:42.1013277Z        at Microsoft.Azure.SignalR.Tests.ClientConnectionManagerTests.TestAllClientConnectionsCompleted() in D:\a\azure-signalr\azure-signalr\test\Microsoft.Azure.SignalR.Tests\ClientConnectionManagerTests.cs:line 35
2020-02-04T17:54:42.1925150Z   [xUnit.net 00:00:02.22]     Microsoft.Azure.SignalR.Tests.ServiceHubDispatcherTests.TestShutdown [FAIL]
2020-02-04T17:54:42.4282020Z     X Microsoft.Azure.SignalR.Tests.ServiceHubDispatcherTests.TestShutdown [1s 118ms]
2020-02-04T17:54:42.4283872Z     Error Message:
2020-02-04T17:54:42.4286930Z      Assert.True() Failure
2020-02-04T17:54:42.4288725Z   Expected: True
2020-02-04T17:54:42.4290206Z   Actual:   False
2020-02-04T17:54:42.4291734Z     Stack Trace:
2020-02-04T17:54:42.4293325Z        at Microsoft.Azure.SignalR.Tests.ServiceHubDispatcherTests.TestShutdown() in D:\a\azure-signalr\azure-signalr\test\Microsoft.Azure.SignalR.Tests\ServiceHubDispatcherTests.cs:line 36
2020-02-04T17:54:42.4293969Z      at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__139_0(Object state)
```

@vicancy 
is there a bot or anything else than a commit to re-trigger the build here ? I'm not sure how this happened.  
The first one seems like a timeout or unexpected connection clause  
The others are weird, i supposed the ASR ConnectionString is still valid, else the errors would be way more dramatic and massive, i don't know what to do here

vicancy:
Should be some known random ut failure because only OSX fails, rerun can help. I guess only admin can manually rerun the check. A new commit can help retrigger the build.

tebeco:
Unless i missed anything in the PR, or i might said i would do and forgot about it i probably won't touch anything else in that PR i think

src/Microsoft.Azure.SignalR.Protocols/Microsoft.Azure.SignalR.Protocols.csproj
```diff
@@ -1,21 +1,48 @@
...
+    <Compile Include="..\Common\BinaryMessageParser.cs" Link="Internal\BinaryMessageParser.cs" />
+    <Compile Include="..\Common\MemoryBufferWriter.cs" Link="Internal\MemoryBufferWriter.cs" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Primitives" Version="$(MicrosoftExtensionsPrimitivesPackageVersion)" />
+    <PackageReference Include="System.Buffers" Version="$(SystemBuffersPackageVersion)" />
+    <PackageReference Include="System.Memory" Version="$(SystemMemoryPackageVersion)" />
+
+    <PackageReference Include="Microsoft.VisualStudio.Validation" Version="$(MicrosoftVisualStudioValidationPackageVersion)" />
```
vicancy: `<PrivateAssets>all</PrivateAssets>` to prevent the reference from exposing?

.github/workflows/osx.yml
```diff
@@ -12,8 +12,10 @@ jobs:
   build_osx:
     runs-on: macOS-latest
     steps:
-    - uses: actions/checkout@v1
-    - name: Build with dotnet
-      run: "./build.sh --ci /p:DisableNet461Tests=true"
+    - uses: actions/checkout@v2
+
+    - name: Update SubModules
+      run: git submodule update --init --recursive
```
vicancy: could you also update https://github.com/Azure/azure-signalr/blob/dev/docs/build-source.md#clone-the-source-code to include `submodule` setup similar to:
> Use --recursive to clone all submodules.
> `git clone https://github.com/Azure/azure-signalr --recursive`
> If you have already cloned, run this to initialize all submodules.
> `git submodule update --init`

src/Microsoft.Azure.SignalR.Protocols/Microsoft.Azure.SignalR.Protocols.csproj
AArnott: Given this requires a runtime assembly, `PrivateAssets=all` would generate runtime failures. Use `PrivateAssets=compile` instead so that others don't see it as a reference assembly but it still ships with the app.

tebeco: humm ok i though it was kinda like a `Debug.Assert` so that it would not be built/output in `Release` configuration.

the  2 questions, what the general idea aside the if check > fail
do i still needs a `PrivateAsset` then if it's useful at runtime ? is it a good/bad idea to let this dependency be visible as a transitive ? (probably not important i guess then ?

i'll have to reopen the docs on PrivateAssets to get a better understanding at it ^^
especially ... what if it's set on "Compile" and some use it too (yeah i have a weird mind i try to break everything before it even work)

AArnott: `PrivateAssets="compile"` prevents your consumers from seeing the types inside it. They can still add a package ref to it themselves and to reference it if they want. The primary 'win' from using this attribute is if you later drop the reference altogether and folks update to a later package, they won't have source breaks from the dropped reference if they had started actually *using* the library directly.

That said, I suspect you only need this reference to your SequenceOfT.cs file, and it would be relatively trivial for you to rewrite the few lines that need it to just validate arguments directly if you want to drop this reference altogether.

3rdPartyNotices.txt
```diff
@@ -0,0 +1,26 @@
+Portions of this repo (The file SequenceOfT.cs) come from the https://github.com/AArnott/Nerdbank.Streams/ project,
+Exact commit : https://github.com/AArnott/Nerdbank.Streams/blob/23063b83512b3478848e514075e01cafed279955/src/Nerdbank.Streams/Sequence%601.cs
```
AArnott: This is good. I think you should include another license notice for MessagePack as well.

tebeco: I changed a little bit how the `3rdPartNotice` file is "scaffolded", is there a standard to respect.
Is that ok like this ?

Side question: Do you know why github seems to detect the `LICENSE` from `NerdBank.Streams` but not from `MessagePack` ?
https://github.com/AArnott/Nerdbank.Streams/blob/23063b83512b3478848e514075e01cafed279955/LICENSE
https://github.com/neuecc/MessagePack-CSharp/blob/2e421efb7bc53956fc66356565f9402d3271a588/LICENSE
image

.github/workflows/osx.yml
tebeco: oO sorry i did not think about that, done

3rdPartyNotices.txt
tebeco: (can you resolve this thread if it's ok for you ?)

src/Microsoft.Azure.SignalR.Protocols/Microsoft.Azure.SignalR.Protocols.csproj
vicancy: +1 to "rewrite the few lines inside SequenceOfT and completely drop the reference"

3rdPartyNotices.txt
AArnott: >  Do you know why github seems to detect the LICENSE from NerdBank.Streams but not from MessagePack ?

Yes: the messagepack one isn't the standard license text. It has two licenses in the same file.

AArnott: I don't see a "Resolve Conversation" button any more on this thread. But I would hit it if I could.

src/Microsoft.Azure.SignalR.Protocols/Microsoft.Azure.SignalR.Protocols.csproj
tebeco: done

3rdPartyNotices.txt
tebeco: (resolving from here then ^^ thx)

tebeco:
Done ?
image

build/dependencies.props
```diff
@@ -37,6 +36,8 @@
     <MicrosoftExtensionHttpVersion>2.2.0</MicrosoftExtensionHttpVersion>
     <MicrosoftAspNetCoreSignalRProtocolsMessagePackPackageVersion>1.0.11</MicrosoftAspNetCoreSignalRProtocolsMessagePackPackageVersion>
 
+    <MicrosoftVisualStudioValidationPackageVersion>15.5.31</MicrosoftVisualStudioValidationPackageVersion>
```
vicancy: This can be removed

tebeco:
@vicancy
`1h30` to get 108MB `net472` tooling :( is that 20KB/s ?
I have a VERY VERY bad idea to test locally to workaround that (as it could harm developer on this repo)

```
2020-02-07T15:31:24.1721020Z [0;36m>>> wget --quiet --tries 10 -O /tmp/tmp.lLHWX1H1g5 https://aspnetcore.blob.core.windows.net/buildtools/netfx/4.7.2/netfx.4.7.2.tar.gz[0m
2020-02-07T17:01:29.6322875Z Extracting .NET Framework reference assemblies to /home/runner/.dotnet/buildtools/netfx/4.7.2
```

tebeco:
yeay \o/
thx everyone

PR #806, opened 2020-01-19
[Part2.2][Live-scale] sync RuntimeServicePingMessage

JialinXin:
update ServicePingMessage to sync with ASRS.


src/Microsoft.Azure.SignalR.Common/ServiceMessages/RuntimeServicePingMessages.cs
```diff
@@ -0,0 +1,109 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.Azure.SignalR.Protocol;
+using ServicePingMessage = Microsoft.Azure.SignalR.Protocol.PingMessage;
+
+namespace Microsoft.Azure.SignalR
+{
+    public static class RuntimeServicePingMessage
```
vicancy: internal, never make class/interface public until they are intended to be exposed to customers in SDK

JialinXin: Nice catch.

PR #807, opened 2020-01-20
[Part2.3][Live-scale] Support migratable ping and add retry

JialinXin:


src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -301,6 +309,39 @@ protected virtual ServiceConnectionStatus GetStatus()
...
+            else
+            {
+                await c.WriteAsync(_shutdownFinMessage);
+            }
+        }
+
+        protected async Task RemoveConnectionAsync(IServiceConnection c, bool migratable)
+        {
+            var retry = 0;
+            using var source = new CancellationTokenSource();
```
vicancy: move init into the loop, and add some test?

JialinXin: Fixed cancellation token source. And leave migration related tests for @terencefan. live-scale related tests will be added after merged most PRs for complete e2e tests.

PR #808, opened 2020-01-20
[Part2.4][Live-scale] Support servers ping and update statusPing for all

JialinXin:
1. move status ping which return global client connection status to base container to help detect whether good to drop server connection when remove endpoint
2. support `servers` ping when adding endpoint to help detect new endpoint server connection status.


src/Microsoft.Azure.SignalR.Common/Interfaces/IServiceConnectionContainer.cs
```diff
@@ -19,8 +20,16 @@ internal interface IServiceConnectionContainer
...
+
+        Task StopGetServersPingAsync();
+
         ServiceConnectionStatus Status { get; }
 
         Task ConnectionInitializedTask { get; }
+
+        HashSet<string> GlobalServerIds { get; }
+
+        bool HasClients { get; }
```
vicancy: will these properties be used outside of the container? Is private propeties of `ContainerBase` enough?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -279,11 +314,25 @@ public async Task<bool> WriteAckableMessageAsync(ServiceMessage serviceMessage,
...
+        public Task StartGetServersPingAsync()
+        {
+            _addingEndpoint = true;
+            return Task.CompletedTask;
+        }
+
+        public Task StopGetServersPingAsync()
+        {
+            _addingEndpoint = false;
+            GlobalServerIds.Clear();
```
vicancy: >Clear [](start = 28, length = 5)

this is not thread safe

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -143,10 +162,26 @@ protected async Task StartCoreAsync(IServiceConnection connection, string target
...
+            if (RuntimeServicePingMessage.TryGetStatus(pingMessage, out var status))
+            {
+                Log.ReceivedServiceStatusPing(Logger, status, Endpoint);
+                HasClients = status;
+            }
+            else if (RuntimeServicePingMessage.TryGetServerIds(pingMessage, out var serverIds, out var updatedTime))
+            {
+                if (updatedTime > _serverIdsLastUpdated)
+                {
+                    GlobalServerIds = serverIds;
```
vicancy: >GlobalServerIds [](start = 20, length = 15)

this is also not thread safe.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -279,11 +314,25 @@ public async Task<bool> WriteAckableMessageAsync(ServiceMessage serviceMessage,
                     // should not be hit.
                     return false;
             }
+        }
+
+        public Task StartGetServersPingAsync()
+        {
+            _addingEndpoint = true;
```
vicancy: >_addingEndpoint [](start = 12, length = 15)

maybe move the logic to another timer is clearer

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -282,16 +314,34 @@ public async Task<bool> WriteAckableMessageAsync(ServiceMessage serviceMessage,
...
+        }
+
+        public virtual Task OfflineAsync(bool migratable)
+        {
+            return Task.WhenAll(FixedServiceConnections.Select(c => RemoveConnectionAsync(c, migratable)));
+        }
+
+        public Task StartGetServersPing()
+        {
+            _serverIdsPing = new CustomizedPingTimer(Logger, Constants.CustomizedPingTimer.Servers, WriteServerIdsPingAsync, DefaultServersPingInterval, DefaultServersPingInterval);
```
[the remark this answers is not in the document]
JialinXin: then it requires to have start interfaces? this is only used in multiple endpoints cases, do we need to make it initialized for all?

---
In reply to: [381782935](https://github.com/Azure/azure-signalr/pull/808#discussion_r381782935) [](ancestors = 381782935)

src/Microsoft.Azure.SignalR.Common/ServiceConnections/WeakServiceConnectionContainer.cs
```diff
@@ -96,53 +88,6 @@ internal bool GetServiceStatus(bool active, int checkWindow, TimeSpan checkTimeS
...
-            {
-                _timer.Stop();
-            }
-
-            base.Dispose(disposing);
-        }
-
         private static class Log
         {
             private static readonly Action<ILogger, double, Exception> _startingServiceStatusPingTimer =
```
JialinXin: >_startingServiceStatusPingTimer [](start = 71, length = 31)

will remove some logs.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -389,21 +445,157 @@ private IEnumerable<IServiceConnection> CreateFixedServiceConnection(int count)
...
+                _dueTime = dueTime;
+                _intervalTime = intervalTime;
+                _defaultPingTicks = intervalTime.Seconds * Stopwatch.Frequency;
+
+                _timer = Init();
+            }
+
+            public void Start()
+            {
+                if (_counter == 0)
```
vicancy:
```cs
if (Interlocked.Increment(ref _counter) == 1){
...
}
```
so that the whole decision is atomic.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -389,21 +445,157 @@ private IEnumerable<IServiceConnection> CreateFixedServiceConnection(int count)
...
+
+                _lastSendTimestamp = Stopwatch.GetTimestamp();
+                var timer = new TimerAwaitable(_dueTime, _intervalTime);
+
+                return timer;
+            }
+
+            private async Task PingAsync(TimerAwaitable timer)
+            {
+                using (timer)
```
vicancy: remove `using` and let `Dispose` dispose timer?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -389,21 +445,157 @@ private IEnumerable<IServiceConnection> CreateFixedServiceConnection(int count)
...
+                if (_counter == 1)
+                {
+                    _timer.Stop();
+                }
+                Interlocked.Decrement(ref _counter);
+            }
+
+            public void Dispose()
+            {
+                if (_counter > 0)
```
vicancy: shall we always stop ` _timer` when `Dispose`?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -389,21 +445,157 @@ private IEnumerable<IServiceConnection> CreateFixedServiceConnection(int count)
...
+                {
+                    _timer.Start();
+                    _ = PingAsync(_timer);
+                }
+                Interlocked.Increment(ref _counter);
+            }
+
+            public void Stop()
+            {
+                if (_counter == 0)
```
vicancy: also here, 

```cs
if (Interlocked.Decrement(ref _counter) == 0)
{...}

```

JialinXin: Considering avoid wrong stop to break further operation, use long type to be able to use Interlocked.Read( ) for check first.

PR #814, opened 2020-02-04
Release graceful shutdown feature.

terencefan:


tebeco:
@terencefan

Can you add a detailed description of what the changes does.
I'm only a consumer here, but i'm curious to know what this achieve as it might be related to some issue we have at work when trying to do deployment of multiple Hub Instances backed by the same `Azure SignalR Service`

terencefan:
@tebeco 

In the case of you turned it on, instead of dropping server connections immediately, our SDK will wait until the last client connection has been dropped (or reaches a pre-defined timeout), and then cut-off server connections to prevent data-missing.

We called this procedure `graceful shutdown`. In common cases, you have to trigger a message (broadcast) to all your clients and let them drop client connections by themselves to benefit from this feature.

We will also introduce another feature called `server connection migration` which will also benefit from this.

In the case of this feature has been enabled. After a shutdown request has been sent to ASRS service, we will try migrating the influenced client connections to another valid server connection (to prevent client connection from being disconnected). Though the process will not always succeed, it will bring our customers a better connection quality, especially when they are trying to rolling-update their app servers.

tebeco:
sooo if i understand correctly it will enable a real 0 down time from the point of view of the client ?

the only solution so far was yo rely on the reconnect feature but that would still be both downtime + data loss

thx you so much for that

the only question i have is about that "Broadcasted message to disconnect the client"
is that something that we would need to do on our own ? or will it be part of the client lib for signalr ?

because if you send a message and it's lost or unhandled or discarded, the client will have an impact right ?

i understand that the client lib is in another repo and team (dotnet/aspnetcore/src/signalr/...)
but would that make sense to do something similar there or not ?

src/Microsoft.Azure.SignalR/ServiceOptions.cs
```diff
@@ -63,21 +63,21 @@ public class ServiceOptions : IServiceEndpointOptions
         /// Specifies if server will shutdown gracefully. 
         /// Default value is false.
         /// </summary>
-        internal bool EnableGracefulShutdown { get; set; } = false;
+        public bool EnableGracefulShutdown { get; set; } = false;
```
vicancy, line 66: I am thinking, if providing a sub option class clearer, e.g. GracefulShutdownOptions, or MigrationOptions?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -65,6 +66,16 @@ internal abstract class ServiceConnectionBase : IServiceConnection
 
         public event Action<StatusChange> ConnectionStatusChanged;
 
+        public GracefulShutdownMode ShutdownMode { get; set; } = GracefulShutdownMode.Off;
+
+        internal HandshakeRequestMessage HandshakeRequest
+        {
+            get {
+                _handshakeRequest.MigrationLevel = ShutdownMode == GracefulShutdownMode.OnAndMigrateConnections ? 1 : 0;
```
vicancy, line 74: this can be determined in ctor.

terencefan: ShutdownMode is not given in the ctor.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServerConnectionMigrationLevel.cs
```diff
@@ -1,9 +1,9 @@
 ﻿namespace Microsoft.Azure.SignalR
 {
-    internal enum ServerConnectionMigrationLevel
+    public enum GracefulShutdownMode
     {
         Off = 0,
-        ShutdownOnly = 1,
-        Any = 2,
+        On = 1,
+        OnAndMigrateConnections = 1,
```
vicancy, line 7: both `1`? typo`?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
vicancy: Ah.. this is tricky, use ctor instead.

tebeco:
is it possible not to amend/force push so that it's possible to see various change during the PR from a public point of view ?
I'm very interested by this PR in order to use it too, but each notification points to a kind of 404, and it's hard almost impossible to understand the "last change" because there's only 1 commit / 30 file change

I suppose that maintainer could Squash on MErge, or just before merge you could squash too ?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -27,7 +27,8 @@ internal abstract class ServiceConnectionBase : IServiceConnection
         private static readonly long DefaultKeepAliveTicks = DefaultKeepAliveInterval.Seconds * Stopwatch.Frequency;
 
         private readonly ReadOnlyMemory<byte> _cachedPingBytes;
-        private readonly HandshakeRequestMessage _handshakeRequest;
+
+        internal readonly HandshakeRequestMessage _handshakeRequest;
```
vicancy, line 31: consider a better way

docs/use-signalr-service.md
```diff
@@ -125,6 +125,24 @@ They can be accessed at [`Hub.Context.User`](https://github.com/aspnet/SignalR/b
...
     - When `Required`, Azure SignalR always tries to find the app server which the client first negotiates with. This options can be useful when some client context is fetched from `negotiate` step and stored in memory, and then to be used inside `Hub`s. However, this option may have performance drawbacks because it requires Azure SignalR to take additional efforts to find this particular app server globally, and to keep globally routing traffics between client and server.
 
+
+#### `GracefulShutdown`
+
+##### Mode
+
+- Default value is `Off`
+- When set to `WaitForClientsClose`, instead of stopping the server immediately, we remove it from the ASRS to prevent new client connections from being assigned to this server.
+- When set to `MigrateClients`, in addition to the steps we mentioned in `WaitForClientsClose`, we will try migrate client connections to another valid server at a proper time.
```
vicancy, line 137: try **migrating**

docs/use-signalr-service.md
```diff
@@ -125,6 +125,24 @@ They can be accessed at [`Hub.Context.User`](https://github.com/aspnet/SignalR/b
...
+#### `GracefulShutdown`
+
+##### Mode
+
+- Default value is `Off`
+- When set to `WaitForClientsClose`, instead of stopping the server immediately, we remove it from the ASRS to prevent new client connections from being assigned to this server.
+- When set to `MigrateClients`, in addition to the steps we mentioned in `WaitForClientsClose`, we will try migrate client connections to another valid server at a proper time. 
+  We could make sure the migration will only happen on the message boundaries, which means each of old/new servers will receive intact messages.
+  `OnConnected` and `OnDisconnected` will be triggered when connections be migrated in/out, and a `IConnectionMigrationFeature` will be set to help you determine if the connection has been migrated.
+  Please visit our sample codes for detail usage.
```
vicancy, line 140: also provide the link of the line of sample code using this IConnectionMigrationFeature?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/GracefulShutdownMode.cs
```diff
@@ -1,9 +1,41 @@
...
+        /// It happens on the message boundaries, considering if each of your message consist of 3 packages. The migration will happen at here:
+        /// 
+        /// | P1 - P2 - P3 | [HERE] | P4 - P5 - P6 |
+        /// | Message 1    |        | Message 2    |
+        ///
+        /// We do this by finding message boundaries on-fly,
+        /// For JSON protocol, we simply find seperators (,)
+        /// For MessagePack protocol, we preserve the length header and count body length to determine if the message was finished.
+        /// 
+        /// This mode always works well with context-free scenarios.
```
vicancy, line 35: nice written up 👍

PR #818, opened 2020-02-11
[Part3.1][Live-scale] Minor refactor prepare for live-scale

JialinXin:


src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
```diff
@@ -33,23 +39,21 @@ internal class MultiEndpointServiceConnectionContainer : IServiceConnectionConta
             if (generator == null)
             {
                 throw new ArgumentNullException(nameof(generator));
-            }
-
+            }
+
+            _router = router ?? throw new ArgumentNullException(nameof(router));
```
vicancy: when _endpoints is 1, it is ok to not have a `_router`?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
```diff
@@ -33,23 +39,21 @@ internal class MultiEndpointServiceConnectionContainer : IServiceConnectionConta
...
-                _router = router ?? throw new ArgumentNullException(nameof(router));
-                ConnectionContainers = _endpoints.ToDictionary(s => (ServiceEndpoint)s, s => generator(s));
-            }
+            foreach (var endpoint in _endpoints)
+            {
+                _connectionContainers[endpoint] = generator(endpoint);
+            }
+
+            // assign first element to _inner for quick access in single endpoint case
+            _inner = _connectionContainers.First().Value;
```
vicancy: `First()` -> `[0]`

vicancy: it throws when _endpoints.Length == 0

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
```diff
@@ -92,24 +96,24 @@ public Task ConnectionInitializedTask
         {
             get
             {
-                if (_inner != null)
+                if (!_needRouter)
```
vicancy: `_needRouter` need to be in sync with `_inner` otherwise it is not correct. when `_connectionContainers` can change, we need to make sure all the connections added, adding to the container are in control.

src/Microsoft.Azure.SignalR.Common/Interfaces/IMultiEndpointServiceConnectionContainer.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.Azure.SignalR
+{
+    internal interface IMultiEndpointServiceConnectionContainer : IServiceConnectionContainer
+    {
+        public IReadOnlyDictionary<ServiceEndpoint, IServiceConnectionContainer> ConnectionContainers { get; }
```
vicancy: why need to expose the property? maybe exposing method like `GetContainer(ServiceEndpoint endpoint)` is safer

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
JialinXin: In current logic, DefaultRouter will be added anyway.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
JialinXin: Offline synced to remove _inner to work in same path.

src/Microsoft.Azure.SignalR.Common/Interfaces/IMultiEndpointServiceConnectionContainer.cs
JialinXin: Only used in tests. Currently `IMultiEndpointServiceConnectionContainer` is not directly used by outside `MultiEndpointServiceConnectionContainer`, remove the property is good.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
JialinXin: offline synced to change the way to a little.

JialinXin:
>         private Task WriteMultiEndpointMessageAsync(ServiceMessage serviceMessage, Func<IServiceConnectionContainer, Task> inner)

@lianwei@microsoft.com Please be aware and confirm now both single/multiple will not throw `ServiceConnectionNotActiveException` when write failure.

---
Refers to: src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs:189 in 8445e76. [](commit_id = 8445e760f2eeaab8b7db57f7cb45edd7c34996b1, deletion_comment = False)

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
```diff
@@ -13,15 +14,20 @@
...
+    {
+        private readonly ConcurrentDictionary<ServiceEndpoint, IServiceConnectionContainer> _connectionContainers =
+               new ConcurrentDictionary<ServiceEndpoint, IServiceConnectionContainer>();
+
         private readonly IMessageRouter _router;
         private readonly ILogger _logger;
-        private readonly IServiceConnectionContainer _inner;
 
-        private IReadOnlyList<HubServiceEndpoint> _endpoints;
+        // use lock when modify below fields to ensure synced.
```
vicancy: >use lock [](start = 11, length = 8)

can use Tuple to avoid lock

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
```diff
@@ -71,7 +75,7 @@ ILoggerFactory loggerFactory
 
         public IEnumerable<ServiceEndpoint> GetOnlineEndpoints()
         {
-            return ConnectionContainers.Keys.Where(s => s.Online);
+            return _connectionContainers.Keys.Where(s => s.Online);
```
vicancy: >Keys [](start = 41, length = 4)

for concurrent dict, use `dict.Where(s=>s.Key...)`, `.Keys` lock the dict instead of the bucket

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
```diff
@@ -22,9 +22,8 @@ internal class MultiEndpointServiceConnectionContainer : IMultiEndpointServiceCo
         private readonly IMessageRouter _router;
         private readonly ILogger _logger;
 
-        // use lock when modify below fields to ensure synced.
-        private volatile IReadOnlyList<HubServiceEndpoint> _endpoints;
-        private volatile bool _needRouter;
+        // <needRouter, endpoints>
+        private volatile Tuple<bool, IReadOnlyList<HubServiceEndpoint>> _routerEndpoints;
```
vicancy: `(bool needRouter, IReadOnlyList<HubServiceEndpoint> endpoints)`

PR #820, opened 2020-02-12
add sequence id field to ConnectionDataMessage

terencefan:


src/Microsoft.Azure.SignalR.Protocols/ConnectionMessage.cs
```diff
@@ -136,5 +136,10 @@ public ConnectionDataMessage(string connectionId, ReadOnlySequence<byte> payload
         /// Gets or sets the binary payload.
         /// </summary>
         public ReadOnlySequence<byte> Payload { get; set; }
+
+        /// <summary>
+        /// Gets or sets the sequence id.
+        /// </summary>
+        public int SeqId { get; set; } = 0;
```
vicancy: nit: naming: use full name `SequenceId`?

src/Microsoft.Azure.SignalR.Protocols/ServiceProtocol.cs
```diff
@@ -526,12 +527,15 @@ private static CloseConnectionMessage CreateCloseConnectionMessage(ref MessagePa
             return new CloseConnectionMessage(connectionId, errorMessage, headers);
         }
 
-        private static ConnectionDataMessage CreateConnectionDataMessage(ref MessagePackReader reader)
+        private static ConnectionDataMessage CreateConnectionDataMessage(ref MessagePackReader reader, int arrayLength)
         {
             var connectionId = ReadString(ref reader, "connectionId");
+            var seqId = arrayLength > 3 ? ReadInt32(ref reader, "sequenceId") : 0;
```
vicancy: `-1` as default?

src/Microsoft.Azure.SignalR.Protocols/ConnectionMessage.cs
```diff
@@ -136,5 +136,10 @@ public ConnectionDataMessage(string connectionId, ReadOnlySequence<byte> payload
         /// Gets or sets the binary payload.
         /// </summary>
         public ReadOnlySequence<byte> Payload { get; set; }
```
vicancy: please revert `submodule` updates

src/Microsoft.Azure.SignalR.Protocols/ConnectionMessage.cs
KKhurin: long? Or we're ready to deal with wrapping around max int?

terencefan: I don't think a client connection would hold over 0.2b messages in its lifecycle?

davidfowl: It should be a long

davidfowl: Also the property should be SequenceId

KKhurin: 2bln messages is just ~20 days with ~1K messages/second. Might be not that common but not unrealistic at all.

PR #821, opened 2020-02-14
Fix client request culture

### LINKED ISSUE #813: CurrentCulture changed to en-US
Hi,

I've already opened an issue [here ](https://github.com/dotnet/aspnetcore/issues/18508).
After establishing a Azure SignalR Connection within my Blazor App the culture is suddenly changed to en-US. 
Could you please look into this?
Is there any way to set set the culture for Azure SignalR service?

Thanks
Sven

JialinXin:
Fix #813

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/Constants.cs
+++ b/src/Microsoft.Azure.SignalR.Common/Constants.cs
@@ -60,6 +60,7 @@ namespace Microsoft.Azure.SignalR
         {
             public const string OriginalPath = "asrs.op";
             public const string ConnectionRequestId = "asrs_request_id";
+            public const string RequestCulture = "asrs_request_culture";
         }
     }
 }
--- a/src/Microsoft.Azure.SignalR.Common/Constants.cs
+++ b/src/Microsoft.Azure.SignalR.Common/Constants.cs
@@ -61,7 +61,7 @@ namespace Microsoft.Azure.SignalR
         {
             public const string OriginalPath = "asrs.op";
             public const string ConnectionRequestId = "asrs_request_id";
-            public const string RequestCulture = "asrs_request_culture";
+            public const string RequestCulture = "asrs_lang";
         }
     }
 }
--- a/src/Microsoft.Azure.SignalR/HubHost/NegotiateHandler.cs
+++ b/src/Microsoft.Azure.SignalR/HubHost/NegotiateHandler.cs
@@ -6,8 +6,10 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Net;
 using System.Security.Claims;
+using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Connections;
+using Microsoft.AspNetCore.Localization;
 using Microsoft.AspNetCore.SignalR;
 using Microsoft.Extensions.Options;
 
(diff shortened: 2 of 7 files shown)


src/Microsoft.Azure.SignalR/HubHost/NegotiateHandler.cs
```diff
@@ -63,19 +66,23 @@ public NegotiationResponse Process(HttpContext context, string hubName)
             };
         }
 
-        private string GetQueryString(string originalQueryString)
+        private string GetQueryString(string originalQueryString, string cultureName)
```
vicancy: >cultureName [](start = 73, length = 11)

how about using the same pattern as ConnectionRequestIdProvider that we provide the default provider to get culture from IRequestCultureFeature?

src/Microsoft.Azure.SignalR.Common/Constants.cs
```diff
@@ -61,6 +61,7 @@ public static class QueryParameter
         {
             public const string OriginalPath = "asrs.op";
             public const string ConnectionRequestId = "asrs_request_id";
+            public const string RequestCulture = "asrs_request_culture";
```
vicancy: >asrs_request_culture [](start = 50, length = 20)

maybe a shorter name...

src/Microsoft.Azure.SignalR/ServerConnections/ClientConnectionContext.cs
```diff
@@ -290,6 +294,17 @@ private HttpContext BuildHttpContext(OpenConnectionMessage message)
             }
 
             return new DefaultHttpContext(httpContextFeatures);
+        }
+
+        private static void SetCurrentThreadCulture(string queryString)
```
vicancy: >SetCurrentThreadCulture [](start = 28, length = 23)

merge with the OriginalPath parse to avoid passing twice?

src/Microsoft.Azure.SignalR.Common/Constants.cs
JialinXin: any suggestions? asrs_culture? asrs_request_lang? asrs_lang?

src/Microsoft.Azure.SignalR/HubHost/NegotiateHandler.cs
JialinXin: It seems a little complicated to use provider way if it's internal only and is rely on request culture anyway. On the other hand, if customer could use AspNetCore request culture support there seems no need to use ours. So I think current implementation should be OK.

src/Microsoft.Azure.SignalR.Common/Constants.cs
JialinXin: Offline synced let's use `asrs_lang`.

src/Microsoft.Azure.SignalR/ServerConnections/ClientConnectionContext.cs
```diff
@@ -303,12 +306,41 @@ internal static bool TryGetRemoteIpAddress(IHeaderDictionary headers, out IPAddr
...
+        {
+            if (!string.IsNullOrEmpty(cultureName))
+            {
+                try
+                {
+                    var requestCulture = new RequestCulture(cultureName);
+                    CultureInfo.CurrentCulture = requestCulture.Culture;
+                    CultureInfo.CurrentUICulture = requestCulture.UICulture;
+                }
+                catch (Exception)
```
vicancy: nit `catch` is enough?

PR #827, opened 2020-02-18
Releasing 1.4.0

vicancy:


AndreyBnc:
@vicancy any plan on updating the nuget packages. We are affected by the MessagePack v2 which we saw was released in v1.3.0 but still no package for that version or this one in Nuget.

Thanks

PR #832, opened 2020-02-25
[Part3.2][Live-scale] add interfaces for scale ServiceEndpoint

JialinXin:


src/Microsoft.Azure.SignalR.Common/Endpoints/HubServiceEndpoint.cs
```diff
@@ -16,5 +18,15 @@ public HubServiceEndpoint(string hub, IServiceEndpointProvider provider, Service
         public string Hub { get; }
 
         public IServiceEndpointProvider Provider { get; }
+
+        /// <summary>
+        /// Task waiting for HubServiceEndpoint turn ready to add negotiation, use when live add ServiceEndpoint
+        /// </summary>
+        public TaskCompletionSource<bool> Ready { get; set; }
```
vicancy: make it readonly and immutable, set the value when `ctor` instead

src/Microsoft.Azure.SignalR/EndpointProvider/ServiceEndpointManager.cs
```diff
@@ -60,6 +60,12 @@ private void OnChange(ServiceOptions options)
 
             var updatedEndpoints = GetChangedEndpoints(Endpoints);
 
+            await RenameSerivceEndpoints(updatedEndpoints.RenamedEndpoints);
```
vicancy: can it throw and affect the below call? does throw matter?

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpointManagerBase.cs
```diff
@@ -96,15 +98,127 @@ protected ServiceEndpoint[] GetValuableEndpoints(IEnumerable<ServiceEndpoint> en
...
+        }
+
+        private async Task AddHubServiceEndpointAsync(HubServiceEndpoint endpoint)
+        {
+            Log.StartAddingEndpoint(_logger, endpoint.Endpoint, endpoint.Name);
+            endpoint.Ready = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
+
+            OnAdd?.Invoke(endpoint);
+            // Wait for new endpoint turn Ready
+            await endpoint.Ready.Task;
```
vicancy: who handles if it hung?

vicancy: who handles if it hung?

src/Microsoft.Azure.SignalR/EndpointProvider/ServiceEndpointManager.cs
JialinXin: Basically, rename/add/remove are separate, I think we can try-catch inside with some logs for exceptions to ensure each operation won't affect each other.

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpointManagerBase.cs
JialinXin: Suppose in Container side will set true finally. But to ensure unblock, we may add timeout check here as well. But in this case I think it probably means connection may not even starts, and shall we block add negotiation as well?

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpointManagerBase.cs
```diff
@@ -96,15 +100,199 @@ protected ServiceEndpoint[] GetValuableEndpoints(IEnumerable<ServiceEndpoint> en
...
+            }
+            return Task.CompletedTask;
+        }
+
+        private HubServiceEndpoint CreateHubServiceEndpoint(string hub, ServiceEndpoint endpoint, bool needScaleTcs = false)
+        {
+            var provider = GetEndpointProvider(endpoint);
+
+            TaskCompletionSource<bool> scaleTcs = null;
+            if (needScaleTcs)
```
vicancy: this makes things complicated, so the consumers need to check null everytime when call it? can we unify the conditions?

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpointManagerBase.cs
```diff
@@ -96,15 +100,199 @@ protected ServiceEndpoint[] GetValuableEndpoints(IEnumerable<ServiceEndpoint> en
...
+        {
+            if (endpoints.Count > 0)
+            {
+                try
+                {
+                    var hubEndpoints = CreateHubServiceEndpoints(endpoints, true);
+
+                    // TODO: update local store for negotiation
+
+                    await Task.WhenAll(hubEndpoints.Select(e => RemoveHubServiceEndpointAsync(e, timeout)));
```
vicancy: why not timeout here but inside every single task?

JialinXin: Cause it's originally planned for single endpoint scale. Considering parallel do task, I've added timeout check in `ServiceEndpointManager` side which could reduce by pass timeout all the way, and use the CancellationToken inside to clean the task and ensure process won't be blocked.

src/Microsoft.Azure.SignalR.Common/Endpoints/HubServiceEndpoint.cs
```diff
@@ -1,20 +1,40 @@
...
         public string Hub { get; }
 
         public IServiceEndpointProvider Provider { get; }
+
+        /// <summary>
+        /// Task waiting for HubServiceEndpoint turn ready when live add/remove endpoint
+        /// </summary>
+        public Task ScaleTask => _scaleTcs?.Task ?? Task.CompletedTask;
+
+        public void SetScaleTaskComplete()
```
vicancy: nit naming: `CompleteScale`?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
```diff
@@ -242,6 +250,51 @@ private Task WriteMultiEndpointMessageAsync(ServiceMessage serviceMessage, Func<
             return Task.WhenAll(routed);
         }
 
+        private void OnAdd(HubServiceEndpoint endpoint)
+        {
+            if (!endpoint.Hub.Equals(_hubName, StringComparison.OrdinalIgnoreCase))
+            {
+                return;
```
vicancy: Can it ever happen?

vicancy: throw instead if it is defense code?

JialinXin: Yes. In `ServiceEndpointManager` side endpoint is not hub distinguished, after split to hub level, it will trigger multiple times scale in Container side. So need to identify which is correct one by compare hub.

PR #836, opened 2020-03-03
[Part3.3][Live-scale] Implement rename endpoint.

JialinXin:


vicancy:
>         public ServiceEndpoint[] Endpoints { get; protected set; }

I think this one should be the one to be reset every time when the config changes, while `_endpointsPerHub` is the cache that should only be updated by `GetEndpoints(string hub)`

---
Refers to: src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpointManagerBase.cs:24 in 9524ab7. [](commit_id = 9524ab759b40426a487712546078d6423972a67f, deletion_comment = False)

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
```diff
@@ -292,7 +293,35 @@ private void OnRename(HubServiceEndpoint endpoint)
...
                 return;
             }
-            // TODO: update local store names
+
+            Log.StartRenamingEndpoint(_logger, endpoint.Endpoint);
+            var oldEndpoint = _connectionContainers.FirstOrDefault(e => e.Key.Endpoint == endpoint.Endpoint).Key;
+
+            if (oldEndpoint != null)
+            {
+                _connectionContainers.TryRemove(oldEndpoint, out var container);
```
vicancy: > _connectionContainers.TryRemove(oldEndpoint, out var container); [](start = 15, length = 65)

After this is executed, messages routed to this `oldEndpoint` is lost. I think it is ok to keep this entry and don't remove it?

JialinXin:
>         public ServiceEndpoint[] Endpoints { get; protected set; }

yes, `Endpoints` is updated right after configuration reload in `ServiceEndpointManager`. [https://github.com/Azure/azure-signalr/blob/dev/src/Microsoft.Azure.SignalR/EndpointProvider/ServiceEndpointManager.cs#L62](https://github.com/Azure/azure-signalr/blob/dev/src/Microsoft.Azure.SignalR/EndpointProvider/ServiceEndpointManager.cs#L62).
And `_endpointsPerHub` is the value actually in use when get negotiation endpoints, so need to be update relate to the result of connection setup.

---
In reply to: [593779589](https://github.com/Azure/azure-signalr/pull/836#issuecomment-593779589) [](ancestors = 593779589)

---
Refers to: src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpointManagerBase.cs:24 in 9524ab7. [](commit_id = 9524ab759b40426a487712546078d6423972a67f, deletion_comment = False)

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
JialinXin: If not remove, there'll be duplicated endpoints in cache and leads to duplicated messages like endpoint name not specified when broadcasts. Since container is no change, so old message should be able to processed and new messages should be able to route by new names. 

---
In reply to: [386811805](https://github.com/Azure/azure-signalr/pull/836#discussion_r386811805) [](ancestors = 386811805)

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpointManagerBase.cs
```diff
@@ -214,6 +214,39 @@ private Task RenameHubServiceEndpoint(HubServiceEndpoint endpoint)
...
+            foreach (var hubEndpoint in _endpointsPerHub)
+            {
+                var updatedEndpoints = endpoints.Where(e => e.Hub == hubEndpoint.Key).ToList();
+                var oldEndpoints = hubEndpoint.Value;
+                var newEndpoints = new List<HubServiceEndpoint>();
+                switch (scaleOperation)
+                {
+                    case ScaleOperation.Rename:
+                        newEndpoints = oldEndpoints.Except(updatedEndpoints, new HubServiceEndpointWeakComparer()).ToList();
+                        newEndpoints.AddRange(updatedEndpoints);
```
vicancy: >newEndpoints [](start = 24, length = 12)

there is no need to propagate the change to MultiEndpointContainer, if the container holds the same `HubServiceEndpoint` object

JialinXin: Good catch. Need to update the logic a little to avoid HubServiceEndpoint total replacement in ServiceEndpointManager side, so that the logic in container could be removed entirely for rename part.

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpointManagerBase.cs
```diff
@@ -206,12 +205,43 @@ private async Task RemoveHubServiceEndpointAsync(HubServiceEndpoint endpoint, Ca
             endpoint.CompleteScale();
         }
 
-        private Task RenameHubServiceEndpoint(HubServiceEndpoint endpoint)
+        private void UpdateNegotiationEndpointsStore(IReadOnlyList<HubServiceEndpoint> endpoints, ScaleOperation scaleOperation)
         {
-            Log.StartRenamingEndpoint(_logger, endpoint.Endpoint, endpoint.Name);
+            foreach (var hubEndpoint in _endpointsPerHub)
+            {
+                var updatedEndpoints = endpoints.Where(e => e.Hub == hubEndpoint.Key).OrderBy(e => e.Endpoint).ToList();
```
vicancy: it looks complex, can we simplify it?

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpointManagerBase.cs
```diff
@@ -21,7 +21,7 @@ internal abstract class ServiceEndpointManagerBase : IServiceEndpointManager
         private readonly ILogger _logger;
 
         // Filtered valuable endpoints from ServiceOptions
-        public ServiceEndpoint[] Endpoints { get; protected set; }
+        public HashSet<ServiceEndpoint> Endpoints { get; private set; } = new HashSet<ServiceEndpoint>(ServiceEndpointWeakComparer.Instance);
```
vicancy: ~~= new HashSet<ServiceEndpoint>(ServiceEndpointWeakComparer.Instance);~~

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpointManagerBase.cs
```diff
@@ -95,65 +95,79 @@ protected ServiceEndpoint[] GetValuableEndpoints(IEnumerable<ServiceEndpoint> en
...
+            return new HashSet<ServiceEndpoint>(groupedEndpoints, ServiceEndpointWeakComparer.Instance);
         }
 
-        protected async Task AddServiceEndpointsAsync(IReadOnlyList<ServiceEndpoint> endpoints, CancellationToken cancellationToken)
+        protected async Task ReloadServiceEndpointsAsync(ServiceEndpoint[] serviceEndpoints, TimeSpan scaleTimeout)
         {
-            if (endpoints.Count > 0)
+            var endpoints = GetValuableEndpoints(serviceEndpoints);
+            // Follow same check but log error only
+            if (endpoints.Count == 0)
```
vicancy: move these checks into `GetValuableEndpoints` and catch exceptions to log?

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpointManagerBase.cs
```diff
@@ -197,48 +211,75 @@ private async Task RemoveHubServiceEndpointAsync(HubServiceEndpoint endpoint, Ca
...
 
-        private void UpdateNegotiationEndpointsStore(IReadOnlyList<HubServiceEndpoint> endpoints, ScaleOperation scaleOperation)
+        private void UpdateEndpoints(HashSet<ServiceEndpoint> updatedEndpoints,
+            out HashSet<ServiceEndpoint> addedEndpoints,
+            out HashSet<ServiceEndpoint> removedEndpoints)
         {
-            foreach (var hubEndpoint in _endpointsPerHub)
+            // Provide a copy of Endpoints to calculate removed
+            removedEndpoints = new HashSet<ServiceEndpoint>(Endpoints, ServiceEndpointWeakComparer.Instance);
+            addedEndpoints = new HashSet<ServiceEndpoint>();
```
vicancy: `ServiceEndpointWeakComparer.Instance`?

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpointManagerBase.cs
```diff
@@ -197,48 +211,75 @@ private async Task RemoveHubServiceEndpointAsync(HubServiceEndpoint endpoint, Ca
...
+            return false;
         }
 
-        private sealed class HubServiceEndpointWeakComparer : IEqualityComparer<HubServiceEndpoint>
+        protected sealed class ServiceEndpointWeakComparer : IEqualityComparer<ServiceEndpoint>
         {
-            public bool Equals(HubServiceEndpoint x, HubServiceEndpoint y)
+            public static readonly ServiceEndpointWeakComparer Instance = new ServiceEndpointWeakComparer();
+
+            public ServiceEndpoint Value { get; private set; }
```
vicancy: this is tricky

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpointManagerBase.cs
```diff
@@ -197,48 +211,75 @@ private async Task RemoveHubServiceEndpointAsync(HubServiceEndpoint endpoint, Ca
...
+            {
+                return true;
             }
+
+            cts.Cancel();
+            return false;
         }
 
-        private sealed class HubServiceEndpointWeakComparer : IEqualityComparer<HubServiceEndpoint>
+        protected sealed class ServiceEndpointWeakComparer : IEqualityComparer<ServiceEndpoint>
```
vicancy: move to another class for non `private sealed` classs

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpointManagerBase.cs
vicancy: Use dict instead?

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpointManagerBase.cs
```diff
@@ -258,17 +299,23 @@ private static class Log
...
 
-            private static readonly Action<ILogger, Exception> _failedRemovingEndpoints =
-                LoggerMessage.Define(LogLevel.Error, new EventId(6, "FailedRemovingEndpoints"), "Failed removing endpoints.");
+            private static readonly Action<ILogger, Exception> _timeoutAddEndpoints =
+                LoggerMessage.Define(LogLevel.Error, new EventId(7, "TimeoutAddEndpoints"), "Timeout waiting for adding endpoints.");
 
-            private static readonly Action<ILogger, Exception> _failedRenamingEndpoints =
-                LoggerMessage.Define(LogLevel.Error, new EventId(7, "StartRenamingEndpoints"), "Failed renaming endpoints.");
+            private static readonly Action<ILogger, Exception> _timeoutRemoveEndpoints =
+                LoggerMessage.Define(LogLevel.Error, new EventId(8, "TimeoutRemoveEndpoints"), "Timeout waiting for removing endpoints.");
```
vicancy: nit: `TimeoutRemoving...` or `RemoveEndpointsTimedOut`?

src/Microsoft.Azure.SignalR/EndpointProvider/ServiceEndpointManager.cs
```diff
@@ -48,123 +43,28 @@ public override IServiceEndpointProvider GetEndpointProvider(ServiceEndpoint end
...
-            }
-
-            cts.Cancel();
-            return false;
+            ReloadServiceEndpoints(options.Endpoints);
         }
 
-        private sealed class ServiceEndpointWeakComparer : IEqualityComparer<ServiceEndpoint>
+        // TODO: make public for non hot-reload plans
+        private Task ReloadServiceEndpoints(ServiceEndpoint[] serviceEndpoints)
```
vicancy: if this is to be public, the name should have `Async` for async methods

PR #841, opened 2020-03-06
[Part3.4][Live-scale] Implement add endpoint.

JialinXin:


src/Microsoft.Azure.SignalR/EndpointProvider/ServiceEndpointManager.cs
```diff
@@ -48,32 +48,39 @@ public override IServiceEndpointProvider GetEndpointProvider(ServiceEndpoint end
             return new ServiceEndpointProvider(endpoint, _options);
         }
 
-        private async void OnChange(ServiceOptions options)
+        private void OnChange(ServiceOptions options)
         {
             Log.DetectConfigurationChanges(_logger);
 
+            // synchronize scale for quick and clean status sync
+            ReloadServiceEndpointsAsync(options.Endpoints).Wait();
```
davidfowl: We shouldn't be blocking here...

JialinXin: We'd complete reload endpoints before accept next change. Most of the operations are not able to easy rollback. Add a task to avoid Wait()?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
```diff
@@ -18,58 +18,66 @@ internal class MultiEndpointServiceConnectionContainer : IMultiEndpointServiceCo
...
-
-            _hubName = hub;
+            }
+
+            _hubName = hub;
             _router = router ?? throw new ArgumentNullException(nameof(router));
             _logger = loggerFactory?.CreateLogger<MultiEndpointServiceConnectionContainer>() ?? throw new ArgumentNullException(nameof(loggerFactory));
             _serviceEndpointManager = endpointManager;
-            
+            _scaleTimeout = scaleTimeout ?? TimeSpan.FromSeconds(Constants.DefaultScaleTimeoutInSeconds);
```
vicancy: store `TimeSpan.FromSeconds(Constants.DefaultScaleTimeoutInSeconds)` as static readonly?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
```diff
@@ -240,44 +248,120 @@ private Task WriteMultiEndpointMessageAsync(ServiceMessage serviceMessage, Func<
...
+        }
+
+        private async Task AddHubServiceEndpointAsync(HubServiceEndpoint endpoint)
+        {
+            var container = _generator(endpoint);
+            endpoint.ConnectionContainer = container;
+
+            try
+            {
+                _ = container.StartAsync();
```
vicancy: do we need to await container until it successfully started?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
```diff
@@ -240,44 +248,120 @@ private Task WriteMultiEndpointMessageAsync(ServiceMessage serviceMessage, Func<
...
+            if (!allMatch)
+            {
+                // return directly if local server list is not set yet.
+                return false;
+            }
+
+            // ensure strong consistency of server Ids for new endpoint towards exists
+            foreach (var endpoint in _routerEndpoints.endpoints)
+            {
+                allMatch = endpoint.ConnectionContainer.GlobalServerIds != null
```
vicancy: is there any special usage for `GlobalServerIds`? why isn't this a string or some hash so that we can avoid serialization and complex comparison?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
```diff
@@ -240,44 +248,120 @@ private Task WriteMultiEndpointMessageAsync(ServiceMessage serviceMessage, Func<
...
+        {
+            var startTime = DateTime.UtcNow;
+            while (DateTime.UtcNow - startTime < _scaleTimeout)
+            {
+                if (IsServerReady(container))
+                {
+                    return;
+                }
+                // status ping interval is 5 seconds, delay to do next check
+                await Task.Delay(_scaleWaitIntervalInSeconds * 1000);
```
vicancy: nit: make this a TimeSpan static readonly?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
JialinXin: order is not guaranteed. And to give a full list for some information in server side, for example log.

---
In reply to: [392108114](https://github.com/Azure/azure-signalr/pull/841#discussion_r392108114) [](ancestors = 392108114)

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
JialinXin: not required. it has multiple connections inside and the procedure might be longer then expected. message send during the time would be no harm cause no client is listening.

---
In reply to: [392102714](https://github.com/Azure/azure-signalr/pull/841#discussion_r392102714) [](ancestors = 392102714)

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpointManagerBase.cs
```diff
@@ -155,6 +155,25 @@ private async Task AddServiceEndpointsAsync(IEnumerable<ServiceEndpoint> endpoin
             }
         }
 
+        private void UpdateNegotiationEndpointsStore(IReadOnlyList<HubServiceEndpoint> endpoints, ScaleOperation scaleOperation)
+        {
+            foreach (var hubEndpoint in _endpointsPerHub)
+            {
+                var updatedEndpoints = endpoints.Where(e => e.Hub == hubEndpoint.Key).ToList();
```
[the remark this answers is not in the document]
JialinXin: why? merge 2 dictionary doesn't make things easier.

---
In reply to: [392100425](https://github.com/Azure/azure-signalr/pull/841#discussion_r392100425) [](ancestors = 392100425)

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpointManagerBase.cs
```diff
@@ -155,6 +155,25 @@ private async Task AddServiceEndpointsAsync(IEnumerable<ServiceEndpoint> endpoin
...
+                var oldEndpoints = hubEndpoint.Value;
+                var newEndpoints = oldEndpoints.ToList();
+                switch (scaleOperation)
+                {
+                    case ScaleOperation.Add:
+                        newEndpoints.AddRange(updatedEndpoints);
```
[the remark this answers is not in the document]
JialinXin: ConcurrentDictionary<string, ConcurrentBag>?

---
In reply to: [392101319](https://github.com/Azure/azure-signalr/pull/841#discussion_r392101319) [](ancestors = 392101319)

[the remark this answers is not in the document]
JialinXin: `ConcurrentBag` supports add one by one only which needs iteration updated list. I think add won't be a frequent operation and list AddRange is acceptable.

---
In reply to: [392101319](https://github.com/Azure/azure-signalr/pull/841#discussion_r392101319) [](ancestors = 392101319)

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpointManagerBase.cs
```diff
@@ -155,6 +155,25 @@ private async Task AddServiceEndpointsAsync(IEnumerable<ServiceEndpoint> endpoin
...
+                        break;
+                    default:
+                        break;
+                }
+                _endpointsPerHub.TryUpdate(hubEndpoint.Key, newEndpoints, oldEndpoints);
```
vicancy: only update when `updatedEndpoints` contains elements

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
vicancy: if the server connection is not yet established, will the message send throw ConnectionNotActiveException?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
vicancy: I see this list is a `HashSet` from string ping message? why isn't it just a string? if it is for log there is no need to parse the ping message. Order can be guaranteed when service constructs the server list with order.

src/Microsoft.Azure.SignalR.Common/ServiceMessages/RuntimeServicePingMessages.cs
```diff
@@ -62,18 +62,17 @@ public static bool TryGetStatus(this ServicePingMessage ping, out bool isActive)
...
                 || !long.TryParse(value.Substring(0, value.IndexOf(":")), out updatedTime))
             {
-                serverIds = null;
+                serverIds = string.Empty;
                 updatedTime = DateTime.MinValue.Ticks;
                 return false;
             }
-            var servers = value.Substring(value.IndexOf(":") + 1);
-            serverIds = new HashSet<string>(servers.Split(new char[] { ServerListSeparator }, StringSplitOptions.RemoveEmptyEntries));
+            serverIds = value.Substring(value.IndexOf(":") + 1);
```
vicancy: 1. `value.IndexOf` is calculated twice
2. if `serverIds` is empty for every endpoint, it is a match?

JialinXin: For 2, no, empty will be ignored. Empty is invalid value. Cause if server side could receive the servers ping, it should include itself serverId at least.

PR #842, opened 2020-03-09
[Part3.5][Live-scale] Implement remove endpoint

JialinXin:


src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpointManagerBase.cs
```diff
@@ -155,12 +153,31 @@ private async Task AddServiceEndpointsAsync(IEnumerable<ServiceEndpoint> endpoin
...
 
+        private async Task RemoveServiceEndpointsAsync(IEnumerable<ServiceEndpoint> endpoints, CancellationToken cancellationToken)
+        {
+            if (endpoints.Count() > 0)
+            {
+                try
+                {
+                    var hubEndpoints = CreateHubServiceEndpoints(endpoints, true);
+
+                    UpdateNegotiationEndpointsStore(hubEndpoints, ScaleOperation.Remove);
```
vicancy: the 2 steps `create hse` + `delete` can merge into one that to return the list of `HubServiceEndpoints` when removing

PR #854, opened 2020-03-19
Update doc about AccessTokenAlgorithm option

JialinXin:


docs/use-signalr-service.md
```diff
@@ -84,6 +99,10 @@ The access token is returned in the response to client's negotiate request.
 - When `ServerSentEvent` or `LongPolling` is used as transport, client connection will be closed due to authentication failure after the expire time.
 You can increase this value to avoid client disconnect.
 
+#### `AccessTokenAlgorithm`
+- Default value is `HS256`
+- This option provides choice of [`SecurityAlgorithms`](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/Microsoft.IdentityModel.Tokens/SecurityAlgorithms.cs) when generate access token. Now supported optional values are `HS256` and `HS512`.
```
vicancy: also give some note that `HS512` is more secure but the generated token will be comparatively longer

PR #857, opened 2020-03-23
Refactor about HubDispatcher

terencefan:
According to our latest change for the `.NetCore 3.0`, which is using reflection to get `HubDispathcer` and starting it.

We introduced a singleton instance `ServerLifetimeManager` to register shutdown hooks and account for the **graceful shutdown**.

And also,
1. Some default value changes. (timeout, migration level)
2. Add required properties. (migration level)

### DIFF
--- a/src/Microsoft.Azure.SignalR/DependencyInjectionExtensions.cs
+++ b/src/Microsoft.Azure.SignalR/DependencyInjectionExtensions.cs
@@ -2,12 +2,13 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+#if !NETSTANDARD2_0
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Routing;
+#endif
 using Microsoft.AspNetCore.SignalR;
 using Microsoft.Azure.SignalR;
 using Microsoft.Azure.SignalR.Protocol;
-using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Options;
@@ -75,6 +76,7 @@ namespace Microsoft.Extensions.DependencyInjection
                 .AddSingleton(typeof(IServiceEndpointManager), typeof(ServiceEndpointManager))
                 .AddSingleton(typeof(IServerNameProvider), typeof(DefaultServerNameProvider))
                 .AddSingleton(typeof(ServiceHubDispatcher<>))
+                .AddSingleton(typeof(ServerLifetimeManager))
                 .AddSingleton(typeof(AzureSignalRMarkerService))
                 .AddSingleton<IClientConnectionFactory, ClientConnectionFactory>()
                 .AddSingleton<IHostedService, HeartBeat>()
--- a/src/Microsoft.Azure.SignalR/HubHost/ServerLifetimeManager.cs
+++ b/src/Microsoft.Azure.SignalR/HubHost/ServerLifetimeManager.cs
@@ -1,0 +1,47 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace Microsoft.Azure.SignalR
+{
+    internal class ServerLifetimeManager
+    {
+        private readonly IList<Func<Task>> _shutdownHooks = new List<Func<Task>>();
+
+        private object _lock = new object();
+
+        public ServerLifetimeManager(
+            IServiceProvider provider
+        )
+        {
+#if NETCOREAPP
+            var lifetime = provider.GetService<IHostApplicationLifetime>();
+#elif NETSTANDARD
+            var lifetime = provider.GetService<IApplicationLifetime>();
+#else
+            var lifetime = null;
+#endif
+
+            lifetime?.ApplicationStopping.Register(Shutdown);
+        }
+
+        internal void Register(Func<Task> func)
+        {
+            lock (_lock)
+            {
+                _shutdownHooks.Add(func);
+            }
+        }
+
+        private void Shutdown()
+        {
+            lock (_lock)
+            {
+                Task.WaitAll(_shutdownHooks.Select(func => func()).ToArray());
+            }
+        }
+    }
+}
--- a/src/Microsoft.Azure.SignalR/HubHost/ServerLifetimeManager.cs
+++ b/src/Microsoft.Azure.SignalR/HubHost/ServerLifetimeManager.cs
@@ -21,8 +21,6 @@ namespace Microsoft.Azure.SignalR
             var lifetime = provider.GetService<IHostApplicationLifetime>();
 #elif NETSTANDARD
             var lifetime = provider.GetService<IApplicationLifetime>();
-#else
-            var lifetime = null;
 #endif
 
             lifetime?.ApplicationStopping.Register(Shutdown);
--- a/src/Microsoft.Azure.SignalR/HubHost/ServerLifetimeManager.cs
+++ b/src/Microsoft.Azure.SignalR/HubHost/ServerLifetimeManager.cs
@@ -1,5 +1,5 @@
 ﻿using System;
-using System.Collections.Generic;
+using System.Collections.Concurrent;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.Extensions.DependencyInjection;
@@ -9,9 +9,7 @@ namespace Microsoft.Azure.SignalR
 {
     internal class ServerLifetimeManager
     {
-        private readonly IList<Func<Task>> _shutdownHooks = new List<Func<Task>>();
-
-        private object _lock = new object();
+        private readonly ConcurrentBag<Func<Task>> _shutdownHooks = new ConcurrentBag<Func<Task>>();
 
         public ServerLifetimeManager(
             IServiceProvider provider
@@ -28,18 +26,12 @@ namespace Microsoft.Azure.SignalR
 
         internal void Register(Func<Task> func)
         {
-            lock (_lock)
-            {
-                _shutdownHooks.Add(func);
-            }
+            _shutdownHooks.Add(func);
         }
 
         private void Shutdown()
         {
-            lock (_lock)
-            {
-                Task.WaitAll(_shutdownHooks.Select(func => func()).ToArray());
-            }
+            Task.WaitAll(_shutdownHooks.Select(func => func()).ToArray());
         }
     }
 }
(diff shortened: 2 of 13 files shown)


src/Microsoft.Azure.SignalR/HubHost/ServerLifetimeManager.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace Microsoft.Azure.SignalR.HubHost
+{
+    class ServerLifetimeManager
```
vwxyzh: `internal`

src/Microsoft.Azure.SignalR/HubHost/ServerLifetimeManager.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace Microsoft.Azure.SignalR.HubHost
```
JialinXin: >HubHost [](start = 34, length = 7)

use `Microsoft.Azure.SignalR` namespace and remove this one in `DependencyInjection`

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
```diff
@@ -140,7 +142,7 @@ protected override Task OnClientConnectedAsync(OpenConnectionMessage message)
         protected override Task OnClientDisconnectedAsync(CloseConnectionMessage closeConnectionMessage)
         {
             var connectionId = closeConnectionMessage.ConnectionId;
-            if (_enableConnectionMigration && _clientConnectionManager.ClientConnections.TryGetValue(connectionId, out var context))
+            if (EnableMigration && _clientConnectionManager.ClientConnections.TryGetValue(connectionId, out var context))
```
JialinXin: >EnableMigration [](start = 16, length = 15)

looks EnableMigration only used here, why not directly check `use _migrationLevel != ServerConnectionMigrationLevel.Off` ?

src/Microsoft.Azure.SignalR/HubHost/ServerLifetimeManager.cs
```diff
@@ -0,0 +1,42 @@
...
+        public ServerLifetimeManager(
+            IServiceProvider provider
+        )
+        {
+#if NETCOREAPP
+            var lifetime = provider.GetService<IHostApplicationLifetime>();
+#elif NETSTANDARD
+            var lifetime = provider.GetService<IApplicationLifetime>();
+#else
+            var lifetime = null;
```
JialinXin: could be reduced, as our package supports either NETSTANDARD or NETCOREAPP only? use if/else is enough?

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
```diff
@@ -54,16 +56,16 @@ internal partial class ServiceConnection : ServiceConnectionBase
                                  HubServiceEndpoint endpoint,
                                  IServiceMessageHandler serviceMessageHandler,
                                  ServiceConnectionType connectionType = ServiceConnectionType.Default,
-                                 int closeTimeOutMilliseconds = DefaultCloseTimeoutMilliseconds,
-                                 bool enableConnectionMigration = false) :
-            base(serviceProtocol, serverId, connectionId, endpoint, serviceMessageHandler, connectionType, loggerFactory?.CreateLogger<ServiceConnection>())
+                                 ServerConnectionMigrationLevel migrationLevel = ServerConnectionMigrationLevel.Off,
```
JialinXin: > = ServerConnectionMigrationLevel.Off [](start = 78, length = 37)

probably make default `Off` assigned in `ServiceConnectionFactory` instead of here?

src/Microsoft.Azure.SignalR/DependencyInjectionExtensions.cs
```diff
@@ -89,7 +92,6 @@ private static ISignalRServerBuilder AddAzureSignalRCore(this ISignalRServerBuil
             builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<IStartupFilter, AzureSignalRStartupFilter>());
             builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<MatcherPolicy, NegotiateMatcherPolicy>());
 #endif
-
             return builder;
```
JialinXin: remain the empty line for separate.

src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
```diff
@@ -52,6 +54,11 @@ internal class ServiceHubDispatcher<THub> where THub : Hub
             _clientConnectionFactory = clientConnectionFactory;
             _nameProvider = nameProvider;
             _hubName = typeof(THub).Name;
+
+            if (null != serverLifetimeManager)
```
JialinXin: >if (null != serverLifetimeManager) [](start = 12, length = 34)

could this be null?

src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
```diff
@@ -66,13 +73,18 @@ public void Start(ConnectionDelegate connectionDelegate, Action<HttpContext> con
             _ = _serviceConnectionManager.StartAsync();
         }
 
-        public async Task ShutdownAsync(TimeSpan timeout)
+        internal async Task ShutdownAsync()
```
JialinXin: public? as class level internal already.

src/Microsoft.Azure.SignalR/HubHost/ServerLifetimeManager.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        public ServerLifetimeManager(
+            IServiceProvider provider
+        )
+        {
+#if NETCOREAPP
+            var lifetime = provider.GetService<IHostApplicationLifetime>();
+#elif NETSTANDARD
+            var lifetime = provider.GetService<IApplicationLifetime>();
+#else
```
vicancy: `else throw` for safety

src/Microsoft.Azure.SignalR/HubHost/ServerLifetimeManager.cs
```diff
@@ -0,0 +1,42 @@
...
+            var lifetime = null;
+#endif
+
+            if (lifetime != null)
+            {
+                lifetime.ApplicationStopping.Register(Shutdown);
+            }
+        }
+
+        internal void Register(Func<Task> func)
```
vicancy: does it require thread-safety?

src/Microsoft.Azure.SignalR/HubHost/ServerLifetimeManager.cs
terencefan: I think pointing out which platform/framework we have supported explicitly and throw `Exception` in which we haven't supported is better?

src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
terencefan: In test. I have changed this into `?.`

src/Microsoft.Azure.SignalR/HubHost/ServerLifetimeManager.cs
terencefan: I think that disable the feature is enough in this case.
Not very necessary to throw an exception if our SDK ran in an unsupported framework.

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
terencefan: Much easier to understand if I introduced this property?

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
terencefan: I can do both. And it's necessary here since the previous param is already optional.

src/Microsoft.Azure.SignalR/HubHost/ServerLifetimeManager.cs
JialinXin: I think it's not necessary, it won't be able to compile for unsupported framework. You may check AspNetCore repo, which doesn't has the if/elseif/else check like this.

---
In reply to: [396879153](https://github.com/Azure/azure-signalr/pull/857#discussion_r396879153) [](ancestors = 396879153)

src/Microsoft.Azure.SignalR/HubHost/ServerLifetimeManager.cs
vicancy: it is the build time throw not runtime throw

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
JialinXin: temp variable should be enough here?

---
In reply to: [396882589](https://github.com/Azure/azure-signalr/pull/857#discussion_r396882589) [](ancestors = 396882589)

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
JialinXin: Move it before optional parameter? I notice that if make the parameter optional in ServiceConnectionBase/ServiceConnectionFactory could reduce the assign value in AspNet/Management side which seems not supported the feature yet. So just wonderding if we should do like that. 

---
In reply to: [396884644](https://github.com/Azure/azure-signalr/pull/857#discussion_r396884644) [](ancestors = 396884644)

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
```diff
@@ -23,13 +23,13 @@ internal partial class ServiceConnection : ServiceConnectionBase
...
-        private static readonly Dictionary<string, string> CustomHeader = new Dictionary<string, string> { { Constants.AsrsUserAgent, ProductInfo.GetProductInfo() } };
-        
-        private readonly bool _enableConnectionMigration;
-
+        private static readonly Dictionary<string, string> CustomHeader = new Dictionary<string, string> { { Constants.AsrsUserAgent, ProductInfo.GetProductInfo() } };
+
         private const string ClientConnectionCountInHub = "#clientInHub";
         private const string ClientConnectionCountInServiceConnection = "#client";
 
+        private bool EnableMigration { get; set; }
```
vicancy: field for private ones, `private readonly bool _enableMigration;`

PR #858, opened 2020-03-23
update swagger

vicancy:
1. add `operationId`
2. remove duplicate actions
3. add more response details
4. add `head` action

### DIFF
--- a/docs/swagger/v1.json
+++ b/docs/swagger/v1.json
@@ -1,713 +1,972 @@
 {
-    "openapi": "3.0.1",
-    "info": {
-      "title": "Azure SignalR Service REST API",
-      "version": "v1"
+  "openapi": "3.0.1",
+  "info": {
+    "title": "Azure SignalR Service REST API",
+    "version": "v1"
+  },
+  "paths": {
+    "/api/v1/health": {
+      "head": {
+        "tags": [
+          "general"
+        ],
+        "summary": "Get service health status.",
+        "operationId": "HealthApi_Head_Index",
+        "responses": {
+          "200": {
+            "description": "The service is healthy"
+          }
+        }
+      }
     },
-    "paths": {
-      "/api/v1/health": {
-        "get": {
-          "description": "Get service health status",
-          "tags": [],
-          "responses": {
-            "200": {
-              "description": "Ok"
-            },
-            "502": {
-              "description": "ServiceUnavailable"
-            },
-            "503": {
-              "description": "ServiceUnavailable"
-            }
-          }
-        }
-      },
-      "/api/v1/hubs/{hub}": {
-        "post": {
-          "description": "Broadcast a message to all clients connected to target hub.",
-          "tags": [],
-          "parameters": [
-            {
-              "name": "hub",
-              "in": "path",
-              "required": true,
-              "description": "Target hub name, which should start with alphabetic characters and only contain alpha-numeric characters or underscore.",
-              "schema": {
-                "type": "string"
-              }
-            }
-          ],
-          "requestBody": {
-            "content": {
-              "application/json": {
-                "schema": {
-                  "$ref": "#/components/schemas/PayloadMessage"
-                }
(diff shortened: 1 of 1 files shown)


docs/swagger/v1.json
```diff
@@ -1,710 +1,690 @@
...
-            "content": {
-              "application/json": {
-                "schema": {
-                  "$ref": "#/components/schemas/PayloadMessage"
-                }
-              }
+          }
+        },
+        "responses": {
+          "200": {
```
vwxyzh: should be `202`

docs/swagger/v1.json
```diff
@@ -1,710 +1,690 @@
...
+          "503": {
+            "description": "The service is temporarily unavailable"
+          }
+        }
+      }
+    },
+    "/api/v1/hubs/{hub}": {
+      "post": {
+        "tags": [
+          "HubProxyControllerV1"
```
JialinXin: shall we hide this?

docs/swagger/v1.json
```diff
@@ -1,710 +1,690 @@
...
-              }
+          }
+        },
+        "responses": {
+          "200": {
+            "description": "Success"
+          }
+        }
+      },
+      "get": {
```
vwxyzh: not released...

docs/swagger/v1.json
```diff
@@ -1,710 +1,690 @@
...
-              "description": "Close reason.",
-              "schema": {
-                "type": "string"
-              }
+        ],
+        "responses": {
+          "200": {
+            "description": "Success"
+          },
+          "400": {
```
vwxyzh: should we give 400 in swagger file?

docs/swagger/v1.json
```diff
@@ -1,710 +1,690 @@
...
+            "description": "Bad Request"
+          },
+          "404": {
+            "description": "Not Found"
+          },
+          "500": {
```
vwxyzh: should we give `500` in swagger file?

docs/swagger/v1.json
vicancy: any need to ?

vicancy: > Tagged operations may be handled differently by tools and libraries. For example, Swagger UI uses tags to group the displayed operations

Sounds useful

docs/swagger/v1.json
vicancy: will on-hold the change until the new rt is deployed

docs/swagger/v1.json
vicancy: makes sense

docs/swagger/v1.json
JialinXin: shall we tag with some meaningful names instead of internal controller name, which sounds not useful to customer.

vicancy: updated

docs/swagger/v1.json
```diff
@@ -1,710 +1,975 @@
...
+        "tags": [
+          "general"
+        ],
+        "summary": "Get service health status.",
+        "operationId": "HealthApi_Head_Index",
+        "responses": {
+          "200": {
+            "description": "The service is healthy"
+          },
+          "503": {
```
wanlwanl, line 19: also 502? https://github.com/Azure/azure-signalr/blob/dev/src/Microsoft.Azure.SignalR.Management/ServiceManager.cs#L168

vicancy: swagger describes the service behavior, it does not include cases when service/network is down

vicancy: yeah you are right we missed several expected status code

vicancy: I feel we don't need to list 5xx code, every API can encounter 5xx. And for the client side, it should consider all 5xx code as `isHealthy = false`

Reference: https://stackoverflow.com/a/58277444/687394

PR #860, opened 2020-03-25
[AAD] AccessKey object

terencefan:
Introduce a new object type `AccessKey` to storage the relationship between `AccessKey` and its `KeyId`, even expiration time in the future.

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/Endpoints/AccessKey.cs
+++ b/src/Microsoft.Azure.SignalR.Common/Endpoints/AccessKey.cs
@@ -1,0 +1,17 @@
+﻿namespace Microsoft.Azure.SignalR
+{
+    internal class AccessKey
+    {
+        public string Value { get; set; }
+        public string Id { get; set; }
+
+        public AccessKey(string key, string keyId = null)
+        {
+            Value = key;
+            if (string.IsNullOrEmpty(keyId))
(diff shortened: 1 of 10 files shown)


src/Microsoft.Azure.SignalR.Common/Endpoints/AccessKey.cs
```diff
@@ -0,0 +1,14 @@
...
+{
+    internal class AccessKey
+    {
+        public string Value { get; set; }
+        public string Id { get; set; }
+
+        public AccessKey(string key, string keyId = "signalr")
+        {
+            Value = key;
+            Id = keyId;
```
vicancy: use key.GetHashCode as the default value of `Id` to avoid breaking change? keyId is used as the key of the cache for `SignatureProvider` so a const value is not correct for multiple endpoint case

src/Microsoft.Azure.SignalR.Common/Endpoints/AccessKey.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace Microsoft.Azure.SignalR
+{
+    internal class AccessKey
+    {
+        public string Value { get; set; }
```
vicancy: remove `set`

PR #861, opened 2020-03-27
[AAD] make GetAccessToken async

terencefan:
Since a server-generated `AccessKey` may need to be authorized in the future. Change the `GetAccessToken` methods to **async** (for both server and client connections)

### DIFF
--- a/src/Microsoft.Azure.SignalR.AspNet/EndpointProvider/ServiceEndpointProvider.cs
+++ b/src/Microsoft.Azure.SignalR.AspNet/EndpointProvider/ServiceEndpointProvider.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Net;
 using System.Security.Claims;
 using System.Text;
+using System.Threading.Tasks;
 
 namespace Microsoft.Azure.SignalR.AspNet
 {
@@ -53,13 +54,14 @@ namespace Microsoft.Azure.SignalR.AspNet
             return string.IsNullOrEmpty(applicationName) ? hubName.ToLower() : $"{applicationName.ToLower()}_{hubName.ToLower()}";
         }
 
-        public string GenerateClientAccessToken(string hubName = null, IEnumerable<Claim> claims = null, TimeSpan? lifetime = null)
+        public async Task<string> GenerateClientAccessTokenAsync(string hubName = null, IEnumerable<Claim> claims = null, TimeSpan? lifetime = null)
         {
             var audience = $"{_endpoint}/{ClientPath}";
+            await _accessKey.AuthorizedTask;
             return AuthUtility.GenerateAccessToken(_accessKey, audience, claims, lifetime ?? _accessTokenLifetime, _algorithm);
         }
 
-        public string GenerateServerAccessToken(string hubName, string userId, TimeSpan? lifetime = null)
+        public async Task<string> GenerateServerAccessTokenAsync(string hubName, string userId, TimeSpan? lifetime = null)
         {
             IEnumerable<Claim> claims = null;
             if (userId != null)
@@ -70,6 +72,7 @@ namespace Microsoft.Azure.SignalR.AspNet
                 };
             }
 
+            await _accessKey.AuthorizedTask;
             var audience = $"{_endpoint}/{ServerPath}/?hub={GetPrefixedHubName(_appName, hubName)}";
 
             return AuthUtility.GenerateAccessToken(_accessKey, audience, claims, lifetime ?? _accessTokenLifetime, _algorithm);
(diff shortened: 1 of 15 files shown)


src/Microsoft.Azure.SignalR.AspNet/EndpointProvider/ServiceEndpointProvider.cs
```diff
@@ -53,13 +54,14 @@ private string GetPrefixedHubName(string applicationName, string hubName)
             return string.IsNullOrEmpty(applicationName) ? hubName.ToLower() : $"{applicationName.ToLower()}_{hubName.ToLower()}";
         }
 
-        public string GenerateClientAccessToken(string hubName = null, IEnumerable<Claim> claims = null, TimeSpan? lifetime = null)
+        public async Task<string> GenerateClientAccessToken(string hubName = null, IEnumerable<Claim> claims = null, TimeSpan? lifetime = null)
```
vicancy: append `Async` as a convention for public async methods

src/Microsoft.Azure.SignalR.Common/Endpoints/AccessKey.cs
```diff
@@ -1,10 +1,14 @@
...
+﻿using System.Threading.Tasks;
+
+namespace Microsoft.Azure.SignalR
 {
     internal class AccessKey
     {
         public string Value { get; set; }
         public string Id { get; set; }
 
+        public Task AuthorizedTask = Task.CompletedTask;
```
vicancy: `public` ones use `property` instead of field. `{get;} = Task.CompletedTask`

src/Microsoft.Azure.SignalR.AspNet/EndpointProvider/ServiceEndpointProvider.cs
```diff
@@ -53,13 +54,14 @@ private string GetPrefixedHubName(string applicationName, string hubName)
...
         {
             var audience = $"{_endpoint}/{ClientPath}";
+            await _accessKey.AuthorizedTask;
```
vicancy: could `_accesskey` be null? add some defensive code in ctor?

src/Microsoft.Azure.SignalR.AspNet/Middleware/NegotiateMiddleware.cs
```diff
@@ -175,16 +179,17 @@ private Task ProcessNegotiationRequest(IOwinContext owinContext, HostContext con
 
             try
             {
-                accessToken = provider.GenerateClientAccessToken(null, claims);
+                accessToken = await provider.GenerateClientAccessToken(null, claims);
```
vicancy: nit: extract to another method so that you don't need to change others to `await`

src/Microsoft.Azure.SignalR.Common/Endpoints/AccessKey.cs
```diff
@@ -1,10 +1,14 @@
...
+
         public AccessKey(string key, string keyId = null)
```
JialinXin: >string keyId = null [](start = 37, length = 19)

any chance there's need to manual assign keyId? shall we remove this parameter?

src/Microsoft.Azure.SignalR.Common/Endpoints/AccessKey.cs
```diff
@@ -1,10 +1,14 @@
-﻿namespace Microsoft.Azure.SignalR
+﻿using System.Threading.Tasks;
+
+namespace Microsoft.Azure.SignalR
 {
     internal class AccessKey
     {
         public string Value { get; set; }
         public string Id { get; set; }
```
JialinXin: >set; [](start = 32, length = 4)

Make `Value` and `Id` get only? Whenever Value changes, shall we view it as a new AccessKey?

src/Microsoft.Azure.SignalR.AspNet/EndpointProvider/ServiceEndpointProvider.cs
terencefan: I think this should be promised by `Endpoint`? I could make sure that every `Endpoint` has a non-null **AccessKey**.

src/Microsoft.Azure.SignalR.AspNet/Middleware/NegotiateMiddleware.cs
terencefan: We can't.
Every place where is currently using `AccessToken` should wait for its authorization.

src/Microsoft.Azure.SignalR.AspNet/EndpointProvider/ServiceEndpointProvider.cs
vicancy: we can have some defensive code for this class

sellotape:
This was a breaking change; any reason the version wasn't bumped to 2.x.x?  Perhaps we should generally not expect semantic versioning in this repo?

Also it seems like a `GenerateClientAccessTokenAsync()` could instead have been added; at the moment you now have an async method without the [(by now well-established) convention](https://docs.microsoft.com/en-us/dotnet/standard/asynchronous-programming-patterns/task-based-asynchronous-pattern-tap#naming-parameters-and-return-types) of an "Async" suffix for the name.

vicancy:
@sellotape  thanks for pointing it out, this is a bug we are actively working on. The breaking change is introduced in unexpectedly.

PR #870, opened 2020-04-15
Update messagepack version

vicancy:
In old version 2.0 of `MessagePack`, when message length > 64K, the length is treated as `Int64` for `ReadOnlySequence`, however, according to the spec of `MessagePack`, it should be at most treated as `Int32`. The latest version 2.1.115 fixed the issue.

This issue is causing: message length > 64K is not able to get delivered to service.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "src/submodules/MessagePack-CSharp"]
 	path = src/submodules/MessagePack-CSharp
 	url = https://github.com/neuecc/MessagePack-CSharp
-	branch = v2.0
+	branch = v2.1.115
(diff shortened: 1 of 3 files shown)


BrennanConroy:
Can you link to the problem in the MessagePack library?

vicancy:
There is no explicit issue in MessaegPack, a PR fixing some other issue happened to fixed the issue described here. https://github.com/neuecc/MessagePack-CSharp/commit/d100c66bf99e8491379e19b65c25c3c9b9f6bb50#diff-ed970b3daebc708ce49f55d418075979R742

ajbeaven:
Thanks for pushing this fix out so quickly <3

PR #875, opened 2020-04-16
fix UT for JWT token generator in test

wanlwanl:


vicancy:
This PR can be closed?

PR #876, opened 2020-04-17
Revoke public interface change.

terencefan:


src/Microsoft.Azure.SignalR.Management/ServiceManager.cs
```diff
@@ -143,7 +143,10 @@ public Task<string> GenerateClientAccessToken(string hubName, string userId = nu
             {
                 claimsWithUserId.AddRange(claims);
             }
-            return _endpointProvider.GenerateClientAccessTokenAsync(hubName, claimsWithUserId, lifeTime);
+
+            var task = _endpointProvider.GenerateClientAccessTokenAsync(hubName, claimsWithUserId, lifeTime);
+            task.Wait();
+            return task.Result;
```
terencefan, line 149: This is a quick fix. We don't actually have this method **awaitable** at this moment. 

But it would cause a deadlock if you embed a `Task.Wait` inside an async method.
https://stackoverflow.com/questions/13140523/await-vs-task-wait-deadlock

So try avoiding doing something like this:
```C#
async Task Get(){ var token = manager.GenerateClientAccessToken; }
```
We will provide an async version of this API later on another PR.

terencefan: #861 861

src/Microsoft.Azure.SignalR.Management/ServiceManager.cs
```diff
@@ -143,7 +143,7 @@ public Task<string> GenerateClientAccessToken(string hubName, string userId = nu
             {
                 claimsWithUserId.AddRange(claims);
             }
-            return _endpointProvider.GenerateClientAccessTokenAsync(hubName, claimsWithUserId, lifeTime);
+            return _endpointProvider.GenerateClientAccessTokenAsync(hubName, claimsWithUserId, lifeTime).Result;
```
terencefan, line 146: This is a quick fix of #861 . We don't actually have this method **awaitable** at this moment. 

But it would probably cause a deadlock if you embed a `Task.Wait` inside an async method.
https://stackoverflow.com/questions/13140523/await-vs-task-wait-deadlock

So try avoiding doing something like this:
```C#
async Task Get(){ var token = manager.GenerateClientAccessToken; }
```
We will provide an async version of this API later on another PR.

PR #885, opened 2020-04-22
Message order fix part 1

KKhurin:
In this part we install an async local context each time we process a new OpenConnectionMessage. 
Each client connection will get an individual context created allowing us to carry per-client data (such as the service connection which sent this message) along with the async execution flow.


src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
```diff
@@ -56,90 +57,99 @@ internal partial class ServiceConnection : ServiceConnectionBase
...
+
+            var scopeProperties = new ClientConnectionScopeProperties() 
+            { 
+                ServiceConnection = this
+                // todo add more properties here, e.g. message.Headers.TryGetValue("Tracing", ...)
+            };
+
+            using (new ServiceConnectionScopeInternal(scopeProperties))
+            {
+                _ = ProcessClientConnectionAsync(connection);
```
BrennanConroy, line 135: Does the async local persist in `ProcessClientConnectionAsync` even though it's unset immediately in `Dispose`?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionScopeInternal.cs
```diff
@@ -0,0 +1,63 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Diagnostics;
+using System.Threading;
+
+namespace Microsoft.Azure.SignalR
+{
+    internal class ClientConnectionScopeProperties
```
vicancy, line 10: nit: per class per file

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionScopeInternal.cs
```diff
@@ -0,0 +1,63 @@
...
+    {
+        public IServiceConnection ServiceConnection { get; set; }
+
+        // todo: add additional properties to flow with the scope here
+    }
+
+    internal class ServiceConnectionScopeHolder
+    {
+        private static AsyncLocal<ServiceConnectionScopeHolder> s_AsyncLocalSCC = new AsyncLocal<ServiceConnectionScopeHolder>();
```
vicancy, line 19: nit: naming: `_currentScope`?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionScopeInternal.cs
```diff
@@ -0,0 +1,63 @@
...
+        internal static ServiceConnectionScopeHolder Current
+        {
+            get => s_AsyncLocalSCC.Value;
+            set => s_AsyncLocalSCC.Value = value;
+        }
+
+        public ClientConnectionScopeProperties Properties { get; set; }
+    }
+
+    internal class ServiceConnectionScopeInternal : IDisposable
```
vicancy, line 29: naming: ServiceConnectionScope~~Internal~~?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionScopeInternal.cs
```diff
@@ -0,0 +1,63 @@
...
+namespace Microsoft.Azure.SignalR
+{
+    internal class ClientConnectionScopeProperties
+    {
+        public IServiceConnection ServiceConnection { get; set; }
+
+        // todo: add additional properties to flow with the scope here
+    }
+
+    internal class ServiceConnectionScopeHolder
```
vicancy, line 17: it can be a private inner class for ServiceConnectionScopeInternal ?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionScopeInternal.cs
```diff
@@ -0,0 +1,63 @@
...
+        internal ServiceConnectionScopeInternal(ClientConnectionScopeProperties properties)
+        {
+            if (ServiceConnectionScopeHolder.Current == null)
+            {
+                ServiceConnectionScopeHolder.Current = new ServiceConnectionScopeHolder() { Properties = properties };
+                _needCleanup = true;
+            }
+            else
+            {
+                Debug.Assert(properties == null,
```
vicancy, line 46: shall we allow replacing?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionScopeInternal.cs
```diff
@@ -0,0 +1,63 @@
...
+                    $"Attempt to replace existing connection scope  {ServiceConnectionScopeHolder.Current.Properties.ServiceConnection?.GetHashCode()} with new connection: {properties?.ServiceConnection?.GetHashCode()}");
+            }
+        }
+
+        internal static ServiceConnectionScopeHolder Holder => ServiceConnectionScopeHolder.Current;
```
vicancy, line 51: Is there a need to expose this `Holder` for the scope? I feel `Holder` as some internal implementation of the scope that we can hide from the users

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionScopeInternal.cs
KKhurin: Agree, the current name is weird. I do want the name to signify that it is async local and static though.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionScopeInternal.cs
KKhurin: Agree

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
KKhurin: It persists only inside `ProcessClientConnectionAsync` and its child execution contexts (e.g. Task.Run, QUWI, etc, including task continuations). We don't want it to leak it outside of `ProcessClientConnectionAsync` because it has a chance to spread to other unrelated execution contexts.

I believe there are some subtle ways it could leak though, for example if we awaited the call to `ProcessClientConnectionAsync` and it happened to complete synchronously, in theory the continuation could run on its child frame and Dispose won't be able to clear out the async local in the original frame... In our case we don't await for the call so `Dispose` is guaranteed to run on the same frame as `using`.

KKhurin: To be precise, the async local also persists till the end of `OnClientConnectedAsync` frame but the rest of its execution (after `Dispose`) won't reference scopeProperties so it won't carry large amount of data.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionScopeInternal.cs
vwxyzh: add `readonly`

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
vwxyzh: I think we should use `Task.Run`:
```c#
_ = Task.Run(async () =>
{
  using (...)
  {
    await ...
  }
});
```

KKhurin: I understand the idea is to prevent future changes from accidentally leaking the async local?  Also, we're not awaiting the call to `ProcessClientConnectionAsync` so why should we await it inside `Task.Run`?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionScopeInternal.cs
KKhurin: inner classes?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionScopeInternal.cs
KKhurin: This is a good question. There are properties that we want to carry through the client connection lifetime and not allow user code to change (or even look up). There are also properties that we want to flow with outgoing messages (e.g. tracing context) and various parts of the code should be able to change/restore according to nested scopes.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionScopeInternal.cs
chenkennt: The name is misleading since it's now designed to contain more properties than service connection.

KKhurin: Agree - the scope is not even service bound.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionScopeInternal.cs
KKhurin: Added a separate class `ServiceCommunicationScope<T>` which wraps `ServiceConnectionScopeInternal` and allows nested scopes to replace the current async local values. Derived classes could use it to carry various settings to the protocol level, e.g.:
~~~
    public class TracingScope : ServiceCommunicationScope<bool>
    {
        public TracingScope(bool trace) : base(trace)
        {
        }
    }
~~~

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionScopeInternal.cs
KKhurin: This is internal class and deep at connection level we will need to know if the client connection scope has been established (so when we need to choose or replace the connection we know that it will persist). See https://github.com/Azure/azure-signalr/pull/658/files#diff-72c9ae4e6a93ede091bc3b08bae82afaR299 in the old PR.

The additional user-facing class `ServiceCommunicationScope<T>` (or its sub-classes) does not expose the holder anymore.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionScopeInternal.cs
KKhurin: This is now `ClientConnectionScopeInternal`.
It carries the `ClientConnectionScopeProperties` with `IServiceConnection OutboundServiceConnection` and more properties in the future.

src/Microsoft.Azure.SignalR.Common/ServiceCommunicationScope.cs
```diff
@@ -0,0 +1,48 @@
...
+    /// Represents a generic nestable scope for grouping operations with the same service communication policy.
+    /// Derived classes must ensure that the service protocol can understand and transmit this policy to the service.
+    /// </summary>
+    /// The following is ensured for all send operations to the service within the scope:
+    /// -  the same service connection is used in ServiceTransportType.Persistent mode
+    /// -  the same service connection is used for calls outside of hub call context
+    /// -  only the value of the most nested scope will flow along with the messages
+    /// 
+    /// <typeparam name="TNestableProps">Type representing policy to flow</typeparam>
+    public class ServiceCommunicationScope<TNestableProps> : IDisposable
```
vicancy, line 18: is it `public` intentionally and is designed to be used by our customers?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ClientConnectionScopeInternal.cs
```diff
@@ -0,0 +1,69 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Diagnostics;
+using System.Dynamic;
+using System.Threading;
+
+namespace Microsoft.Azure.SignalR
+{
+
```
vicancy, line 11: remove empty line

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ClientConnectionScopeInternal.cs
```diff
@@ -0,0 +1,69 @@
...
+    internal class ScopePropertiesAccessor<TProps>
```
vicancy, line 12: convention: per class per file, and don't have nested classes except when the nested is private

src/Microsoft.Azure.SignalR.Common/ServiceCommunicationScope.cs
```diff
@@ -0,0 +1,48 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Microsoft.Azure.SignalR.Common
+{
+    /// <summary>
+    /// Represents a generic nestable scope for grouping operations with the same service communication policy.
+    /// Derived classes must ensure that the service protocol can understand and transmit this policy to the service.
```
vicancy, line 10: > Derived classes must ensure that the service protocol can understand and transmit this policy to the service.

What does this mean?

src/Microsoft.Azure.SignalR.Common/ServiceCommunicationScope.cs
```diff
@@ -0,0 +1,48 @@
...
+    public class ServiceCommunicationScope<TNestableProps> : IDisposable
+    {
+        private ScopePropertiesAccessor<TNestableProps> _previousScope;
+        private ClientConnectionScopeInternal _serviceConnectionScope;
+
+        protected ServiceCommunicationScope(TNestableProps properties)
+        {
+            _previousScope = ScopePropertiesAccessor<TNestableProps>.Current;
+            ScopePropertiesAccessor<TNestableProps>.Current = new ScopePropertiesAccessor<TNestableProps>() { Properties = properties };
+            _serviceConnectionScope = new ClientConnectionScopeInternal();
```
vicancy, line 27: looks like the logic in `ClientConnectionScopeInternal`'s ctor is to manipulate `ScopePropertiesAccessor<TNestableProps>.Current`? how about moving the logic into `Current`'s `set` logic and we can remove this class `ClientConnectionScopeInternal`?

src/Microsoft.Azure.SignalR.Common/ServiceCommunicationScope.cs
KKhurin: Customer's won't be able to use it as its only constructor is protected. Its derived class is public so the base has to be public too, e.g. 
~~~
public class TraceScope : ServiceCommunicationScope<bool> 
{}
~~~

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ClientConnectionScopeInternal.cs
KKhurin: Ok!

src/Microsoft.Azure.SignalR.Common/ServiceCommunicationScope.cs
KKhurin: Agree, the wording is confusing/misleading - its not the derived class' responsibility to ensure that the service protocol can understand... What I wanted to say was: if for example `class TraceScope : ServiceCommunicationScope<bool> {..}` is designed to carry a flag signifying whether we should tell the service to trace messages within this scope, then whatever part that uses the service protocol to send messages to the service should check for this scope's flag and transmit this information to the service... I am afraid even this explanation is not clear enough until we start using it in tracing.

src/Microsoft.Azure.SignalR.Common/ServiceCommunicationScope.cs
KKhurin: Hmm... and use something like this:
~~~
try {
  ScopePropertiesAccessor<ClientConnectionScopeProperties>.Current = new ClientConnectionScopeProperties();
  <scope work>
}
finally
{
  ScopePropertiesAccessor<ClientConnectionScopeProperties>.Current = null;
}
~~~
I would not mind but `ClientConnectionScopeInternal` is being used in 2 places and both places would have to take care of correctly handling instantiating/cleaning up the async local (e.g. lines 40-42, 53-57). This class seems like a nice wrapper to handle this logic.

src/Microsoft.Azure.SignalR.Common/ServiceCommunicationScope.cs
vicancy: Customer can write his own derived class? is it expected?

KKhurin: Customers are only expected to use derived class (such as `TraceScope' in example above).
I don't see any good reason why customers would want to derive themselves.

src/Microsoft.Azure.SignalR.Common/ServiceCommunicationScope.cs
vicancy: The implementation of this class looks a little fragile to me that it involves both `ClientConnectionScopeInternal` and `ScopePropertiesAccessor<TNestableProps>.Current` while inside `ClientConnectionScopeInternal` it also manipulates `ScopePropertiesAccessor<TNestableProps>.Current` and the order of calling these two in `ctor` and `dispose` I believe matters to make it work. Why it goes this way depends on how `ClientConnectionScopeInternal` is implemented internally while I don't quite get it why `ServiceCommunicationScope` has something to do with `ClientConnectionScopeInternal`. Maybe directly manipulate `ScopePropertiesAccessor` instead of depending on `ClientConnectionScopeInternal` can make the logic clearer?

src/Microsoft.Azure.SignalR.Common/ServiceCommunicationScope.cs
KKhurin: moving nestable scopes to a separate PR

KKhurin:
Split this to only establish the client connection scope. Will move the NestableScope to a separate PR.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ClientConnectionScopeInternal.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        private bool _needCleanup;
+
+        internal ClientConnectionScopeInternal() : this(default)
+        {
+        }
+
+        protected internal ClientConnectionScopeInternal(ClientConnectionScopeProperties properties)
+        {
+            if (ScopePropertiesAccessor<ClientConnectionScopeProperties>.Current == null)
```
vicancy, line 24: One question, one can override the behavior by setting `ScopePropertiesAccessor<ClientConnectionScopeProperties>.Current` from outside this scope? Is this the concern leading to a nested ClientConnectionScopeProperties class? Is making `ClientConnectionScopeProperties` private and let ctor construct the properties better?

KKhurin: Yes, setting `ScopePropertiesAccessor<ClientConnectionScopeProperties>.Current` would allow nested contexts :( (although this is arguably abuse)  Making `ClientConnectionScopeProperties` private will enforce the non-nestable nature of this class. However it would also force us to add explicit API to change the current values carried by the only copy of `ClientConnectionScopeProperties`. Perhaps this is a better way. Will update the PR shortly!

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ClientConnectionScope.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            if (_needCleanup)
+            {
+                // shallow cleanup since we don't want any execution contexts in unawaited tasks 
+                // to suddenly change behavior once we're done with disposing
+                ScopePropertiesAccessor<ClientConnectionScopeProperties>.Current = null;
+            }
+        }
+
+        internal static bool IsScopeEstablished => ScopePropertiesAccessor<ClientConnectionScopeProperties>.Current != null;
```
KKhurin, line 49: We can postpone adding this property to some later time. We will need it to know if we need to fallback to some other mechanism of establishing *destination_client_for_the_message <-> service_connections_to_use*. For example:
~~~
            if (ClientConnectionScope.IsScopeEstablished)
            {
                //use and/or update ClientConnectionScope.OutboundServiceConnection
            }
            else
            {
                // we don't have async local context - use some less efficient fallback mechanism (cache)
            }
~~~

PR #886, opened 2020-04-23
Diagnostic log option

wanlwanl:


samples/ChatSample/ChatSample/appsettings.json
```diff
@@ -2,7 +2,17 @@
   "Azure": {
     "SignalR": {
       "Enabled": true,
-      "ConnectionString": ""
+      "ConnectionString": "",
+      "DiagnosticLogs": [
```
vicancy, line 6: I am thinking if there is any way to leverage the `Logging` settings https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-3.1

wanlwanl: But this option is about tracking client, and it's only an setting for Azure SignalR. Looks like it's fine to place in `Azure:SignalR` section. Maybe the improve the naming like this:
```
  "Azure": {
    "SignalR": {
      "Enabled": true,
      "ConnectionString": "",
      "TrackingClient": [
        {
          "LogType": "Messaging",
          "Accept": true
        },
        {
          "LogType": "Connectivity",
          "Accept": false
        }
      ]
    }
  }
```

wanlwanl:
Consider only need to change the settings in server, no need to change anything in client.
Server controls what client matches the tracking rules.

```
  "Azure": {
    "SignalR": {
      ...
      "TrackingRules": [
        IP: ["x.x.x.x", "y.y.y.y"]
      ]
    }
  }
```

wanlwanl:
Need to change client code and app server options.
Consider a website. Need to restart.

client query + server option:
client: `var conn = new HubConnectionBuilder().WithUrl($"{endpoint}?track=true")`
server: appsettings.json: `"acceptTracking":"true"`

wanlwanl:
Add claim claims in client. No need to change server.
Consider a website. Need to restart.

wanlwanl:
Finally, put all settings in server side, add a method like `ServerSticky`

PR #889, opened 2020-04-24
Service protocol: add message ID to broadcast data message

wanlwanl:
Add for broadcast first. Others later.

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/Utilities/ServiceMessageHelper.cs
+++ b/src/Microsoft.Azure.SignalR.Common/Utilities/ServiceMessageHelper.cs
@@ -1,0 +1,45 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Security.Cryptography;
+using Microsoft.Azure.SignalR.Protocol;
+
+namespace Microsoft.Azure.SignalR
+{
+    internal static class ServiceMessageHelper
+    {
+        private static readonly RNGCryptoServiceProvider _keyGenerator = new RNGCryptoServiceProvider();
+        private static readonly char[] _padding = { '=' };
+
+        // todo: generate a shorter ID
+        public static string GenerateMessageId()
+        {
+            // 128 bit buffer / 8 bits per byte = 16 bytes
+            Span<byte> buffer = stackalloc byte[16];
+            var bufferArray = buffer.ToArray();
+            _keyGenerator.GetBytes(bufferArray);
+            // Generate the id with RNGCrypto because we want a cryptographically random id, which GUID is not
+            return Base64UrlEncode(bufferArray);
+        }
+
+        public static string GetMessageId(ServiceMessage serviceMessage)
+        {
+            if (serviceMessage is MulticastDataMessage multicastDataMessage)
+            {
+                return multicastDataMessage.MessageId;
+            }
+
+            return null;
+        }
+
+        private static string Base64UrlEncode(byte[] input)
+        {
+            return Convert
+                .ToBase64String(input)
+                .TrimEnd(_padding)
+                .Replace('+', '-')
+                .Replace('/', '_');
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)


src/Microsoft.Azure.SignalR.Protocols/ServiceProtocolConstants.cs
```diff
@@ -15,6 +15,7 @@ public static class ServiceProtocolConstants
         public const int UserDataMessageType = 8;
         public const int MultiUserDataMessageType = 9;
         public const int BroadcastDataMessageType = 10;
+        public const int BroadcastDataMessageWithMessageIdType = 100010; // 1xxxxx mean the message with ID, otherwise without ID
```
vwxyzh, line 18: Should we define different types for with message id? Or just add an optional field for message id?

src/Microsoft.Azure.SignalR.Common/Utilities/ServiceMessageHelper.cs
```diff
@@ -0,0 +1,45 @@
...
+    {
+        private static readonly RNGCryptoServiceProvider _keyGenerator = new RNGCryptoServiceProvider();
+        private static readonly char[] _padding = { '=' };
+
+        // todo: generate a shorter ID
+        public static string GenerateMessageId()
+        {
+            // 128 bit buffer / 8 bits per byte = 16 bytes
+            Span<byte> buffer = stackalloc byte[16];
+            var bufferArray = buffer.ToArray();
```
vwxyzh, line 20: why not `new byte[16]`?

src/Microsoft.Azure.SignalR.Common/Utilities/ServiceMessageHelper.cs
```diff
@@ -0,0 +1,45 @@
...
+            _keyGenerator.GetBytes(bufferArray);
```
vwxyzh, line 21: should we use a prefix + counter as a message id?
e.g.: `prefix + Interlocked.Increase(ref count).ToString("x")`

src/Microsoft.Azure.SignalR.Protocols/ServiceProtocolConstants.cs
wanlwanl: I just add a optional field for message ID in `MulticastDataMessage`, so that we only need to add a new message type in `ServiceProtocolConstants`. 

If add a new type for adding a field for message ID, then all messages that need the new message class...

Or we change the message type, since diffent message type share some common fileds, we can make every digit meaningful, one digit to one field, so we know how to write and parse the message, then it might be easier to extend message in the future.

src/Microsoft.Azure.SignalR.Common/Utilities/ServiceMessageHelper.cs
wanlwanl: Sounds good! Then we have the message order .

wanlwanl: Then prefix is a 1:1 ID to server connection, but shorter than server connection ID.
The message ID will be: `Prefix.MessageIndexHex`

src/Microsoft.Azure.SignalR.Protocols/ServiceProtocol.cs
```diff
@@ -313,7 +315,15 @@ private static void WriteMultiUserDataMessage(ref MessagePackWriter writer, Mult
...
-            writer.Write(ServiceProtocolConstants.BroadcastDataMessageType);
+            if (message.MessageId == null)
+            {
+                writer.Write(ServiceProtocolConstants.BroadcastDataMessageType);
+            }
+            else
+            {
+                writer.Write(ServiceProtocolConstants.BroadcastDataMessageWithMessageIdType);
+                writer.Write(message.MessageId);
+            }
```
vwxyzh, line 324: invalid array header, it should be 4.
we should create a standard for var-length messages.
my prefer: [type, wellknown properties, ["ext-prop1", "value-1", "ext-prop2", "value-2"]]

wanlwanl:
zhenghui make it in #892, so close this pr.

PR #890, opened 2020-04-24
Add check-user-existence-in-group API for Transient mode

wanlwanl:


vicancy:
There are several updates to swagger in next service release, as swagger updated in this PR https://github.com/Azure/azure-signalr/pull/858 

Is this PR the first step and other changes are in the way?

wanlwanl:
I found the the new REST API are defined in our docs... So I add them in Management SDK. 
@vicancy you think update in Management SDK after the swagger being released? But they are in different SDK, I think it's fine to update in the same release.

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
```diff
@@ -196,6 +197,18 @@ public Task UserRemoveFromAllGroupsAsync(string userId, CancellationToken cancel
             return _restClient.SendAsync(api, HttpMethod.Delete, _productInfo, cancellationToken: cancellationToken);
         }
 
+        public async Task<bool> IsUserInGroup(string userId, string groupName, CancellationToken cancellationToken = default)
+        {
+            var isUserInGroup = false;
+            var api = _restApiProvider.GetUserGroupManagementEndpoint(_appName, _hubName, userId, groupName);
+            await _restClient.SendAsync(api, HttpMethod.Get, _productInfo, handleResponse: (request, response) =>
+                {
+                    isUserInGroup = response.StatusCode == HttpStatusCode.OK;
```
vwxyzh, line 206: what about 503?

wanlwanl: when service is unavailable?

wanlwanl: Let's throw when the status code isn't defined in the [doc.](https://github.com/Azure/azure-signalr/blob/dev/docs/rest-api.md#Check-user-existence-in-a-group)

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
```diff
@@ -196,6 +198,28 @@ public Task UserRemoveFromAllGroupsAsync(string userId, CancellationToken cancel
...
+                    switch (response.StatusCode)
+                    {
+                        case HttpStatusCode.OK:
+                            isUserInGroup = true;
+                            break;
+                        case HttpStatusCode.NotFound:
+                            break;
+                        default:
+                            var innerException = _restClient.GenerateInnerExceptionOnResponseFailure(response.StatusCode, response.ReasonPhrase);
+                            throw new AzureSignalRRuntimeException(response.RequestMessage.RequestUri.ToString(), innerException);
```
vwxyzh, line 216: what about 401 unauthorized, 400 bad request?

wanlwanl: Will throw in next commit....

wanlwanl: Fixed.

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
```diff
@@ -196,6 +198,26 @@ public Task UserRemoveFromAllGroupsAsync(string userId, CancellationToken cancel
...
+                {
+                    switch (response.StatusCode)
+                    {
+                        case HttpStatusCode.OK:
+                            isUserInGroup = true;
+                            return Task.CompletedTask;
+                        case HttpStatusCode.NotFound:
+                            return Task.CompletedTask;
+                        default:
+                            return _restClient.ThrowExceptionOnResponseFailureAsync(request, response);
```
vwxyzh, line 215: a bit confused with invoke `_restClient.ThrowExceptionOnResponseFailureAsync` in the callback function.
what about following:
Callback returns `true` for handled, `false` for unhandled.
If callback return false (i.e. unhandled), then `SendAsync` call `ThrowExceptionOnResponseFailureAsync`.

wanlwanl: How about changing the `handleResponse` to `Func<HttpResponseMessage, Func<HttpResponseMessage>, Task>`? (We don't need `HttpRequestMessage`, since `HttpResponseMessage` already contains an property for it).
```
public async Task<bool> IsUserInGroup(string userId, string groupName, CancellationToken cancellationToken = default)
        {
            var isUserInGroup = false;
            var api = _restApiProvider.GetUserGroupManagementEndpoint(_appName, _hubName, userId, groupName);
            await _restClient.SendAsync(api, HttpMethod.Get, _productInfo, handleResponse: ( response, ensureSuccessStatusCode) =>
                {
                    switch (response.StatusCode)
                    {
                        case HttpStatusCode.OK:
                            isUserInGroup = true;
                            return Task.CompletedTask;
                        case HttpStatusCode.NotFound:
                            return Task.CompletedTask;
                        default:
                            return ensureSuccessStatusCode(response); // change ThrowExceptionOnResponseFailureAsync to ensureSuccessStatusCode, since HttpResponseMessage already use this name, resue it here
                    }
                }, cancellationToken: cancellationToken);
            return isUserInGroup;
        }
```

wanlwanl: If `handleResponse` returns true/false,
then if `handleResponse` returns true, we need to write the throw exception code in the `handleResponse`, since we won't call `ThrowExceptionOnResponseFailureAsync`.
or if `handleResponse` returns false, we still can't skip some failure status code (e.g. 404 in this case) in `ThrowExceptionOnResponseFailureAsync`.

vwxyzh: I means:
```cs
public async Task<bool> IsUserInGroup(string userId, string groupName, CancellationToken cancellationToken = default)
{
    var isUserInGroup = false;
    var api = _restApiProvider.GetUserGroupManagementEndpoint(_appName, _hubName, userId, groupName);
    await _restClient.SendAsync(api, HttpMethod.Get, _productInfo, handleResponse: response =>
    {
        switch (response.StatusCode)
        {
            case HttpStatusCode.OK:
                isUserInGroup = true;
                return Task.FromResult(true);
            case HttpStatusCode.NotFound:
                return Task.FromResult(true);
            default:
                return Task.FromResult(false);
        }
    }, cancellationToken: cancellationToken);
    return isUserInGroup;
}
```
An exception should be always throw, when your handler returns `false`.
If you do not pass the handler, the default handler should check status code is 2XX.

vwxyzh: And it is better to have overloads for handleResponse, one for `Task<bool>`, another for `bool`

src/Microsoft.Azure.SignalR.Management/ServiceManager.cs
```diff
@@ -158,13 +158,19 @@ public async Task<bool> IsServiceHealthy(CancellationToken cancellationToken)
...
+            await _restClient.SendAsync(api, HttpMethod.Get, _productInfo, handleExpectedResponse: response =>
                 {
-                    if (response.StatusCode == HttpStatusCode.OK)
+                    switch (response.StatusCode)
                     {
-                        isHealthy = true;
+                        case HttpStatusCode.OK:
+                            isHealthy = true;
+                            return true;
+                        case HttpStatusCode.ServiceUnavailable:
```
vwxyzh, line 168: maybe 502

wanlwanl: I found 503 in [rest api doc](https://github.com/Azure/azure-signalr/blob/dev/docs/rest-api.md#Service-Health). Do you mean signalr service works as a proxy, so 502 is better?

wanlwanl: will update in another PR

PR #892, opened 2020-04-26
add extensible members in protocol.

vwxyzh:
Allow adding more optional members in future.

## Issue

Serialize and deserialize optional members.

## Scenario

We will have more optional members in near future.

* Tracing Id -  for diagnosis
* Sequence Id - for ordering in some special time (client connection switch to another server connection in the same server).

We cannot use current length based deserializing to detect which field occurred.

## Solution
Add a map field represent all optional fields in the binary.

* All of optional fields should write in this field.
* Non-optional fields are not recommended in this field.
* Key is field hint for identifying the optional field (e.g.: tracing id is 1).
* Value is the real value for the optional field.

### Sample
The original format for `ConnectionDataMessage` is:
```
[(type)6, (ConnectionId) "abc", (payload) binary]
```
Updated format is:
```
[(type)6, (ConnectionId) "abc", (payload) binary, (optional fields) { }]
```
And with tracing id `id123`:
```
[(type)6, (ConnectionId) "abc", (payload) binary, (optional fields) { (hint)1: (value)"id123" }]
```

## About this PR

This PR just want to standardize the optional fields serializing/deserializing.
Define optional fields or implement them is out of scope.

### DIFF
--- a/src/Microsoft.Azure.SignalR.Protocols/IMessageWithTracingId.cs
+++ b/src/Microsoft.Azure.SignalR.Protocols/IMessageWithTracingId.cs
@@ -1,0 +1,7 @@
+﻿namespace Microsoft.Azure.SignalR.Protocol
+{
+    public interface IMessageWithTracingId
+    {
+        string TracingId { get; set; }
+    }
+}
--- a/src/Microsoft.Azure.SignalR.Protocols/ServiceMessage.cs
+++ b/src/Microsoft.Azure.SignalR.Protocols/ServiceMessage.cs
@@ -3,6 +3,8 @@
 
 using System;
 
+using MessagePack;
+
 namespace Microsoft.Azure.SignalR.Protocol
 {
     /// <summary>
@@ -10,6 +12,49 @@ namespace Microsoft.Azure.SignalR.Protocol
     /// </summary>
     public abstract class ServiceMessage
     {
+    }
+
+    /// <summary>
+    /// Base class of messages between Azure SignalR Service and SDK.
+    /// </summary>
+    public abstract class ExtensibleServiceMessage : ServiceMessage
+    {
+        private const int TracingId = 1;
+
+        internal void WriteExtensionMembers(ref MessagePackWriter writer)
+        {
+            int count = 0;
+            var tracingId = (this as IMessageWithTracingId)?.TracingId;
+            if (!string.IsNullOrEmpty(tracingId))
+            {
+                count++;
+            }
+            writer.WriteArrayHeader(count * 2);
+            if (!string.IsNullOrEmpty(tracingId))
+            {
+                writer.Write(TracingId);
+                writer.Write(tracingId);
+            }
+        }
+
+        internal void ReadExtensionMembers(ref MessagePackReader reader)
+        {
+            int count = reader.ReadArrayHeader() / 2;
+            for (int i = 0; i < count; i++)
+            {
+                switch (reader.ReadInt32())
+                {
+                    case TracingId:
+                        if (this is IMessageWithTracingId withTracingId)
+                        {
+                            withTracingId.TracingId = reader.ReadString();
+                        }
+                        break;
+                    default:
+                        break;
+                }
+            }
+        }
     }
 
     /// <summary>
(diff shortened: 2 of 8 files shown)


vicancy:
There is already one in the way https://github.com/Azure/azure-signalr/pull/820

I also have a requirement to add some additional `Metadata` field

src/Microsoft.Azure.SignalR.Protocols/ServiceMessage.cs
```diff
@@ -12,6 +14,49 @@ public abstract class ServiceMessage
...
     }
 
+    /// <summary>
+    /// Base class of messages between Azure SignalR Service and SDK.
+    /// </summary>
+    public abstract class ExtensibleServiceMessage : ServiceMessage
+    {
+        private const int TracingId = 1;
+
+        internal void WriteExtensionMembers(ref MessagePackWriter writer)
```
wanlwanl, line 24: This method isn't used

src/Microsoft.Azure.SignalR.Protocols/IMessageWithTracingId.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Microsoft.Azure.SignalR.Protocol
+{
+    public interface IMessageWithTracingId
+    {
+        string TracingId { get; set; }
```
wanlwanl, line 5: How about call it `Id` instead?

src/Microsoft.Azure.SignalR.Protocols/ServiceMessage.cs
```diff
@@ -12,6 +14,49 @@ public abstract class ServiceMessage
...
+            }
+            writer.WriteArrayHeader(count * 2);
+            if (!string.IsNullOrEmpty(tracingId))
+            {
+                writer.Write(TracingId);
+                writer.Write(tracingId);
+            }
+        }
+
+        internal void ReadExtensionMembers(ref MessagePackReader reader)
```
wanlwanl, line 40: This method isn't used

vwxyzh: pls extend diff in serviceprotocol.cs

src/Microsoft.Azure.SignalR.Protocols/IMessageWithTracingId.cs
vwxyzh: pls extend diff in serviceprotocol.cs

vwxyzh: just a sample, you can change it.

src/Microsoft.Azure.SignalR.Protocols/ServiceMessage.cs
```diff
@@ -12,6 +14,49 @@ public abstract class ServiceMessage
     {
     }
 
+    /// <summary>
+    /// Base class of messages between Azure SignalR Service and SDK.
+    /// </summary>
+    public abstract class ExtensibleServiceMessage : ServiceMessage
```
vicancy, line 20: what if I only need extensible to `ConnectionDataMessage`? just have `ExtensibleServiceMessage` to `ServiceMessage` as TraceId sounds applicable for all the messages?

vwxyzh: yes, leave it to @wanlwanl 😃

davidfowl:
What's this for exactly? Why isn't it extensible today? Is there a spec on this design?

vwxyzh:
@davidfowl I add more detials for this pr.

davidfowl:
1. Do all new fields go into the extension field?
2. How does it affect the size of the message?

vwxyzh:
@davidfowl 
1. In my proposal, only optional fields should go into the extension field.
    If the field is required, I do not suggest, although I don't think there will be a new required field.
2. Message size: all kinds of messages allowing optional fields will append one more byte (an empty map) at least.
   And for messages have optional field, one more integer value (bytes based on the value of integer, -31 to 127 takes one byte) for hint, and the bytes for value itself.

PR #893, opened 2020-04-26
Put messaging logs into "Microsoft.Azure.SignalR.Messaging" category

wanlwanl:


src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -112,7 +114,28 @@ protected set
...
+            var currentNamespace = typeof(ServiceConnectionBase).Namespace;
+            Logger = loggerFactory.CreateLogger($"{currentNamespace}.ServiceConnection"); // keep original category name
+            /*
+             * Put all messaging log to Info level isn't appropriate, to Debug level, there are too many ping logs. So plan to put into a new category'
+             * Let CX filter messaging log by appsettings.json's logging filter: https://github.com/Azure/azure-signalr/blob/dev/docs/howto-tsg.md#add_logs_server_aspnetcore
+             * {
+                    "Logging": {
+                        "LogLevel": {
+                            ...
+                            "Microsoft.Azure.SignalR.Messaging": "Debug",
```
vicancy, line 131: if customer writes `"Microsoft.Azure.SignalR": "Debug"`, it will also include all the message logs. Is that expected?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -112,7 +114,28 @@ protected set
                 _handshakeRequest = new HandshakeRequestMessage(serviceProtocol.Version, (int)connectionType, (int)migrationLevel);
             }
 
-            Logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            if (loggerFactory == null)
+            {
+                throw new ArgumentNullException(nameof(loggerFactory));
+            }
+
+            var currentNamespace = typeof(ServiceConnectionBase).Namespace;
```
vicancy, line 122: `const` is enough?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
vicancy, line 131: what is the experience for ASP.NET SignalR which also leverages this class?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
vwxyzh: nameof

davidfowl:
Can I ask why? There’s no explanation as to why we’re doing this change

wanlwanl:
@davidfowl I plan to add message logs in the future, and provide a way to filter message logs for customers.

davidfowl:
What are message logs?

wanlwanl:
Message logs are logging down when message go into SignalR service and go out from SignalR service, and when there're warning or error on message delivery. The message log will be put in [Azure diagnostic log](https://docs.microsoft.com/en-us/azure/cdn/cdn-azure-diagnostic-logs) and also log in server side. I've send you the spec of message log, please take a look if you're interested.

PR #895, opened 2020-04-28
Add TrackingClientFilter in ServiceOption

wanlwanl:
There is a scenario for [_message log feature_](https://microsoft.sharepoint.com/teams/IoTToolingTeam/_layouts/OneNote.aspx?id=%2Fteams%2FIoTToolingTeam%2FSiteAssets%2FAzure%20SignalR%20Service%20Notebook&wd=target%28SignalR%20Service%20-%20SignalR%20Service%20-Engineering%20Doc.one%7CCE337738-D146-4D4C-B03E-9FFF6F433FCD%2F-%20%5BWIP%5D%20Improve%20Customers%27%20Experience%3A%20Message%20Log%7C38C10298-91B4-4D6C-9B34-93B4B29C8E4C%2F%29) that customers only want to add message logs for some specific clients. i.e. _tracking clients_. To support this, app server sould need to support filter client connections to tell which clients are _tracking clients_.

This PR add a new option in `ServiceOption` to let customers determine which are _tracking client_ based on `HttpContext`. Server SDK add a internal claim internally and pass to SignalR service, then SignalR service will determine whether the clients are _tracking clients_ finally.

### DIFF
--- a/src/Microsoft.Azure.SignalR.AspNet/Middleware/NegotiateMiddleware.cs
+++ b/src/Microsoft.Azure.SignalR.AspNet/Middleware/NegotiateMiddleware.cs
@@ -31,7 +31,7 @@ namespace Microsoft.Azure.SignalR.AspNet
 
         private readonly string _appName;
         private readonly Func<IOwinContext, IEnumerable<Claim>> _claimsProvider;
-        private readonly Func<IOwinContext, bool> _trackingClientFilter;
+        private readonly Func<IOwinContext, bool> _tracingClientFilter;
         private readonly ILogger _logger;
 
         private readonly IServiceEndpointManager _endpointManager;
@@ -53,7 +53,7 @@ namespace Microsoft.Azure.SignalR.AspNet
             _provider = configuration.Resolver.Resolve<IUserIdProvider>();
             _appName = appName ?? throw new ArgumentNullException(nameof(appName));
             _claimsProvider = options?.ClaimsProvider;
-            _trackingClientFilter = options?.TrackingClientFilter;
+            _tracingClientFilter = options?.TracingClientFilter;
             _endpointManager = endpointManager ?? throw new ArgumentNullException(nameof(endpointManager));
             _router = router ?? throw new ArgumentNullException(nameof(router));
             _connectionRequestIdProvider = connectionRequestIdProvider ?? throw new ArgumentNullException(nameof(connectionRequestIdProvider));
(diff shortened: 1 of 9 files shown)


src/Microsoft.Azure.SignalR.AspNet/Middleware/NegotiateMiddleware.cs
```diff
@@ -229,6 +231,11 @@ private Func<IEnumerable<Claim>> GetClaimsProvider(IOwinContext context)
             return () => _claimsProvider.Invoke(context);
         }
 
+        private bool IsTrackingClient(IOwinContext context)
+        {
+            return _trackingClientFilter != null && _trackingClientFilter(context);
```
vicancy, line 236: `?.Invoke`

src/Microsoft.Azure.SignalR.AspNet/ServiceOptions.cs
```diff
@@ -38,6 +38,12 @@ public class ServiceOptions : IServiceEndpointOptions
         /// </summary>
         public Func<IOwinContext, IEnumerable<Claim>> ClaimsProvider { get; set; } = null;
 
+        /// <summary>
+        /// Gets or sets the func to set tracking client filter from <see cref="IOwinContext" />.
+        /// The clients will be regarded as tracking client only if the function returns true.
+        /// </summary>
+        public Func<IOwinContext, bool> TrackingClientFilter { get; set; } = null;
```
vicancy, line 45: naming: `ConfigureDiagnosticClients`?

src/Microsoft.Azure.SignalR.AspNet/Middleware/NegotiateMiddleware.cs
wanlwanl: But it's a `bool?`

src/Microsoft.Azure.SignalR.AspNet/ServiceOptions.cs
wanlwanl: Looks like `Filter` give a condition to the input, something like https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter,
`Configure` pass in an `Action`, somthing like https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-3.1#add-providers.
So I prefer to use `xxxFilter`. does it make sense?

src/Microsoft.Azure.SignalR.AspNet/Middleware/NegotiateMiddleware.cs
vicancy: `?.Invoke ?? false`

src/Microsoft.Azure.SignalR.AspNet/ServiceOptions.cs
vicancy: `TrackingClient` this is my main concern, it sounds too big a term while actually it is only about logging.

wanlwanl: So `DiagnosticClientFilter`?

wanlwanl: DiagnosticClientEnableCondition? Too long?

JialinXin: DiagnosticClientTracker ?

davidfowl: ClientLoggingFilter?

Issue #846, opened 2020-03-15
Make DefaultKeepAliveInterval Configurable


emilroman:
Hi,

We're currently building a web app making use of SignalR. Our clients are devices that don't always have a reliable internet connection (ex: wearables) and resort to long polling as the preferred transport type. Because of this and given the resource limitations of the devices, the time interval between 2 long polling requests can be > 5 seconds. We would really like to use Azure SignalR Service, because of its scaling benefits, but to do so, we need to be able to configure the DefaultKeepAliveInterval.

Would it be possible to make this configurable?

vicancy:
For AspNetCore SignalR it should be natively supported when using Azure SignalR Service. Are you using the old AspNet SignalR?

emilroman:
Thank you very much for your reply @vicancy, we are using AspNetCore SignalR.
On our client devices, we are not able to leverage the existing SignalR client libraries and will build our own according to the specs (https://github.com/aspnet/SignalR/tree/master/specs).

The main issue we faced so far is that SignalR's "HttpConnectionManager.Scan()" method, removes inactive connections after 15 seconds, by default. In our use case, we might sometimes have > 15 seconds between 2 long polling requests, because of the nature of the devices.
After looking at the SignalR source code, we've found a way to get around this, by setting the value of the public "ConnectionOptionsSetup.DefaultDisconectTimeout" property to something else while mapping the hub endpoint:

"endpoints.MapHub<TestHub>("/test", opts =>
{
	ConnectionOptionsSetup.DefaultDisconectTimeout = TimeSpan.FromMinutes(2);
});"

However, this (arguably hack-ish) approach doesn't work when using Azure SignalR Service, which by default, seems to disconnect inactive connections after 5 seconds. After looking at the source code, I wonder if this is due to "ServiceConnectionBase.DefaultKeepAliveInterval" having a hardcoded value of 5 seconds?

vicancy:
Thanks for the detailed explanation, yes you are right, we will look into it.

emilroman:
Thank you very much for your help @vicancy . Looking forward to see if it will be possible to accommodate our scenario.
Just wanted to add that our SignalR clients will not be pinging the server (in order to preserve battery life), which should theoretically update the last seen timestamp, but even if they did, we would still face the same potential network lag issues between 2 consecutive requests (be it long polling or server pings).

BrennanConroy:
[...]

PR #896, opened 2020-04-29
Allow customize DisconnectTimeout for ASRS timeout idle connections

JialinXin:
support #846


src/Microsoft.Azure.SignalR.AspNet/Middleware/NegotiateMiddleware.cs
```diff
@@ -209,7 +209,7 @@ private IEnumerable<Claim> BuildClaims(IOwinContext owinContext, IRequest reques
             var user = owinContext.Authentication?.User;
             var userId = _provider?.GetUserId(request);
 
-            var claims = ClaimsUtility.BuildJwtClaims(user, userId, GetClaimsProvider(owinContext), _serverName, _mode, _enableDetailedErrors, _endpointsCount);
+            var claims = ClaimsUtility.BuildJwtClaims(user, userId, GetClaimsProvider(owinContext), _serverName, _mode, _enableDetailedErrors, _endpointsCount, _disconnectTimeout);
```
vwxyzh, line 212: should we use builder pattern?
I find wangpeng also add an parameter in this build.

src/Microsoft.Azure.SignalR/ServiceOptionsExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace Microsoft.Azure.SignalR
+{
+    internal static class ServiceOptionsExtensions
+    {
+        public static void Validate(this ServiceOptions options)
+        {
+            if (options.DisconnectTimeoutInSeconds.HasValue &&
+                (options.DisconnectTimeoutInSeconds < 1 || options.DisconnectTimeoutInSeconds > 300))
+            {
+                throw new ArgumentOutOfRangeException("DisconnectTimeoutInSeconds", "Value should be [1,300].");
```
vwxyzh, line 15: * ArgumentException, the argument is 'options'
* Can we check it when set the property `DisconnectTimeoutInSeconds`?

JialinXin: Not able to move this into ServiceOptionsSetup cause the value is not coming from file configuration but code option settings.

AspNetCore support [`IOptionsValidate`](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-3.1) after 2.2 which is not compatible with 2.1, so choose to use `PostConfigure` to do the validation.

JialinXin: I think `ArgumentOutOfRangeException` should be more appropriate for our scenario. https://docs.microsoft.com/en-us/dotnet/api/system.argumentoutofrangeexception?view=netcore-3.1#remarks

src/Microsoft.Azure.SignalR.AspNet/Middleware/NegotiateMiddleware.cs
JialinXin: Good idea. Will use another PR to do the refactor.

PR #902, opened 2020-05-09
add ttl for adding user to group

vwxyzh:


src/Microsoft.Azure.SignalR.Management/IUserGroupManager.cs
```diff
@@ -19,7 +20,17 @@ public interface IUserGroupManager
         /// <param name="cancellationToken">The token to monitor for cancellation requests. The default value is System.Threading.CancellationToken.None.</param>
         /// <returns>A System.Threading.Tasks.Task that represents the asynchronous add.</returns>
         Task AddToGroupAsync(string userId, string groupName, CancellationToken cancellationToken = default);
```
vicancy, line 22: question: when this one is called, user lives forever in the group until it is explicitly removed from the group, right?

wanlwanl: Yes, cx can use [another api](https://github.com/Azure/azure-signalr/blob/dev/docs/rest-api.md#Remove-a-user-from-all-groups) to remove the user from all groups.

PR #911, opened 2020-05-22
mark diagnostic client on client connected

wanlwanl:


src/Microsoft.Azure.SignalR.Common/ServiceConnections/ClientConnectionScope.cs
```diff
@@ -27,10 +27,15 @@ protected internal ClientConnectionScope(IServiceConnection outboundConnection)
...
+                        Properties = new ClientConnectionScopeProperties()
+                        {
+                            OutboundServiceConnection = outboundConnection,
+                            IsDiagnosticClient = IsDiagnosticClient
+                        }
                     };
             }
-            else if (outboundConnection != null)
+            else if (outboundConnection != null || 
+                ScopePropertiesAccessor<ClientConnectionScopeProperties>.Current.Properties.IsDiagnosticClient != IsDiagnosticClient)
```
KKhurin, line 38: `Current.Properties` could be null..

Since this is a sanity check I am just thinking this would be enough:
~~~
 else if (outboundConnection != default || IsDiagnosticClient != default)
~~~

wanlwanl: Can we add a member `bool _established` to avoid the replacement `Debug.Assert(false, "Attempt to replace an already established scope");`?

KKhurin: This would need to be a static property like this:
~~~
static bool ScopeIsEstablished => ScopePropertiesAccessor<ClientConnectionScopeProperties>.Current != null;
~~~

Then the cctors will look like this:
~~~
 internal ClientConnectionScope() : this(default, default)
 {
    if(ScopeIsEstablished)
    {
      // Debug.Assert - not comfortable throwing so I'd leave it noop in retail builds
    }
 }

protected internal ClientConnectionScope(IServiceConnection outboundConnection, bool IsDiagnosticClient)
{
    // Only allow to carry one copy of connection properties regardless of how many nested scopes are created
    if (!ScopeIsEstablished)
    {    
        ScopePropertiesAccessor<ClientConnectionScopeProperties>.Current =
                    new ScopePropertiesAccessor<ClientConnectionScopeProperties>()
                    {
                        Properties = new ClientConnectionScopeProperties()
                        {
                            OutboundServiceConnection = outboundConnection,
                            IsDiagnosticClient = IsDiagnosticClient
                        }
                    };
    }
}
~~~

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ClientConnectionScope.cs
```diff
@@ -61,12 +65,25 @@ internal static IServiceConnection OutboundServiceConnection
...
+            {
+                var currentProps = ScopePropertiesAccessor<ClientConnectionScopeProperties>.Current?.Properties;
+                if (currentProps != null)
+                {
+                    currentProps.IsDiagnosticClient = value;
+                }
+            }
+        }
 
+        // todo: extend with client connection tracking/logging accessors
```
KKhurin, line 81: nit: remove the todo (since you've just added get/set for IsDiagnosticClient

PR #913, opened 2020-05-25
generate tracing id

wanlwanl:
tracing ID (data type: long) has two part: 
{p1}.{p2}.{p3}
* p1: indicates whether the message comes from tracking client
* p2: server related: different if server restarts, also different among servers.
* p3: message related: simple counter, minimal performance impact.


src/Microsoft.Azure.SignalR.Common/IMessageWithTracingIdExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace Microsoft.Azure.SignalR
+{
+    internal static class IMessageWithTracingIdExtensions
+    {
+        private static int _prefix = Guid.NewGuid().GetHashCode();
+        private static int counter = 0;
+
+        public static void UpdateTracingId(this IMessageWithTracingId msg)
+        {
+            msg.TracingId = $"{(uint)_prefix}:{(uint)counter}";
```
vicancy, line 17: prefix to reuse serverName? maybe a Id generator

Also I think in netcore it choose to use the naming convention `Logging` instead `Tracing`

TracingId is string, shall we use a fixed length for counter?

wanlwanl: Better to use something unrelated to customer's custom name, to avoid privacy issue in service.
Fixed length sounds good, also make it a HEX to shortern the length.

vicancy: Reuse this one https://github.com/Azure/azure-signalr/blob/dev/src/Microsoft.Azure.SignalR.Common/DefaultServerNameProvider.cs#L13

> to avoid privacy issue in service
Hm, if privacy is a concern, we need to update the default server name provider's logic instead. cc @JialinXin

JialinXin: Right, Guid itself is good, do not add server name. We'd update the provider's logic.

wanlwanl: updated

wanlwanl: updated

src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManagerBase.cs
```diff
@@ -45,7 +46,7 @@ public override Task SendAllAsync(string methodName, object[] args, Cancellation
             }
 
             return ServiceConnectionContainer.WriteAsync(
-                new BroadcastDataMessage(null, SerializeAllProtocols(methodName, args)));
+                new BroadcastDataMessage(null, SerializeAllProtocols(methodName, args), tracingId: MessageIdHelper.Generate(ClientConnectionScope.IsDiagnosticClient)));
```
vicancy, line 49: how about moving `ClientConnectionScope.IsDiagnosticClient` into the generator, 
```cs
new BroadcastDataMessage(null, SerializeAllProtocols(methodName, args).WithMessageId()
```

src/Microsoft.Azure.SignalR.Common/ServiceMessages/ExtensibleServiceMessageExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+using Microsoft.Azure.SignalR.Protocol;
+
+namespace Microsoft.Azure.SignalR
+{
+    internal static class ExtensibleServiceMessageExtensions
+    {
+        public static ExtensibleServiceMessage WithMessageId(this ExtensibleServiceMessage message)
+        {
+            var id = MessageIdHelper.Generate(ClientConnectionScope.IsDiagnosticClient);
+            if (message is IMessageWithTracingId messageWithId)
```
vwxyzh, line 14: check the message first, then generate the id.

src/Microsoft.Azure.SignalR.Common/ServiceMessages/ExtensibleServiceMessageExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.Azure.SignalR.Common.ServiceConnections;
+using Microsoft.Azure.SignalR.Protocol;
+
+namespace Microsoft.Azure.SignalR
+{
+    internal static class ExtensibleServiceMessageExtensions
+    {
+        public static ExtensibleServiceMessage WithMessageId(this ExtensibleServiceMessage message)
```
vwxyzh, line 11: should we extension `IMessageWithTracingId`?

PR #921, opened 2020-06-03
[Asp.Net] Refactor of client connection manager

terencefan:


src/Microsoft.Azure.SignalR.AspNet/ServerConnections/ServiceConnection.cs
```diff
@@ -213,8 +213,8 @@ private async Task OnConnectedAsyncCore(ClientConnectionContext clientContext, O
             var connectionId = message.ConnectionId;
             try
             {
-                clientContext.Transport =
-                    await _clientConnectionManager.CreateConnection(message, this);
+                clientContext.Transport = await _clientConnectionManager.CreateConnection(message);
+                clientContext.ServiceConnection = this;
```
vicancy, line 217: move the set of `ServiceConnection` to `.ctor` so that it is guaranteed to be not null at the very beginning?

src/Microsoft.Azure.SignalR.AspNet/Transports/AzureTransport.cs
```diff
@@ -96,12 +99,13 @@ private async Task LifetimeExecute()
...
         {
-            if (_context.Environment.TryGetValue(AspNetConstants.Context.AzureServiceConnectionKey, out var connection) && connection is IServiceConnection serviceConnection)
-            {
-                // Invoke service connection
-                return serviceConnection.WriteAsync(ConnectionId, value, _serviceProtocol, _serializer, _pool);
+            if (_clientConnectionManager.TryGetClientConnection(ConnectionId, out var connection))
+            {
+                if (connection.ServiceConnection != null)
+                {
+                    return connection.ServiceConnection.WriteAsync(ConnectionId, value, _serviceProtocol, _serializer, _pool);
```
vicancy, line 106: >ServiceConnection.WriteAsync [](start = 38, length = 28)

move `ServiceConnection.WriteAsync` method into the ClientConnectionContext, eg.g. `WriteToTransportAsync`? So that we don't need to have this `ServicConnection` property public

PR #912, opened 2020-05-25
[Experiment] Get message log settings from service

wanlwanl:
Server SDK always generate message log in trace log level.

Receive message log settings via ping message.
Determine whether adding tracing ID.

When service receives a message with tracing ID, it generates message log.

PR #923, opened 2020-06-08
Service connection container scope for handling ping messsage with diagnostic log message

wanlwanl:
## Background

CX turn on "Collect all" for message log in portal.
Need to find a way to notify the server that server should enable message log.

## What the PR do

Service sends pings to server via server connection. 
Ping message contains a bool flag to tell server whether server needs to enable message log.
The flag may change due to customer changing portal settings.
In server, a service connection container receives the ping message and stores the flag in the `ServiceConnectionContainerScope`.
Finally, in `ServiceLifetimeManagerBase`, get the flag from `ServiceConnectionContainerScope` to decide whether it should generate tracing ID in `ServiceMessage`.
The server only write message log when the `ServiceMessage` contains a tracing ID.

## Where is the message log
ep == endpoint.
......................................epC
......................................^
.......................................|
Consider epA -> server -> epA
.......................................| 
.......................................v
......................................epB
And epA and epC enables message log while epB not.
Then message M comes from epA will has message log in server. epA and epC also have message log since they enable.
Due to epB disables message log, so no message log will be generated. If epB wants to have message log, it should enable it. The M stops having message log when the target service disable message log. If the cx disable message log in some ep, we don't want to impact it, unless cx enable manually enable it.

## TODO
Only support aspnetcore for now, aspnet in future PR.


src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -374,6 +398,11 @@ internal static TimeSpan GetRetryDelay(int retryCount)
             return TimeSpan.FromSeconds(1 << retryCount) + ReconnectInterval;
         }
 
+        private void UpdateEnableMessageLogEvent(bool enableMessageLog)
+        {
+            _enableMessageLogEventHandler?.Invoke(enableMessageLog);
```
vicancy, line 403: no need for actions, `Func` is enough?

src/Microsoft.Azure.SignalR.Common/Interfaces/AsyncLocalScopeBase.cs
```diff
@@ -0,0 +1,58 @@
...
+            }
+        }
+
+        protected static Context Current
+        {
+            get => _asyncLocal.Value;
+            set => _asyncLocal.Value = value;
+        }
+
+        protected class Context
```
vicancy, line 48: oh no, what are these nested classes...?

src/Microsoft.Azure.SignalR.Common/Microsoft.Azure.SignalR.Common.csproj
```diff
@@ -26,4 +26,7 @@
     <PackageReference Include="Microsoft.AspNetCore.Connections.Abstractions" Version="$(MicrosoftAspNetCoreConnectionsAbstractionsPackageVersion)" />
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="$(MicrosoftExtensionsLoggingAbstractionsPackageVersion)" />
   </ItemGroup>
+  <ItemGroup>
+    <Folder Include="ServiceConnections\scopes\" />
```
vicancy, line 30: what is this folder for?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -51,6 +51,8 @@ internal abstract class ServiceConnectionContainerBase : IServiceConnectionConta
         private volatile bool _hasClients;
         private volatile bool _terminated = false;
 
+        private event Action<bool> _enableMessageLogEventHandler;
```
vwxyzh, line 54: private event is same as private delegate

src/Microsoft.Azure.SignalR.Common/Interfaces/AsyncLocalScopeBase.cs
wanlwanl: remember in this [pr](https://github.com/Azure/azure-signalr/pull/912)? 
Context -> [ClientConnectionScopeProperties](https://github.com/Azure/azure-signalr/blob/a409e1a3f30f43e1bc35318c70b212a61e329b71/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ClientConnectionScope.cs#L95)
Value -> [DiagnosticLogsContext](https://github.com/Azure/azure-signalr/blob/a409e1a3f30f43e1bc35318c70b212a61e329b71/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ClientConnectionScope.cs#L101)
We need to add a reference to make that we are able to update `ServiceDiagnosticLogsContext` in event handler.

src/Microsoft.Azure.SignalR.Common/Microsoft.Azure.SignalR.Common.csproj
wanlwanl: updated. async local scopes are stored here.

src/Microsoft.Azure.SignalR.Common/Interfaces/AsyncLocalScopeBase.cs
wanlwanl: seperated

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
wanlwanl: left _enableMessageLogEventHandler, removed UpdateEnableMessageLogEvent

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -140,11 +142,27 @@ private set
             _serversPing = new CustomizedPingTimer(Logger, Constants.CustomizedPingTimer.Servers, WriteServersPingAsync, Constants.Periods.DefaultServersPingInterval, Constants.Periods.DefaultServersPingInterval);
         }
 
-        public Task StartAsync() => Task.WhenAll(FixedServiceConnections.Select(c => StartCoreAsync(c)));
+        // todo: maybe create a new container-level ServiceConnectionContainerScope here, to reduce duplicate events on client-level.
+        public async Task StartAsync()
+        {
+            var context = new ServiceDiagnosticLogsContext { IsEnableMessageLog = false };
+            using (new ServiceConnectionContainerScope(context))
```
KKhurin, line 149: I would much rather prefer explicit (if possible) wiring than creating async locals here and there. 
Although this context is not carrying references to any other objects so the overhead and the risk of leaking is likely small.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -185,6 +203,10 @@ public virtual Task HandlePingAsync(PingMessage pingMessage)
                     _serversTagContext = Tuple.Create(serversTag, updatedTime);
                 }
             }
+            else if (RuntimeServicePingMessage.TryGetMessageLogEnableFlag(pingMessage, out var enableMessageLog))
+            {
+                _enableMessageLogEventHandler?.Invoke(enableMessageLog);
```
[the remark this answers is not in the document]
wanlwanl: Since the service message is generated in `ServiceLifetimeManager`, we need the information there too.

davidfowl:
The PR needs a description...

wanlwanl:
> The PR needs a description...

Added. @davidfowl

src/Microsoft.Azure.SignalR.Common/Constants.cs
```diff
@@ -77,6 +77,7 @@ public static class CustomizedPingTimer
         {
             public const string ServiceStatus = "ServiceStatus";
             public const string Servers = "Servers";
+            public const string DiagnosticLogs = "DiagnosticLogs";
```
KKhurin, line 80: nit.. are these logs for diagnostics only? if not maybe change the name to MessageTracingControl or EnableMessageTracing/Logging?

src/Microsoft.Azure.SignalR.Common/Interfaces/IServiceDiagnosticLogsContext.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.SignalR
+{
+    internal interface IServiceDiagnosticLogsContext
+    {
+        bool IsEnableMessageLog { get; set; }
```
KKhurin, line 8: nit.. naming.. maybe IsMessageLogEnabled or simply EnableMessageLog?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceDiagnosticLogsContext.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.SignalR
+{
+    internal class ServiceDiagnosticLogsContext : IServiceDiagnosticLogsContext
+    {
+        public bool IsEnableMessageLog { get; set; } = false;
```
KKhurin, line 8: nit naming..

src/Microsoft.Azure.SignalR.Common/ServiceConnections/scopes/ClientConnectionScope.cs
```diff
@@ -78,7 +78,7 @@ internal static bool IsDiagnosticClient
             }
         }
 
-        private class ClientConnectionScopeProperties
+        internal class ClientConnectionScopeProperties
```
KKhurin, line 81: why internal?

src/Microsoft.Azure.SignalR.Common/ServiceMessages/RuntimeServicePingMessages.cs
```diff
@@ -14,7 +16,12 @@ internal static class RuntimeServicePingMessage
         private const string StatusKey = "status";
         private const string ShutdownKey = "shutdown";
         private const string ServersKey = "servers";
-
+        private const string DiagnosticLogsKey = "diag";
+        private const string DiagnosticLogsMessagingTypeKey = "MessagingLogs";
+
+
+        private const string MessagingLogEnableValue = "1";
```
KKhurin, line 23: these are part of protocol, why can't they go to Microsoft.Azure.SignalR.Protocol?

src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManagerBase.cs
```diff
@@ -43,6 +43,10 @@ public override Task SendAllAsync(string methodName, object[] args, Cancellation
                 throw new ArgumentException(NullOrEmptyStringErrorMessage, nameof(methodName));
             }
 
+            //return ServiceConnectionContainer.WriteAsync(
+            //    new BroadcastDataMessage(null, SerializeAllProtocols(methodName, args), 
+            //    tracingId: ServiceConnectionContainerScope.ServiceDiagnosticLogsContext.IsEnableMessageLog ? "id": null));
```
KKhurin, line 48: ServiceDiagnosticLogsContext?.

src/Microsoft.Azure.SignalR.Common/Constants.cs
wanlwanl: I reuse the concept [Azure Diagnostic Log](https://docs.microsoft.com/en-us/azure/cdn/cdn-azure-diagnostic-logs). I want to keep the same. The message log is part of the azure diagnostic log of ASRS to be onboard. We also have a blade in portal for this log (in Diagnostic Settings blade).

src/Microsoft.Azure.SignalR.Common/ServiceMessages/RuntimeServicePingMessages.cs
```diff
@@ -14,7 +16,12 @@ internal static class RuntimeServicePingMessage
...
         private const string ShutdownKey = "shutdown";
         private const string ServersKey = "servers";
-
+        private const string DiagnosticLogsKey = "diag";
+        private const string DiagnosticLogsMessagingTypeKey = "MessagingLogs";
+
+
+        private const string MessagingLogEnableValue = "1";
+
         private const string StatusActiveValue = "1";
```
wanlwanl, line 25: > these are part of protocol, why can't they go to Microsoft.Azure.SignalR.Protocol?
Agreed. I found these are similar, so I just follow them. 
Our service already use these values, should be part of the protocol.
What do you think? @vicancy @vwxyzh

src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManagerBase.cs
wanlwanl: Look like if `ServiceConnectionContainerScope.ServiceDiagnosticLogsContext == null` there must be a bug, since the scope is created in [StartAsync()](https://github.com/Azure/azure-signalr/blob/7b627a95c68eb59fae4ed898f1d2ab0d90c8a282/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs#L148). So, still `ServiceDiagnosticLogsContext.xxx`?

src/Microsoft.Azure.SignalR.Common/ServiceMessages/RuntimeServicePingMessages.cs
vicancy: yes nice catch. We can address this issue in another PR.

src/Microsoft.Azure.SignalR.Common/ServiceMessages/RuntimeServicePingMessages.cs
wanlwanl: let's fix it in a seperate PR.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -46,7 +46,9 @@ internal abstract class ServiceConnectionContainerBase : IServiceConnectionConta
 
         private volatile ServiceConnectionStatus _status;
 
-        // <serversTag, latestTimestamp>
+        private volatile ServiceDiagnosticLogsContext _serviceDiagnosticLogsContext = new ServiceDiagnosticLogsContext { EnableMessageLog = false };
```
vicancy, line 49: >volatile [](start = 16, length = 8)

`readonly` is enough?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -140,7 +142,13 @@ private set
             _serversPing = new CustomizedPingTimer(Logger, Constants.CustomizedPingTimer.Servers, WriteServersPingAsync, Constants.Periods.DefaultServersPingInterval, Constants.Periods.DefaultServersPingInterval);
         }
 
-        public Task StartAsync() => Task.WhenAll(FixedServiceConnections.Select(c => StartCoreAsync(c)));
+        public async Task StartAsync()
+        {
+            using (new ServiceConnectionContainerScope(_serviceDiagnosticLogsContext))
```
vicancy, line 147: >ServiceConnectionContainerScope [](start = 23, length = 31)

How about those ondemand connections generated?

src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManagerBase.cs
```diff
@@ -43,6 +43,10 @@ public override Task SendAllAsync(string methodName, object[] args, Cancellation
                 throw new ArgumentException(NullOrEmptyStringErrorMessage, nameof(methodName));
             }
 
+            //return ServiceConnectionContainer.WriteAsync(
```
vicancy, line 46: >ServiceConnectionContainer [](start = 21, length = 26)

what if it is a mulitple endpoint container? That one endpoint enables message log while another not?

wanlwanl: ep == endpoint.
                             epC
                               ^
                               |
Consider epA -> server -> epA
                               | 
                               v
                             epB
And epA and epC enables message log while epB not.
Then message M comes from epA will has message log in server. epA and epC also have message log since they enable.
Due to epB disables message log, so no message log will be generated. If epB wants to have message log, it should enable it. The M stops having message log when the target service disable message log. If the cx disable message log in some ep, we don't want to impact it, unless cx enable manually enable it.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
wanlwanl: call chain: `StartCoreAsync` ->`OnConnectionComplete` -> `RestartServiceConnectionCoreAsync` -> `ReplaceFixedConnections`. 
The `ServiceConnectionContainerScope` also flows there even ondemand connection is replaced the current one. So it's ok to use `ServiceConnectionContainerScope` here.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/scopes/AsyncLocalScopeBase.cs
```diff
@@ -0,0 +1,47 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Diagnostics;
+using System.Threading;
+
+namespace Microsoft.Azure.SignalR
+{
+    internal abstract class AsyncLocalScopeBase<TProps> : IDisposable
```
vicancy, line 10: >AsyncLocalScopeBase [](start = 28, length = 19)

who else uses this base class?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/scopes/ClientConnectionScope.cs
```diff
@@ -30,7 +30,7 @@ protected internal ClientConnectionScope(IServiceConnection outboundConnection,
                                 Properties = new ClientConnectionScopeProperties()
                                 {
                                     OutboundServiceConnection = outboundConnection,
-                                    IsDiagnosticClient = isDiagnosticClient
+                                    IsDiagnosticClient = isDiagnosticClient,
```
vicancy, line 33: >IsDiagnosticClient [](start = 36, length = 18)

where will this property be used? Is this also a relevant property to this PR?

wanlwanl: reverted, accidently added a `,`

src/Microsoft.Azure.SignalR.Common/ServiceConnections/scopes/AsyncLocalScopeBase.cs
wanlwanl: We have `ClientConnectionScope`, very similar to `ServiceConnectioncontainerScope`, I want to make them inherit this base class later, but in another PR.

vicancy: consider reuse the same pattern of `ClientConnectionScope` to make the usage instead of adding more classes?

---
In reply to: [450011381](https://github.com/Azure/azure-signalr/pull/923#discussion_r450011381) [](ancestors = 450011381)

wanlwanl: > consider reuse the same pattern of `ClientConnectionScope` to make the usage simple instead of adding more classes?
> 

Try to protect the `IServiceDiagnosticLogsContext` inside asyncLocal. The `ScopePropertiesAccessor` is internal, can be accessed everywhere. I don't want the `IServiceDiagnosticLogsContext` accidently modified. The only right place to update the `IServiceDiagnosticLogsContext` is `ServiceConnectionContainerBase.HandlePingAsync()`, and no plan to modify else where. So I think it's better to limit the ability to modify `IServiceDiagnosticLogsContext`.
I updated the some [code](https://github.com/Azure/azure-signalr/pull/923/files#diff-d4a73abdc56033c8cc3e130bd06d079dR12) and [test](https://github.com/Azure/azure-signalr/pull/923/files#diff-bf96c03afebc543419254b1a6fd1baf6R1538).

src/Microsoft.Azure.SignalR.Common/ServiceConnections/scopes/ServiceConnectionContainerScope.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.SignalR
+{
+    internal class ServiceConnectionContainerScope : AsyncLocalScopeBase<IServiceDiagnosticLogsContext>
+    {
+        public ServiceConnectionContainerScope(IServiceDiagnosticLogsContext props) : base(props)
```
vicancy, line 8: >IServiceDiagnosticLogsContext [](start = 47, length = 29)

bool EnableMessageLog is enough?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/scopes/AsyncLocalScopeBase.cs
vicancy: I mean reuse `ScopePropertiesAccessor` in the `ServiceConnectionContainerScope`, any concerns in reusing that?

---
In reply to: [450045026](https://github.com/Azure/azure-signalr/pull/923#discussion_r450045026) [](ancestors = 450045026)

src/Microsoft.Azure.SignalR.Common/ServiceConnections/scopes/ServiceConnectionContainerScope.cs
wanlwanl: Yes, but we can't simply use a `bool`, otherwise we can't update the flag in `HandshakeAysnc`. We have to use a instance.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/scopes/AsyncLocalScopeBase.cs
wanlwanl: make sense! The `ClientConnectionScope` can limit the acess since the `ClientConnectionScopeProperties` a private class. But I can't since I need to add a member _serviceDiagnosticLogsContext [here](https://github.com/Azure/azure-signalr/pull/923/files#diff-72c9ae4e6a93ede091bc3b08bae82afaR46) to avoid using event and event handlers. So `ServiceDiagnosticLogsContext` should be internal, everywhere can use ServiceDiagnosticLogsContext, including `ScopePropertiesAccessor<IServiceDiagnosticLogsContext>` every where.

src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManagerBase.cs
KKhurin: This will be null if a customer (or the library they use) suppresses execution context flow. 

I am also concerned that the execution context can be swapped pretty easily.
For example user code can have a timer that periodically dispatches signalr messages to different endpoints (some have tracing enabled some don't). Whoever creates this timer will make every message sent from within this timer carry the tracing/logging/diagnostic flag for every endpoint that may have nothing to do with the original endpoint tracing settings. While the services that don't have tracing enabled will not trace anything, the SDK side will trace for all the messages originated from this timer (which can further confuse the users).

wanlwanl: > This will be null if a customer (or the library they use) suppresses execution context flow.

Right, make sense.

> I am also concerned that the execution context can be swapped pretty easily.
For example user code can have a timer that periodically dispatches signalr messages to different endpoints (some have tracing enabled some don't). Whoever creates this timer will make every message sent from within this timer carry the tracing/logging/diagnostic flag for every endpoint that may have nothing to do with the original endpoint tracing settings. While the services that don't have tracing enabled will not trace anything, the SDK side will trace for all the messages originated from this timer (which can further confuse the users).

I don't think the users will be confused. The original endpoint's tracing setting have control of the message inside server, but won't impact other endpoints. And the tracing setting will be updated if the service send a ping to notify. I don't see why it's relevant to timer... Maybe I misunderstand?

wanlwanl: look like your concern is similar to [this](https://github.com/Azure/azure-signalr/pull/923#discussion_r449967890).

src/Microsoft.Azure.SignalR.Common/ServiceConnections/scopes/AsyncLocalScopeBase.cs
vicancy: how about exposing a method in `ServiceConnectionContainerScope` to `EnableMessageLog` instead of having the context outside of the scope?

---
In reply to: [450049529](https://github.com/Azure/azure-signalr/pull/923#discussion_r450049529) [](ancestors = 450049529,450045026)

src/Microsoft.Azure.SignalR.Common/ServiceMessages/RuntimeServicePingMessages.cs
```diff
@@ -14,7 +16,12 @@ internal static class RuntimeServicePingMessage
         private const string StatusKey = "status";
         private const string ShutdownKey = "shutdown";
         private const string ServersKey = "servers";
-
+        private const string DiagnosticLogsKey = "diag";
+        private const string DiagnosticLogsMessagingTypeKey = "MessagingLogs";
```
vicancy, line 20: >MessagingLogs [](start = 63, length = 13)

shorten it, `m`?

src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManagerBase.cs
KKhurin: So if the original endpoint's tracing is on - is it expected that the app server will trace outcoming messages to other endpoints too?

wanlwanl: no, unless other endpoints also enable message log

KKhurin: Then there are many scenarios where a cross-leaked execution context (and the tracing flag it carries) can break this assumption. One such case is the timer example above, but there are really many-many others.

PR #925, opened 2020-06-09
Move RestClient to SignalR.Common

terencefan:


src/Microsoft.Azure.SignalR.Common/Utilities/RestApiProvider.cs
```diff
@@ -3,14 +3,14 @@
 
 using System;
 
-namespace Microsoft.Azure.SignalR.Management
+namespace Microsoft.Azure.SignalR
 {
     internal class RestApiProvider
```
vicancy, line 8: leave this to management project?

build/dependencies.props
```diff
@@ -6,8 +6,11 @@
     <!-- Azure ASP.NET Core SignalR -->
     <MicrosoftAspNetCoreHttpConnectionsClientPackageVersion>1.0.0</MicrosoftAspNetCoreHttpConnectionsClientPackageVersion>
     <MicrosoftAspNetCoreHttpConnectionsClientPackage3_0Version>3.0.0</MicrosoftAspNetCoreHttpConnectionsClientPackage3_0Version>
-    <MicrosoftExtensionsLoggingAbstractionsPackageVersion>1.0.0</MicrosoftExtensionsLoggingAbstractionsPackageVersion>
+    <MicrosoftExtensionsLoggingAbstractionsPackageVersion>2.1.0</MicrosoftExtensionsLoggingAbstractionsPackageVersion>
```
vicancy, line 9: don't update the version unless neccessory

wanlwanl:
Why moving it to common? Anywhere uses the rest client except for Management SDK?

terencefan:
@wanlwanl Our SDK may call `GetAccessKey` API to create a temporary AccessKey in the future.

build/dependencies.props
terencefan: `Extension.Http=2.1.0` has been introduced, who depend on the 2.1.0 version of Logging.

So we have to upgrade it anyway.

PR #945, opened 2020-06-30
Wrapping IServiceConnection in WeakReference 

KKhurin:
...before storing it in async local to prevent memory leaks.

Execution contexts may keep ServiceConnection alive for much longer than the actual service connection lifetime. So we wrap it in a weak reference before storing in async local.


src/Microsoft.Azure.SignalR.Common/ServiceConnections/ClientConnectionScope.cs
```diff
@@ -52,7 +52,7 @@ public void Dispose()
 
         internal static bool IsScopeEstablished => ScopePropertiesAccessor<ClientConnectionScopeProperties>.Current != null;
 
-        internal static IServiceConnection OutboundServiceConnection
+        internal static WeakReference<IServiceConnection> OutboundServiceConnection
```
KKhurin, line 55: It would look cleaner if we returned `IServiceConnection` here instead of a `WeakReference`

PR #952, opened 2020-07-03
Change tracing id protocol 

wanlwanl:
 - change tracing id protocol: 
   - string -> long: 
       - save space
       - index is enough

### DIFF
--- a/src/Microsoft.Azure.SignalR.Protocols/ConnectionMessage.cs
+++ b/src/Microsoft.Azure.SignalR.Protocols/ConnectionMessage.cs
@@ -118,7 +118,7 @@ namespace Microsoft.Azure.SignalR.Protocol
         /// <param name="connectionId">The connection Id.</param>
         /// <param name="payload">Binary data to be delivered.</param>
         /// <param name="tracingId">The tracing Id of the message</param>
-        public ConnectionDataMessage(string connectionId, ReadOnlyMemory<byte> payload, string tracingId = null) : base(connectionId)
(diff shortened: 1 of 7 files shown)


src/Microsoft.Azure.SignalR.Protocols/ServiceMessage.cs
```diff
@@ -37,10 +37,10 @@ internal void WriteExtensionMembers(ref MessagePackWriter writer)
             }
             // todo : count more optional fields.
             writer.WriteMapHeader(count);
-            if (!string.IsNullOrEmpty(tracingId))
+            if (tracingId != null && tracingId.Value != 0)
             {
                 writer.Write(TracingId);
-                writer.Write(tracingId);
+                writer.WriteInt64(tracingId.Value);
```
vwxyzh, line 43: `WriteInt64` -> `Write`

src/Microsoft.Azure.SignalR.Protocols/ConnectionMessage.cs
```diff
@@ -145,6 +145,6 @@ public ConnectionDataMessage(string connectionId, ReadOnlySequence<byte> payload
         /// <summary>
         /// Gets or sets the message ID
         /// </summary>
-        public string TracingId { get; set; }
+        public long TracingId { get; set; }
```
vicancy, line 148: use `long?`?

wanlwanl: ulong, e.g. -1 is bad

PR #954, opened 2020-07-06
add IServerNameProvider dependency

terencefan:
add IServerNameProvider dependency


test/Microsoft.Azure.SignalR.AspNet.Tests/ServiceEndpointProviderTests.cs
```diff
@@ -26,7 +26,7 @@ public class ServiceEndpointProviderTests
         [InlineData("Endpoint=https://localhost;AccessKey=ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789;", "https://localhost/aspnetclient")]
         public async Task TestGenerateClientAccessToken(string connectionString, string expectedAudience)
         {
-            var provider = new ServiceEndpointProvider(new ServiceEndpoint(connectionString), new ServiceOptions() { });
+            var provider = new ServiceEndpointProvider(new DefaultServerNameProvider(), new ServiceEndpoint(connectionString), new ServiceOptions() { });
```
vicancy, line 29: >new DefaultServerNameProvider() [](start = 55, length = 31)

a private readonly static field is enough?

PR #971, opened 2020-08-03
add message log when the direction of service message is from server to service

wanlwanl:
 - add message log when the service message is from server to service
 - add UT

### DIFF
--- a/src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
+++ b/src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
@@ -6,6 +6,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.SignalR;
 using Microsoft.Azure.SignalR.Protocol;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 
 namespace Microsoft.Azure.SignalR.Management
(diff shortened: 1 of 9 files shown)


src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
```diff
@@ -6,14 +6,16 @@
...
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 
 namespace Microsoft.Azure.SignalR.Management
 {
     internal class WebSocketsHubLifetimeManager<THub> : ServiceLifetimeManagerBase<THub>, IHubLifetimeManagerForUserGroup where THub : Hub
     {
         public WebSocketsHubLifetimeManager(IServiceConnectionManager<THub> serviceConnectionManager, IHubProtocolResolver protocolResolver,
-            IOptions<HubOptions> globalHubOptions, IOptions<HubOptions<THub>> hubOptions) : base(serviceConnectionManager, protocolResolver, globalHubOptions, hubOptions)
+            IOptions<HubOptions> globalHubOptions, IOptions<HubOptions<THub>> hubOptions, ILoggerFactory loggerFactory) :
```
vwxyzh, line 17: `ILogger<WebSocketsHubLifetimeManager<Hub>>`?

src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
```diff
@@ -30,7 +32,8 @@ public Task UserAddToGroupAsync(string userId, string groupName, CancellationTok
             }
 
             var message = new UserJoinGroupMessage(userId, groupName).WithTracingId();
-            return ServiceConnectionContainer.WriteAsync(message);
+            Log.StartToAddUserToGroup(Logger, message);
+            return WriteAsync(message);
```
vwxyzh, line 36: `return ServiceConnectionContainer.WriteAsync(message);` -> `return WriteAsync(message);`
is it expected?

wanlwanl: Yes, `WriteAsync(message)` calls `ServiceConnectionContainer.WriteAsync(message);` inside, just add some logs
@vwxyzh

PR #977, opened 2020-08-14
fix e2e test.

vwxyzh:
fix a bug in e2e test.

### DIFF
--- a/test/Microsoft.Azure.SignalR.E2ETests/Management/ServiceHubContextE2EFacts.cs
+++ b/test/Microsoft.Azure.SignalR.E2ETests/Management/ServiceHubContextE2EFacts.cs
@@ -130,7 +130,7 @@ namespace Microsoft.Azure.SignalR.Management.Tests
                     clientEndpoint,
                     clientAccessTokens,
                     () => SendToGroupCore(serviceHubContext, userGroupDict, SendAsync, (c, d) => AddUserToGroupWithTtlAsync(c, d, TimeSpan.FromSeconds(10)), Empty),
-                    _userNames.Length / _groupNames.Length + _userNames.Length % _groupNames.Length,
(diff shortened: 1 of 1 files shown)


test/Microsoft.Azure.SignalR.E2ETests/Management/ServiceHubContextE2EFacts.cs
```diff
@@ -130,15 +130,15 @@ internal async Task TestAddUserToGroupWithTtl(ServiceTransportType serviceTransp
                     clientEndpoint,
                     clientAccessTokens,
                     () => SendToGroupCore(serviceHubContext, userGroupDict, SendAsync, (c, d) => AddUserToGroupWithTtlAsync(c, d, TimeSpan.FromSeconds(10)), Empty),
-                    _userNames.Length / _groupNames.Length + _userNames.Length % _groupNames.Length,
+                    (_userNames.Length / _groupNames.Length + _userNames.Length % _groupNames.Length) * 2,
```
wanlwanl, line 133: looks like `SendAsync()` only send to the 1st group, each group has 2 connections, so 2 should be expected?

vwxyzh: send twice: [first](https://github.com/Azure/azure-signalr/blob/54570018cc8b19f31e5f5afdbe01a696c458fd82/test/Microsoft.Azure.SignalR.E2ETests/Management/ServiceHubContextE2EFacts.cs#L351) [second](https://github.com/Azure/azure-signalr/blob/54570018cc8b19f31e5f5afdbe01a696c458fd82/test/Microsoft.Azure.SignalR.E2ETests/Management/ServiceHubContextE2EFacts.cs#L355)

PR #981, opened 2020-08-19
Catch OperationCancelledException

terencefan:


src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
```diff
@@ -76,15 +76,24 @@ public async Task ShutdownAsync()
...
+            {
+                var source = new CancellationTokenSource(_options.GracefulShutdown.Timeout);
 
-            await Task.WhenAny(
-                OfflineAndWaitForCompletedAsync(_options.GracefulShutdown.Mode),
-                Task.Delay(TimeSpan.FromHours(1), source.Token)
-            );
+                await Task.WhenAny(
+                    OfflineAndWaitForCompletedAsync(_options.GracefulShutdown.Mode),
+                    Task.Delay(TimeSpan.FromDays(1), source.Token)
```
vicancy, line 85: is 1 day too long for an application shutdown? I would say 5 minutes....

src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
```diff
@@ -76,15 +76,24 @@ public async Task ShutdownAsync()
...
+                );
 
-            await _serviceConnectionManager.StopAsync();
+                await Task.WhenAny(
+                    _serviceConnectionManager.StopAsync(),
```
vicancy, line 89: how about `StopAsync` to accept CancellationToken parameter

src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
vicancy, line 85: `Timeout.InfiniteTimeSpan`, there is also an `OrTimeout` extension method to leverage

src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
```diff
@@ -76,15 +76,24 @@ public async Task ShutdownAsync()
                 return;
             }
 
-            using CancellationTokenSource source = new CancellationTokenSource();
-            source.CancelAfter(_options.GracefulShutdown.Timeout);
+            try
+            {
+                var source = new CancellationTokenSource(_options.GracefulShutdown.Timeout);
```
vicancy, line 82: how about adding an info log here to notify customer that we are entering graceful shutdown?

src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcher.cs
```diff
@@ -76,15 +76,24 @@ public async Task ShutdownAsync()
...
 
-            await _serviceConnectionManager.StopAsync();
+                await Task.WhenAny(
+                    _serviceConnectionManager.StopAsync(),
+                    Task.Delay(TimeSpan.FromDays(1), source.Token)
+                );
+            }
+            catch (OperationCanceledException)
+            {
+                _logger.LogWarning("GracefulShutdownTimeout reached, all existing connections will be dropped immediately.");
```
vicancy, line 95: also have the value of `_options.GracefulShutdown.Timeout` in the log message?

PR #985, opened 2020-08-20
Handle ack timeout when client already disconnected

### LINKED ISSUE #750: [Improvement] AckMessage not throw timeout if related connection is already closed

JialinXin:
 - Fix ack timeout while the connection already disconnected to avoid noise.

Fixes #750


test/Microsoft.Azure.SignalR.Tests/ServiceLifetimeManagerFacts.cs
```diff
@@ -215,6 +215,83 @@ public async void TestSendConnectionAsyncisOverwrittenWhenClientConnectionExiste
...
+            proxy.ClientConnectionManager.TryAddClientConnection(context);
+
+            var task = proxy.WaitForApplicationMessageAsync(typeof(JoinGroupWithAckMessage));
+
+            var invokeTask = InvokeMethod(serviceLifetimeManager, "AddToGroupAsync");
+
+            // client quickly leaves
+            proxy.ClientConnectionManager.TryRemoveClientConnection(TestConnectionIds[0], out var conn);
+
+            if (typeof(IAckableMessage).IsAssignableFrom(typeof(JoinGroupWithAckMessage)))
```
vwxyzh, line 283: always `true`

src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManagerBase.cs
```diff
@@ -247,12 +249,21 @@ protected static bool IsInvalidArgument(IReadOnlyList<object> list)
...
+        private async Task WriteCoreAsync<T>(T message, Func<T, Task> task, string connectionId = null) where T : ServiceMessage, IMessageWithTracingId
         {
             try
             {
                 await task(message);
             }
+            catch (TimeoutException ex)
+            {
+                // Regard the case connection already disconnected when send group message got timeout as success
+                if (connectionId == null || _clientConnectionManager.ClientConnections.TryGetValue(connectionId, out var connection))
```
wanlwanl, line 261: is `connectionId == null` an `ArgumentNullException`? can we actually get a null connection id?

test/Microsoft.Azure.SignalR.Tests/ServiceLifetimeManagerFacts.cs
```diff
@@ -215,6 +215,83 @@ public async void TestSendConnectionAsyncisOverwrittenWhenClientConnectionExiste
             Assert.True(false);
         }
 
+        [Fact]
+        public async void TestSendAckMessageWithTimeoutWhenClientConnectedThrowsExcetpion()
```
wanlwanl, line 219: async void ?

src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManagerBase.cs
JialinXin: Common send message doesn't have connectionId information and it's default set null. Now only join/leave group is using ackable message.

test/Microsoft.Azure.SignalR.Tests/ServiceLifetimeManagerFacts.cs
JialinXin: copy paste errors. 😂

src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManagerBase.cs
wanlwanl: How about checking:
```
if (message is Join/LeaveGroupWithAckMessage msg && _clientConnectionManager.ClientConnections.TryGetValue(msg.ConnectionId, out var connection))
{
....
}
```
then no need to change the function parameters

src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManagerBase.cs
```diff
@@ -253,6 +255,28 @@ protected static bool IsInvalidArgument(IReadOnlyList<object> list)
             {
                 await task(message);
             }
+            catch (TimeoutException ex)
+            {
+                string connectionId = null;
+                switch (message)
+                {
+                    case JoinGroupWithAckMessage jmg:
```
vicancy, line 263: I am concerned about the extensibility of this change. How about adding an interface `IConnectionAckableMessage : IAckableMessage`?

JialinXin: Yes. Also an option considered which will include protocol change. Do we need any further discuss?

src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManagerBase.cs
```diff
@@ -253,6 +255,21 @@ protected static bool IsInvalidArgument(IReadOnlyList<object> list)
...
+            catch (TimeoutException ex)
+            {
+                string connectionId = null;
+                if (message is IConnectionAckableMessage msg)
+                {
+                    connectionId = msg.ConnectionId;
+                }
+
+                // Regard the case connection already disconnected when send group message got timeout as success
+                if (connectionId == null || _clientConnectionManager.ClientConnections.TryGetValue(connectionId, out var connection))
```
vicancy, line 267: this logic can be moved into L263

src/Microsoft.Azure.SignalR.Protocols/ServiceMessage.cs
```diff
@@ -85,6 +85,14 @@ public interface IAckableMessage
         int AckId { get; set; }
     }
 
+    /// <summary>
+    /// Interface of connection ack-able message 
+    /// </summary>
+    public interface IConnectionAckableMessage : IAckableMessage
```
vwxyzh, line 91: what about `IAckableMessageForConnection`?

src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManagerBase.cs
JialinXin: Considering this is a common method and would like to cover most situation. Do you suggest to do like this?
```
string connectionId = null;
// connection still exists and meet time-out.
if (message is IConnectionAckableMessage msg)
{
    if (_clientConnectionManager.ClientConnections.TryGetValue(connectionId, out var connection))
    {
        Log.FailedToSendMessage(Logger, message, ex);
        throw;
    }
}

// other ackable messages meet time-out
if (connectionId == null)
{
    Log.FailedToSendMessage(Logger, message, ex);
    throw;
}
```

src/Microsoft.Azure.SignalR.Protocols/ServiceMessage.cs
JialinXin: Shall we avoid preposition word in middle of naming and it seems not align with most others? For example, we don't use ConnectionMessageForOpen nor JoinGroupMessageForUser ?

src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManagerBase.cs
```diff
@@ -253,6 +255,18 @@ protected static bool IsInvalidArgument(IReadOnlyList<object> list)
             {
                 await task(message);
             }
+            catch (TimeoutException ex)
+            {
+                // Regard the case connection already disconnected when send group message got timeout as success
+                if (message is IAckableMessageWithConnectionId msg && !_clientConnectionManager.ClientConnections.TryGetValue(msg.ConnectionId, out var connection))
+                {
+                    Log.SucceededToSendMessage(Logger, message);
```
vicancy, line 263: don't log success but just ignore?

wanlwanl: I need it for message log. will check whether we need to actual log it inside.

JialinXin:
Fix in RT.

PR #987, opened 2020-08-20
small name fixes

terencefan:
 - Detail 1
 - Detail 2


src/Microsoft.Azure.SignalR.Common/Auth/AadApplicationOptions.cs
```diff
@@ -38,7 +37,7 @@ public AadApplicationOptions WithClientSecret(string secret)
             return this;
         }
 
-        public AadApplicationOptions WithClientCert(X509Certificate2 cert)
+        public AadApplicationOptions WIthCertificate(X509Certificate2 cert)
```
vicancy, line 40: typo `I` => `With`

src/Microsoft.Azure.SignalR.Common/Auth/IAadAuthOptions.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Threading.Tasks;
+
+namespace Microsoft.Azure.SignalR
+{
+    internal interface IAadAuthOptions
```
vicancy, line 5: I think `Options` in aspnetcore has some special meaning, it usually means the Options obtained from Configuration, how about changing it to some `IGenerator`?

src/Microsoft.Azure.SignalR.Common/Auth/AadApplicationOptions.cs
```diff
@@ -4,11 +4,10 @@
 using System;
 using System.Security.Cryptography.X509Certificates;
 using System.Threading.Tasks;
-using Microsoft.Azure.SignalR.Common.Auth;
 
 namespace Microsoft.Azure.SignalR
 {
-    public class AadApplicationOptions : AuthOptions, ITokenBasedAuthOptions
+    public class AadApplicationOptions : AuthOptions, IAadTokenGenerator
```
vicancy, line 10: this is a public class, so this is a breaking change? does `1.5.0` have this already?

PR #992, opened 2020-08-24
Replace manually crafted REST client with auto-generated one

Y-Sindo:
changed files can be categorized into three groups which are committed in three separate three commit:
1. `src/Common/REST/*`: files generated by Autorest
2. `src\Microsoft.Azure.SignalR.Common\REST`: enable generated REST client to authorize with JWT token.
3. other files: replace the old REST client


src/Microsoft.Azure.SignalR.Common/Generated/GeneratedRestClient.cs
```diff
@@ -0,0 +1,326 @@
+// <auto-generated>
```
wanlwanl, line 1: can auto-generated code be ignored?

src/Microsoft.Azure.SignalR.Common/Utilities/JwtTokenCredentials.cs
```diff
@@ -0,0 +1,46 @@
...
+// ----------------------------------------------------------------------------------
+
+using System;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Rest;
+
+namespace Microsoft.Azure.SignalR.Management
```
wanlwanl, line 22: should not be `Microsoft.Azure.SignalR.Management` here

src/Microsoft.Azure.SignalR.Common/Utilities/RestClient.cs
```diff
@@ -12,6 +12,7 @@
 using Microsoft.Azure.SignalR.Common;
 using Microsoft.Extensions.Primitives;
 using Newtonsoft.Json;
+using Microsoft.Azure.SignalR.Models;
```
wanlwanl, line 15: Is it used?

src/Microsoft.Azure.SignalR.Common/Utilities/JwtTokenCredentials.cs
```diff
@@ -0,0 +1,46 @@
...
+{
+    class JwtTokenCredentials : ServiceClientCredentials
```
wanlwanl, line 24: add protection level

src/Microsoft.Azure.SignalR.Common/Generated/GeneratedRestClient.cs
Y-Sindo: Currently the code is not generated during building. It has to be generated by Autorest which requireds Node.js environment. As the generated codes change little, I think it is more convenient to keep it here.

src/Microsoft.Azure.SignalR.Common/Utilities/RestClient.cs
Y-Sindo: Yes, it's used by [Payload class](https://github.com/Azure/azure-signalr/blob/00b898138dd6e0047ba174c4bf034e7c94201b30/src/Microsoft.Azure.SignalR.Common/Utilities/RestClient.cs#L128). In the past, Payload is under Microsoft.Azure.SignalR.Models. Now Payload class is generated, so the namespace is changed.

src/Microsoft.Azure.SignalR.Common/Utilities/JwtTokenCredentials.cs
Y-Sindo: fixed in [fix](https://github.com/Azure/azure-signalr/pull/992/commits/d58ec08f9e255560335ffc4bdf60be844ca8ee9b)

src/Microsoft.Azure.SignalR.Common/Utilities/JwtTokenCredentials.cs
Y-Sindo: fixed in [fix](https://github.com/Azure/azure-signalr/pull/992/commits/d58ec08f9e255560335ffc4bdf60be844ca8ee9b)

run.cmd
```diff
@@ -1,2 +0,0 @@
-@ECHO OFF
```
vicancy, line 1: revert the change?

src/Microsoft.Azure.SignalR.Common/Utilities/GeneratedRestClient.cs
```diff
@@ -0,0 +1,43 @@
...
+// limitations under the License.
+// ----------------------------------------------------------------------------------
+
+using System;
+using System.Net.Http.Headers;
+using Microsoft.Rest;
+
+namespace Microsoft.Azure.SignalR
+{
+    public partial class GeneratedRestClient
```
vicancy, line 21: a better naming, this is not Generated

Y-Sindo: This a a partial class and one part is generated.

src/Microsoft.Azure.SignalR.Common/REST/GeneratedRestClient.cs
```diff
@@ -0,0 +1,36 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using Microsoft.Rest;
+
+namespace Microsoft.Azure.SignalR
+{
+    public partial class GeneratedRestClient
+    {
+        string ProductInfo { get; }
```
vwxyzh, line 12: visibility

src/Common/REST/HealthApi.cs
```diff
@@ -0,0 +1,158 @@
...
+            cancellationToken.ThrowIfCancellationRequested();
+            _httpResponse = await Client.HttpClient.SendAsync(_httpRequest, System.Net.Http.HttpCompletionOption.ResponseHeadersRead, cancellationToken).ConfigureAwait(false);
+            if (_shouldTrace)
+            {
+                ServiceClientTracing.ReceiveResponse(_invocationId, _httpResponse);
+            }
+            HttpStatusCode _statusCode = _httpResponse.StatusCode;
+            cancellationToken.ThrowIfCancellationRequested();
+            string _responseContent = null;
+            if ((int)_statusCode != 200)
```
wanlwanl, line 124: In original logic, badgateway and serviceunavailable doesn't throw exceptions.

src/Microsoft.Azure.SignalR.Management/ServiceManager.cs
```diff
@@ -160,27 +158,15 @@ public string GenerateClientAccessToken(string hubName, string userId = null, IL
...
-                            isHealthy = false;
-                            return true;
-                        default:
-                            return false;
-                    }
-                },
-                cancellationToken: cancellationToken);
-            return isHealthy;
+            var healthApi = _restClient.HealthApi;
+            var httpOperationResponse = await healthApi.GetHealthStatusWithHttpMessagesAsync(cancellationToken: cancellationToken);
```
wanlwanl, line 162: the original code has some azure exceptions like this
```
            throw response.StatusCode switch
            {
                HttpStatusCode.BadRequest => new AzureSignalRInvalidArgumentException(response.RequestMessage.RequestUri.ToString(), innerException, detail),
                HttpStatusCode.Unauthorized => new AzureSignalRUnauthorizedException(response.RequestMessage.RequestUri.ToString(), innerException),
                HttpStatusCode.NotFound => new AzureSignalRInaccessibleEndpointException(response.RequestMessage.RequestUri.ToString(), innerException),
                _ => new AzureSignalRRuntimeException(response.RequestMessage.RequestUri.ToString(), innerException),
            };
```
Is it possible to config the exceptions for autorest client? otherwise it's a breaking change.

src/Common/REST/HealthApi.cs
wanlwanl: add tests

src/Microsoft.Azure.SignalR.Common/Generated/GeneratedRestClient.cs
wanlwanl: So why not choose a library only requires dotnet?

What about adding a script to prepare the environment like this:
https://github.com/dotnet/aspnetcore/blob/master/docs/BuildFromSource.md

Y-Sindo: Autorest is widely used in Azure-Powershell and azure-sdk-for-net projects and their common practice is just including the generated code into version control. I haven't found similar library like it only depending on dotnet, and if you know about this, we could have discussion. Actually I have tried to include the generation script into our build.cmd or build.sh and added Node.js environment into CI workflow but failed to start the build. Maybe I got something wrong with workflow files. 
I think another reason to upload the generated file is that we can review the generated code on Github.

src/Common/REST/HealthApi.cs
Y-Sindo: Yes, it's a bug. Thank you for pointing out that.

src/Microsoft.Azure.SignalR.Common/Generated/GeneratedRestClient.cs
wanlwanl: What about this one? https://github.com/RicoSuter/NSwag/wiki/CSharpClientGenerator

I think a new script is better, since build.sh/cmd is used for generating nuget package, we don't need to wait for the whole build.

Why we need to review the auto generated code, we just need to test the api whether the behavior is expected.

src/Microsoft.Azure.SignalR.Common/REST/JwtTokenCredentials.cs
```diff
@@ -0,0 +1,35 @@
...
+    internal class JwtTokenCredentials : ServiceClientCredentials
+    {
+        private RestApiAccessTokenGenerator RestApiAccessTokenGenerator { get; }
+
+        public JwtTokenCredentials(string accessKey)
+        {
+            RestApiAccessTokenGenerator = new RestApiAccessTokenGenerator(new AccessKey(accessKey));
+        }
+
+
```
wanlwanl, line 22: extra line

src/Microsoft.Azure.SignalR.Common/REST/JwtTokenCredentials.cs
```diff
@@ -0,0 +1,35 @@
...
+        public override async Task ProcessHttpRequestAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
+            if (request == null)
+            {
+                throw new ArgumentNullException("request");
```
wanlwanl, line 27: nameof(request)

src/Microsoft.Azure.SignalR.Common/REST/GeneratedRestClient.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            _productInfo = productInfo;
+        }
+
+        partial void CustomInitialize()
+        {
+            HttpClient.DefaultRequestHeaders.Add(Constants.AsrsUserAgent, _productInfo);
+        }
+
+        public static GeneratedRestClient Build(string connectionString, string productInfo)
```
wanlwanl, line 24: use `_productInfo`? remove `productInfo`?

Y-Sindo: This is a static method and productInfo will be passed to _productInfo?

src/Microsoft.Azure.SignalR.Common/REST/JwtTokenCredentials.cs
Y-Sindo: fix in https://github.com/Azure/azure-signalr/pull/992/commits/1ee9abc780816f67280c57052714e2e8087fa34e

src/Microsoft.Azure.SignalR.Common/REST/JwtTokenCredentials.cs
Y-Sindo: fix in https://github.com/Azure/azure-signalr/pull/992/commits/1ee9abc780816f67280c57052714e2e8087fa34e

src/Microsoft.Azure.SignalR.Common/REST/GeneratedRestClient.cs
wanlwanl: Your goal is to `HttpClient.DefaultRequestHeaders.Add(Constants.AsrsUserAgent, _productInfo);`, is `HttpClient` static, no need to set it every time the `GeneratedRestClient` is created?

src/Microsoft.Azure.SignalR.Common/Generated/GeneratedRestClient.cs
Y-Sindo: There is already a script (client-generation.ps1)[https://github.com/Azure/azure-signalr/pull/992/files#diff-9c01492b97d08ccb869602c016adac19]

wanlwanl:
How about exclude generated code in this PR to make it small and clean?

Y-Sindo:
> How about exclude generated code in this PR to make it small and clean?

Generated code commits in [Add REST client generated by Autorest](https://github.com/Azure/azure-signalr/pull/992/commits/016e7e0b83b33b34373578f4c20ba3d8362e14ef), you can exclude this commit to view file changes.

src/Microsoft.Azure.SignalR.Common/REST/GeneratedRestClient.cs
Y-Sindo: HttpClient is not a static field. It belongs to the base class of GeneratedRestClient.

wanlwanl:
> This is a pretty large PR, how about separating it into several pieces:
> 
> 1. Only contains the generated code, and the script file to generate it, and the doc describing the steps
> 2. The convenience layer (logic layer) leveraging the generated code
> 3. Replace the current REST client with the new one

I think Liangying suggested you split this PR into 3 PRs as above. Exclude some commit isn't convenient for reviewers, especially for those who don't see your comment.

Y-Sindo:
> > This is a pretty large PR, how about separating it into several pieces:
> > 
> > 1. Only contains the generated code, and the script file to generate it, and the doc describing the steps
> > 2. The convenience layer (logic layer) leveraging the generated code
> > 3. Replace the current REST client with the new one
> 
> I think Liangying suggested you split this PR into 3 PRs as above. Exclude some commit isn't convenient for reviewers, especially for those who don't see your comment.

Thanks for pointing out my misunderstanding. The first of the 3 PRs is #1007 .

src/Microsoft.Azure.SignalR.Management/ServiceManager.cs
Y-Sindo: According to Liangying, this breaking change is acceptable.

vicancy: does our Function output binding depend on these exceptions @zackliu? If not, is changing this exception to some general exceptions/not throw acceptable? cc @vwxyzh

wanlwanl: @vicancy It's designed for helping customer self-troubleshooting, they can catch AzureSignalRException, and the exceptions listed are having messages include some mitigate suggestions that customers can task action. Function output binding doesn't depend on them.

Issue #693, opened 2019-10-22
Fix Blazor + SignalR


BrennanConroy:
Lets discuss how and when we want to fix the issue described in https://github.com/aspnet/AspNetCore/issues/15128.

There are a few options.
We can fix the issue in a 1.1.2 patch to the SDK by checking the list of protocols for "blazorpack" and excluding it from the list of serialized messages unless it is the only protocol on the list. (This works because blazor clears the list and adds only "blazorpack").

We can wait until SignalR 3.1 ships and apply part of the same change that is going into SignalR by pulling the `HubOptions` from DI and using the "SupportedProtocols" list to determine what protocols to serialize the message to.

I'm willing to help/make the changes if we agree on a solution.

@davidfowl @anurse @vicancy @halter73

vicancy:
The impacted code is here https://github.com/Azure/azure-signalr/blob/dev/src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManagerBase.cs#L225

`ServiceLifetimeManager` is DI'ed in as a Singleton.

> (This works because blazor clears the list and adds only "blazorpack")
Is it that `IHubProtocolResolver.AllProtocols` only contains `blazorpack` when blazor is sending messages? Or there are more to change?

I'd prefer to fix it ASAP with a 1.1.2 patch.

BrennanConroy:
> Is it that `IHubProtocolResolver.AllProtocols` only contains `blazorpack` when blazor is sending messages?

No, `AllProtocols` contains all protocols from DI. But we only want "blazorpack" to be used when it's the only configured protocol. So we can inject the huboptions and inspect the SupportProtocols list and exclude blazorpack except in the case that it's the only protocol listed.

vicancy:
> inspect the SupportProtocols list and exclude `blazorpack` except in the case that it's the only protocol listed

Thanks Brennan, sounds acceptable to me.

PR #1006, opened 2020-08-28
Add BlazorDetector to require sticky.

### LINKED ISSUE #690: Turn on sticky mode for blazor app.
Blazor App is required **Sticky** but sticky mode is by default disabled for perf concern. Customers may unaware of this and encounter randomly break. 
Better to feature out a solution to improvement the experience of Blazor App users.

JialinXin:
Add `BlazorDetector` to mark hub by hub level supported protocol contains `blazorpack` and make server sticky required

Fixes #690


src/Microsoft.Azure.SignalR.Common/Constants.cs
```diff
@@ -79,5 +79,10 @@ public static class CustomizedPingTimer
             public const string ServiceStatus = "ServiceStatus";
             public const string Servers = "Servers";
         }
+
+        public static class Protocol
+        {
+            public const string BlazorPack = "blazorpack";
```
vicancy, line 85: this is a little bit tricky, is there another way instead of depending on hardcoding blazor's protocol?

JialinXin: Come up by another fix on the supported protocol: https://github.com/Azure/azure-signalr/issues/693, I think protocol is something public and stable to use. Another choice is to detect DI related components, also hard code and it varies in different versions of Blaozr.

vicancy: yeah, that other fix is planned to be removed for net5 though....

src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManager.cs
```diff
@@ -40,6 +42,11 @@ internal class ServiceLifetimeManager<THub> : ServiceLifetimeManagerBase<THub> w
             }
 #endif
             _clientConnectionManager = clientConnectionManager;
+
+            if (hubOptions.Value.SupportedProtocols != null && hubOptions.Value.SupportedProtocols.Any(x => x.Equals(Constants.Protocol.BlazorPack, StringComparison.OrdinalIgnoreCase)))
+            {
+                blazorDetector?.TrySetBlazor(typeof(THub).Name, true);
```
vicancy, line 48: DI `hubOptions` into `blazorDetector` and there is no need to have this `TrySetBlazor` method?

JialinXin: It's hub level settings. And DI `HubOptions<THub>` will introduce `THub` and hard to use in `NegotiateHandler`.

src/Microsoft.Azure.SignalR.Common/DefaultBlazorDetector.cs
```diff
@@ -0,0 +1,23 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Concurrent;
+
+namespace Microsoft.Azure.SignalR
+{
+    internal class DefaultBlazorDetector: IBlazorDetector
+    {
+        private readonly ConcurrentDictionary<string, bool> _blazor = new ConcurrentDictionary<string, bool>();
```
vicancy, line 10: should it be case insensitive, the hub name as the key?

src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManager.cs
```diff
@@ -40,6 +42,11 @@ internal class ServiceLifetimeManager<THub> : ServiceLifetimeManagerBase<THub> w
             }
 #endif
             _clientConnectionManager = clientConnectionManager;
+
+            if (hubOptions.Value.SupportedProtocols != null && hubOptions.Value.SupportedProtocols.Any(x => x.Equals(Constants.Protocol.BlazorPack, StringComparison.OrdinalIgnoreCase)))
```
vicancy, line 46: can the protocol add to `globalHubOptions` only?

src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManager.cs
vicancy: get it

src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManager.cs
JialinXin: It's set with `AddServerSideBlazor()`(https://github.com/dotnet/aspnetcore/blob/master/src/Components/Server/src/DependencyInjection/ComponentServiceCollectionExtensions.cs#L48) which only affected `ComponentHub`.

src/Microsoft.Azure.SignalR.Common/DefaultBlazorDetector.cs
JialinXin: Seems no need. Both set and get use `typeof(THub).Name` which is consistent.

vicancy (approved):

PR #1012, opened 2020-09-01
Add generated rest client for health API only

Y-Sindo:


client-generation.ps1
```diff
@@ -0,0 +1,9 @@
+node -v
+if(! $? ){
+	throw "To use Autorest to generate REST client, Node.js is required. Version 10.15.x LTS is preferred. "
+}
+autorest --info  | Out-Null
+if(-not $? ){
+	npm install -g autorest
+}
+autorest --csharp  --v3 swagger/readme.md
```
wanlwanl, line 9: please add to https://github.com/Azure/azure-signalr/blob/dev/docs/build-source.md later.

src/Common/REST/Generated/code-model-v1
```diff
@@ -0,0 +1,81 @@
...
+        "x-ms-metadata": {
+         "apiVersions": [
+          "v1"
+         ],
+         "filename": [
+          "mem:///109?oai3.shaken.json"
+         ],
+         "name": "paths·a2exod·api-v1-health·head·responses·200",
+         "originalLocations": [
+          "file:///C:/Users/zityang/source/repos/azure-signalr/src/Common/REST/health.json#/components/responses/paths·a2exod·api-v1-health·head·responses·200"
```
wanlwanl, line 42: Does it matters? I can see your alias

wanlwanl (approved):

src/Common/REST/Generated/code-model-v1
wanlwanl: Can it be relative path?

Y-Sindo: I haven't found settings about this file, and I find that there are no these file in azure-sdk-for-net repo, so I just deleted it.

client-generation.ps1
Y-Sindo: As the generated file are already in the repo, they don't need to generate themselves.

src/Common/REST/Generated/GeneratedRestClient.cs
```diff
@@ -0,0 +1,319 @@
...
+{
+    using Microsoft.Rest;
+    using Microsoft.Rest.Serialization;
+    using Newtonsoft.Json;
+    using System.Collections;
+    using System.Collections.Generic;
+    using System.Net;
+    using System.Net.Http;
+
+    public partial class GeneratedRestClient : ServiceClient<GeneratedRestClient>, IGeneratedRestClient
```
vicancy, line 21: is the auto-generated class naming customizable?

src/Common/REST/Generated/GeneratedRestClient.cs
```diff
@@ -0,0 +1,319 @@
...
+        /// An optional partial-method to perform custom initialization.
+        ///</summary>
+        partial void CustomInitialize();
+        /// <summary>
+        /// Initializes client properties.
+        /// </summary>
+        private void Initialize()
+        {
+            HealthApi = new HealthApi(this);
+            BaseUri = new System.Uri("http://localhost");
```
vicancy, line 290: what is this for?

swagger/readme.md
```diff
@@ -0,0 +1,18 @@
...
+input-file: ../src/Common/REST/health.json
+```
+
+## Alternate settings
+
+``` yaml $(csharp)
+license-header: MICROSOFT_MIT_NO_VERSION
+override-client-name: GeneratedRestClient
+namespace: Microsoft.Azure.SignalR
+output-folder: ../src/Common/REST/Generated
```
vicancy, line 17: put all these files into folder `src\Microsoft.Azure.SignalR.Common\RestClients`?

src/Common/REST/Generated/GeneratedRestClient.cs
Y-Sindo: yes

src/Common/REST/Generated/GeneratedRestClient.cs
Y-Sindo: `Initialize()` is called by each constructor. This `BaseUri = new System.Uri("http://localhost");`  setting would be covered afterwards if we pass `baseUri `parameter to the constructors. For example, see [this](https://github.com/Azure/azure-signalr/blob/8b7bece7da1c98b012e0c41ceb145901ca78c10b/src/Common/REST/Generated/GeneratedRestClient.cs#L272).

vicancy (approved):

PR #1014, opened 2020-09-02
improve message logs

wanlwanl:
 - remove condition for warning log, since other warning log about message is always logged
 - move condition out of logging functions
 - use tracing id as condition for logging. It's enough, and more accurate, since diagnostic client always generate tracing id for wanted log (filter out ping messages, control messages, etc), and tracing is always available for service message to be sent to service when "collect all" or the source is diagnostic client


JialinXin (approved):
:shipit:

PR #1026, opened 2020-09-09
add message log to asp.net

wanlwanl:
 - message arrive server from diagnostic client
 - message send to service from server

### DIFF
--- a/src/Microsoft.Azure.SignalR.AspNet/HubHost/ServiceMessageBus.cs
+++ b/src/Microsoft.Azure.SignalR.AspNet/HubHost/ServiceMessageBus.cs
@@ -8,6 +8,7 @@ using Microsoft.AspNet.SignalR;
 using Microsoft.AspNet.SignalR.Infrastructure;
 using Microsoft.AspNet.SignalR.Messaging;
 using Microsoft.Azure.SignalR.Protocol;
+using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Azure.SignalR.AspNet
 {
@@ -17,14 +18,16 @@ namespace Microsoft.Azure.SignalR.AspNet
         private readonly IServiceConnectionManager _serviceConnectionManager;
         private readonly IClientConnectionManager _clientConnectionManager;
         private readonly IAckHandler _ackHandler;
+        private readonly ILogger<ServiceMessageBus> _logger;
 
-        public ServiceMessageBus(IDependencyResolver resolver) : base(resolver)
+        public ServiceMessageBus(IDependencyResolver resolver, ILogger<ServiceMessageBus> logger) : base(resolver)
         {
             // TODO: find a more decent way instead of DI, it can be easily overriden
             _serviceConnectionManager = resolver.Resolve<IServiceConnectionManager>() ?? throw new ArgumentNullException(nameof(IServiceConnectionManager));
             _clientConnectionManager = resolver.Resolve<IClientConnectionManager>() ?? throw new ArgumentNullException(nameof(IClientConnectionManager));
             _parser = resolver.Resolve<IMessageParser>() ?? throw new ArgumentNullException(nameof(IMessageParser));
             _ackHandler = resolver.Resolve<IAckHandler>() ?? throw new ArgumentNullException(nameof(IAckHandler));
+            _logger = logger ?? throw new ArgumentNullException(nameof(ILogger<ServiceMessageBus>));
         }
 
         public override Task Publish(Message message)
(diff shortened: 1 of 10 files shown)


src/Microsoft.Azure.SignalR.AspNet/HubHost/ServiceMessageBus.cs
```diff
@@ -17,14 +18,16 @@ internal class ServiceMessageBus : MessageBus
...
 
-        public ServiceMessageBus(IDependencyResolver resolver) : base(resolver)
+        public ServiceMessageBus(IDependencyResolver resolver, ILogger<ServiceMessageBus> logger) : base(resolver)
         {
             // TODO: find a more decent way instead of DI, it can be easily overriden
             _serviceConnectionManager = resolver.Resolve<IServiceConnectionManager>() ?? throw new ArgumentNullException(nameof(IServiceConnectionManager));
             _clientConnectionManager = resolver.Resolve<IClientConnectionManager>() ?? throw new ArgumentNullException(nameof(IClientConnectionManager));
             _parser = resolver.Resolve<IMessageParser>() ?? throw new ArgumentNullException(nameof(IMessageParser));
             _ackHandler = resolver.Resolve<IAckHandler>() ?? throw new ArgumentNullException(nameof(IAckHandler));
+            _logger = logger ?? throw new ArgumentNullException(nameof(ILogger<ServiceMessageBus>));
```
vicancy, line 30: `nameof(logger)`

src/Microsoft.Azure.SignalR.AspNet/HubHost/SignalRMessageParser.cs
```diff
@@ -60,7 +63,11 @@ public IEnumerable<AppMessage> GetMessages(Message message)
                             var groupName = command.Value;
                             var connectionId = GetName(message.Key, PrefixHelper.ConnectionIdPrefix);
                             var joinGroupWithAckMessage = new JoinGroupWithAckMessage(connectionId, groupName).WithTracingId();
-                            
+                            if (joinGroupWithAckMessage.TracingId != null)
+                            { 
+                                AzureSignalRLog.StartToAddConnectionToGroup(_logger, joinGroupWithAckMessage);
```
vicancy, line 68: start doing sth....

src/Microsoft.Azure.SignalR.Common/Logging/AzureSignalRLog.cs
```diff
@@ -14,6 +14,7 @@ internal static class AzureSignalRLog
         public const string StartToBroadcastMessageTemplate = "Start to broadcast message {0}.";
```
vicancy, line 14: the class name, maybe simply `Logger`

src/Microsoft.Azure.SignalR.AspNet/HubHost/ServiceMessageBus.cs
```diff
@@ -43,14 +46,31 @@ public override Task Publish(Message message)
...
-        private Task ProcessMessage(AppMessage message)
+        private async Task ProcessMessage(AppMessage message)
         {
             if (message is HubMessage hubMessage)
             {
-                return WriteMessage(_serviceConnectionManager.WithHub(hubMessage.HubName), message);
+                try
+                {
+                    await WriteMessage(_serviceConnectionManager.WithHub(hubMessage.HubName), message);
+                    if (hubMessage.Message is IMessageWithTracingId msg && msg.TracingId != null)
```
vicancy, line 56: discussion: is moving the check into Log better?

src/Microsoft.Azure.SignalR.AspNet/HubHost/SignalRMessageParser.cs
wanlwanl: looks like both `start to do` and `start doing` are valid https://ell.stackexchange.com/questions/105862/started-to-do-and-started-doing

src/Microsoft.Azure.SignalR.AspNet/HubHost/ServiceMessageBus.cs
wanlwanl: I just keep the conventions... look at the previous lines from L26-L29

src/Microsoft.Azure.SignalR.Common/Logging/AzureSignalRLog.cs
wanlwanl: Sounds good

src/Microsoft.Azure.SignalR.AspNet/HubHost/ServiceMessageBus.cs
wanlwanl: I don't think so. Actually I try to hide the check inside Log in RT, but later I need to modify some check in one of the callers, and I am not sure in what condition I log it if I don't view the definition of the log, not convenient if you consider it.

KKhurin: Can the message be `IMessageWithTracingId` but its `TracingId` be null?

src/Microsoft.Azure.SignalR.Common/Logging/AzureSignalRLog.cs
wanlwanl: but a lot of classes already has a menber called `Logger`, they will be conflict

src/Microsoft.Azure.SignalR.AspNet/HubHost/ServiceMessageBus.cs
```diff
@@ -43,14 +46,31 @@ public override Task Publish(Message message)
...
+                    {
+                        AzureSignalRLog.FailedToSendMessage(_logger, msg, ex);
+                    }
+                    throw;
+                }
             }
 
-            return WriteMessage(_serviceConnectionManager, message);
+            await WriteMessage(_serviceConnectionManager, message);
+            return;
```
vicancy, line 73: remove redundant `return`

src/Microsoft.Azure.SignalR.AspNet/HubHost/ServiceMessageBus.cs
vicancy: yeah, maybe move the `TracingId` null check into Log? BTW, can message not `IMessageWithTracingId`? it feels annoying to do some if check before Logging everytime

wanlwanl: that's valid for a message, we only need to log in 2 cases:
* message sent from diagnostic client
* the service enable messaing log
Before messages reaches here, the tracing ID is already generated (in ServiceLifetimeManagerBase). 
So we only need to log the message with non-null tracing ID

wanlwanl: if the message isn't `IMessageWithTracingId`, I won't log it. For those messages I'm insterested, I already implement the `IMessageWithTracingId`

vicancy (approved):

wanlwanl:
Discussed offline with @vicancy 
todo in next PR:
* move check into `AzureSignalRLog`
* rename `AzureSignalRLog`

PR #1030, opened 2020-09-10
fix for writing directly to local connection

wanlwanl:

* writing directly to local connection doesn't log


JialinXin (approved):
:shipit:

PR #1040, opened 2020-09-15
Message order fix stage 4: the fix for mainstream scenarios

KKhurin:
 - implemented guaranteed message order fix for mainstream scenarios
 - added extensive tests for the message order for various scenarios
 - addressed feedback
 - fixed memleak in hot reload scenario
 - added a test for hot reload

Note: the current fix does not address some of the corner cases such as guaranteed message order in the face of concurrent writes. Even though the guarantee the SDK can provide in this case is already fairly weak (as it all depends on the order of the messages as they arrive from concurrent threads) the remaining corner cases in the SDK code might introduce additional inconsistencies (such as using several service connections). These corner case issues will be resolved in future fixes.


KKhurin:
Changing the PR to WIP - pending figuring out a new test failure.

KKhurin:
Added more tests...

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -466,32 +467,62 @@ private void OnConnectionStatusChanged(StatusChange obj)
...
             var currentConnections = ServiceConnections;
-            return WriteWithRetry(serviceMessage, currentConnections, StaticRandom.Next(-currentConnections.Count, currentConnections.Count), currentConnections.Count);
+
+            if (ClientConnectionScope.IsScopeEstablished)
+            {
+                // see if the execution context already has the connection stored for this container
+                var containers = ClientConnectionScope.OutboundServiceConnections;
+                Debug.Assert(containers != null);
+                WeakReference<IServiceConnection> connectionWr = null;
+                containers.TryGetValue(Endpoint, out connectionWr);
```
vicancy, line 481: can use `out var connectionWeakReference`?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/scopes/ClientConnectionScope.cs
```diff
@@ -14,25 +15,37 @@ internal class ClientConnectionScope : IDisposable
...
             if (!IsScopeEstablished)
             {
                 _needCleanup = true;
+
+                // The lifetime of the async local we're about to create can be much longer than some of the objects we store inside.
+                // Instances of IServiceConnection can be stopped and replaced over time and nobody should keep references to the old ones.
+                // So we keep them inside async local wrapped in weak references to avoid unnecessarily prolonging their lifetime.
+                // Instances of HubServiceEndpoint (and connection container it references) are expected to exist for the duration
+                // of the process. So we can store them without wrapping in weak reference.
+                ConcurrentDictionary<HubServiceEndpoint, WeakReference<IServiceConnection>> dict = new ConcurrentDictionary<HubServiceEndpoint, WeakReference<IServiceConnection>>();
```
vicancy, line 34: In multiple endpoint hot reload cases, `HubServiceEndpoint` can be removed, shall we clean up the keys when the endpoint is removed? Or such little leak can be ignored?

KKhurin: This indeed is a problem since we used HubServiceEndpoint as dictionary key (and store it unwrapped in async local). In case of hot reload the old HubServiceEndpoint  (which references all old connections) will continue flowing in the new execution contexts. The last PR fixes it by replacing HubServiceEndpoint with its unique id. As the result we won't leak HubServiceEndpoint. We still leak a dictionary entry (with long id and empty weakreference) per old endpoint but these don't carry any references to any heavy objects.

KKhurin:
Replaced by a new PR

PR #1042, opened 2020-09-18
Minor fix for echo method

JialinXin:

### DIFF
--- a/samples/AspNet.ChatSample/AspNet.ChatSample.SelfHostServer/ChatHub.cs
+++ b/samples/AspNet.ChatSample/AspNet.ChatSample.SelfHostServer/ChatHub.cs
@@ -32,7 +32,7 @@ namespace AspNet.ChatSample.SelfHostServer
 
         public void Echo(string message)
         {
-            Clients.Client(Context.ConnectionId).Echo($"{message} (echo from server, Client IP: {GetIpAddress()})");
+            Clients.Caller.Echo($"{message} (echo from server, Client IP: {GetIpAddress()})");
             Trace.TraceInformation("Echo...");
         }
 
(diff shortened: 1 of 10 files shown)


samples/AspNet.ChatSample/AspNet.ChatSample.SelfHostServer/ChatHub.cs
```diff
@@ -32,7 +32,7 @@ public void BroadcastMessage(string name, string message)
 
         public void Echo(string message)
         {
-            Clients.Client(Context.ConnectionId).Echo($"{message} (echo from server, Client IP: {GetIpAddress()})");
+            Clients.Caller.Echo($"{message} (echo from server, Client IP: {GetIpAddress()})");
```
vicancy, line 35: any difference?

JialinXin: `Send to connection` and `Send to caller` are calling different methods in server side, and will lead to different performance especially under sharding scenarios. `Send to connection` will send to all `ServiceEndpoints` and have a lower performance in server side.

vicancy: don't we first check if the service connection for the client `connectionId` can be found locally before broadcasting to other endpoints?

JialinXin: You're right. Both go to `SendConnectionAsync` check local first. Should be some else perf blocked during our tests. And will do some further investigation.

Meanwhile, the `Caller` method is more accurate for the `Echo` scenario, let's do the minor update.

vicancy (approved):

PR #1051, opened 2020-09-25
Add Reload Feature for Azure SignalR Services

taoli-git:
### This PR add a new experimental reload feature for Azure SignalR Services.
 - Modify the Client SDK  a new project called ReliableConnection to support reload feature in SignalR.
 - Add a new Transport Layer protocol with a special Message called RMessage.
 - Modify the ServiceConnection and ClientConnectionContext in App Server SDK to cooperate with the client for reloading.
 - Update the ChatSample and ChatSample.CSharpClient to support this feature.


vicancy:
This PR provides a good idea on how to migrate and reconnect connections.  After 4 years, stateful reconnect and server connection migration are all supported. It is fair to close this PR. Thanks again for the contribution!

PR #1056, opened 2020-09-30
Refactor generated client instantiation mode

Y-Sindo:
The old RestClientBuilder is refactored as RestClientFactory which uses IHttpClientFactory to create httpclient.

## For two reasons:

1. If instantiate SignalRServiceRestClient without passing HttpClient object as constructor parameter, then a HttpClient object is created with defalut ctor which has [known issue](https://docs.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests#issues-with-the-original-httpclient-class-available-in-net-core).

2. Get prepared for multi-endpoint scene. I think it's more consistent with persistent mode to create REST client within ServiceManager.


src/Microsoft.Azure.SignalR.Common/RestClients/RestClientFactory.cs
```diff
@@ -0,0 +1,43 @@
...
+using System.Net.Http;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Azure.SignalR
+{
+    public class RestClientFactory
+    {
+        private const string HttpClientName = "HttpClientWithUserAgent";
+        private IHttpClientFactory _httpClientFactory;
+        private protected readonly IHttpClientBuilder _httpClientBuilder;//for test, enable hook operations on the builder object before generating the _httpClientFactory.
```
vicancy, line 15: for ctor to support `IHttpClientBuilder` input?

src/Microsoft.Azure.SignalR.Common/RestClients/RestClientFactory.cs
```diff
@@ -0,0 +1,43 @@
...
+            _httpClientBuilder = new ServiceCollection()
+                .AddHttpClient(HttpClientName)
+                .ConfigureHttpClient(httpClient => httpClient.DefaultRequestHeaders.Add(Constants.AsrsUserAgent, userAgent));
+        }
+
+        internal SignalRServiceRestClient Create(ServiceEndpoint endpoint)
+        {
+            if (_httpClientFactory == null)
+            {
+                _httpClientFactory = _httpClientBuilder.Services.BuildServiceProvider().GetRequiredService<IHttpClientFactory>();
```
vicancy, line 32: move to ctor?

src/Microsoft.Azure.SignalR.Common/RestClients/RestClientFactory.cs
```diff
@@ -0,0 +1,43 @@
...
+
+namespace Microsoft.Azure.SignalR
+{
+    public class RestClientFactory
+    {
+        private const string HttpClientName = "HttpClientWithUserAgent";
+        private IHttpClientFactory _httpClientFactory;
+        private protected readonly IHttpClientBuilder _httpClientBuilder;//for test, enable hook operations on the builder object before generating the _httpClientFactory.
+
+        internal RestClientFactory(string userAgent)
```
vicancy, line 17: is there any need to customize UserAgent? How about adding it directly into `SignalRServiceRestClient` when creating it? There is a `UserAgent` property for `ServiceClient<T>`, can we leverage that?

src/Microsoft.Azure.SignalR.Common/RestClients/RestClientFactory.cs
```diff
@@ -0,0 +1,43 @@
...
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Net.Http;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Azure.SignalR
+{
+    public class RestClientFactory
```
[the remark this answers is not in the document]
Y-Sindo: Yes, to [derive the class in the test code](https://github.com/Azure/azure-signalr/blob/d1c7e7b04cbaff267cabe782fadb22a152480ffa/test/Microsoft.Azure.SignalR.Tests.Common/TestRestClientFactory.cs#L12)  and [use the derived class in another assembly](https://github.com/Azure/azure-signalr/blob/d1c7e7b04cbaff267cabe782fadb22a152480ffa/test/Microsoft.Azure.SignalR.Management.Tests/ServiceManagerFacts.cs#L121), RestClientFactory has to be public.

src/Microsoft.Azure.SignalR.Common/RestClients/RestClientFactory.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        private const string HttpClientName = "HttpClientWithUserAgent";
```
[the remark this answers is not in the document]
Y-Sindo: name of which variable?

src/Microsoft.Azure.SignalR.Common/RestClients/RestClientFactory.cs
```diff
@@ -0,0 +1,43 @@
...
+        private IHttpClientFactory _httpClientFactory;
```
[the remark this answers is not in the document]
Y-Sindo: It's not a readonly variable. It changes [here](https://github.com/Azure/azure-signalr/blob/d1c7e7b04cbaff267cabe782fadb22a152480ffa/src/Microsoft.Azure.SignalR.Common/RestClients/RestClientFactory.cs#L32)

src/Microsoft.Azure.SignalR.Common/RestClients/RestClientFactory.cs
Y-Sindo: `UserAgent` property of `ServiceClient<T>` isn't the same thing with `AsrsUserAgent'. The former stands for ""User-Agent" while the latter stands for "Asrs-User-Agent".

src/Microsoft.Azure.SignalR.Common/RestClients/RestClientFactory.cs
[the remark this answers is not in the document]
vicancy: try `InternalVisibleTo`, e.g. https://github.com/Azure/azure-signalr/blob/dev/src/Microsoft.Azure.SignalR.Common/Properties/AssemblyInfo.cs#L3

src/Microsoft.Azure.SignalR.Common/RestClients/RestClientFactory.cs
vicancy: get it. Consider adding it as part of `SignalRServiceRestClient`, it can be a mandatory header when using `SignalRServiceRestClient`

src/Microsoft.Azure.SignalR.Common/RestClients/RestClientFactory.cs
```diff
@@ -0,0 +1,36 @@
...
+    {
+        private const string HttpClientName = "HttpClientWithUserAgent";
+        private readonly IHttpClientFactory _httpClientFactory;
+        private readonly string _userAgent;
+
+        internal RestClientFactory(string userAgent, Action<IHttpClientBuilder> action = null)
+        {
+            var httpClientBuilder = new ServiceCollection()
+                .AddHttpClient(HttpClientName);
+            action?.Invoke(httpClientBuilder); //hook for test
```
vicancy, line 20: if the `action` is only designed for test, consider a better way

src/Microsoft.Azure.SignalR.Common/RestClients/SignalRServiceRestClient.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Net.Http;
+using Microsoft.Rest;
+
+namespace Microsoft.Azure.SignalR
+{
+    public partial class SignalRServiceRestClient
+    {
+        private readonly string _userAgent;
+
+        public SignalRServiceRestClient(string userAgent, ServiceClientCredentials credentials, HttpClient httpClient, bool disposeHttpClient) : this(credentials, httpClient, disposeHttpClient)
```
vicancy, line 14: Is in our scenario userAgent always `ProductInfo.GetProductInfo(Assembly.GetCallingAssembly())`? If so, we don't need to expose this property

Y-Sindo: We may want the assembly of function binding extension

src/Microsoft.Azure.SignalR.Common/RestClients/RestClientFactory.cs
```diff
@@ -0,0 +1,43 @@
...
+
+using System;
+using System.Net.Http;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.Azure.SignalR
+{
+    internal class RestClientFactory
+    {
+        private readonly Func<IHttpClientFactory, HttpClient> _genFunc = factory => factory.CreateClient();
```
vwxyzh, line 12: why not add a virtual method for creating http client?

src/Microsoft.Azure.SignalR.Common/RestClients/RestClientFactory.cs
```diff
@@ -0,0 +1,43 @@
...
+            _userAgent = userAgent;
+            _genFunc = genFunc;
+            _httpClientFactory = httpClientFactory;
+        }
+
+        internal SignalRServiceRestClient Create(ServiceEndpoint endpoint)
+        {
+            var httpClient = _genFunc.Invoke(_httpClientFactory);
+            var credentials = new JwtTokenCredentials(endpoint.AccessKey);
+            var restClient = new SignalRServiceRestClient(_userAgent, credentials, httpClient, false)
```
vwxyzh, line 36: disposeHttpClient -> true

src/Microsoft.Azure.SignalR.Management/ServiceManager.cs
```diff
@@ -47,7 +47,7 @@ internal ServiceManager(ServiceManagerOptions serviceManagerOptions, string prod
 
             _productInfo = productInfo;
 
-            _restClient = restClient;
+            _restClient = restClientFactory.Create(_endpoint);
```
vwxyzh, line 50: when dispose the `_restClient`?

src/Microsoft.Azure.SignalR.Common/RestClients/RestClientFactory.cs
Y-Sindo: [> ` It is generally not necessary to dispose of the <see cref="HttpClient"/> as the
        /// <see cref="IHttpClientFactory"/> tracks and disposes resources used by the <see cref="HttpClient"/>.`](https://source.dot.net/#Microsoft.Extensions.Http/IHttpClientFactory.cs,1b20ed1ddca77959)

vwxyzh (approved):

Issue #1058, opened 2020-10-01
Hub going offline after about 3 days


Iguanadad:
We have a project using azure-signalr which is working fine for about 3 days and then, although the server API hosting the hub is still online, no signals are broadcast from it to any of the clients.

Restarting the SignalR service in Azure does nothing, and the only way to re-establish the connections is to restart the hub server.

We are calling...
**GlobalHost.ConnectionManager.GetHubContext<PriorityInspectionHub>()**
...to get the IHubContext each time we attempt to broadcast (because when we tried keeping a static copy of the IHubContext we failed to get any messages broadcast at all) so we cannot see any way we could be getting an out of date context.

Any suggestions on how we can keep the service running for more than 3 days would be appreciated.

After turning on the detailed logging we have the following after one such failure (redacted the service_name and service_cid):
```
Microsoft.Azure.SignalR Information: 0 : [Microsoft.Azure.SignalR.Connections.Client.Internal.WebSocketsTransport]Starting transport. Transfer mode: Binary. Url: 'wss://<service_name>.service.signalr.net/aspnetserver/?hub=priorityinspection_priorityinspectionhub&cid=<service_cid>'.
Microsoft.Azure.SignalR Information: 0 : [Microsoft.Azure.SignalR.Connections.Client.Internal.WebSocketsTransport]WebSocket closed by the server. Close status NormalClosure.
Microsoft.Azure.SignalR Warning: 0 : [Microsoft.Azure.SignalR.AspNet.ServiceConnection]Connection 0115170b-2b71-416d-8b17-06e875c4efac received error message from service: Connection ping timeout.
Microsoft.Azure.SignalR Information: 0 : [Microsoft.Azure.SignalR.Connections.Client.Internal.WebSocketsTransport]Transport is stopping.
Microsoft.Azure.SignalR Information: 0 : [Microsoft.Azure.SignalR.Connections.Client.Internal.WebSocketsTransport]WebSocket closed by the server. Close status NormalClosure.
Microsoft.Azure.SignalR Warning: 0 : [Microsoft.Azure.SignalR.AspNet.ServiceConnection]Connection 09621190-9aa1-4858-803d-ef0794f8d8f7 received error message from service: Connection ping timeout.
Microsoft.Azure.SignalR Information: 0 : [Microsoft.Azure.SignalR.Connections.Client.Internal.WebSocketsTransport]Transport is stopping.
[...]

PR #1073, opened 2020-10-26
Add ConfigureAwait false

vicancy:
For `ASP.NET MVC`, users can call hub methods using `GlobalHost.ConnectionManager.GetHubContext<Hub>()` in it's Controller with `AspNetSynchronizationContext`. Adding `ConfigureAwait(false)` to prevent message send from relying on this context.


vwxyzh (approved):

vicancy:
Fixes https://github.com/Azure/azure-signalr/issues/1058

Issue #969, opened 2020-07-21
Self-Host or Emulate Azure SignalR Service for serverless local development


ryboucher:
### Is your feature request related to a problem? Please describe.
I am trying to self-host an instance of Azure SignalR service for use with my Azure functions project. This would allow me to continue working on my project even when azure servers are down.

### Describe the solution you'd like
A sample on how we could emulate or self-host azure signalR service would be great. I really want to be able to work on my Azure functions project without needing to connect to azure directly.

ModernRonin:
yes, I am looking for the exact same thing

vicancy:
Is `Upstream` used in your Azure Function? If we have a very first version of the emulator that only supports [SignalR Service output binding](https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-signalr-service-output?tabs=csharp), would that help?

ModernRonin:
For me, that would be sufficient.

ryboucher:
Same for me, I am only using output. Is that available?

vicancy:
I am trying to have a very first version this week for you to have a try.

vicancy:
Emulator Preview Version
===========

Features available
------------------
1. Auth
2. Latest Rest API support

Not yet supported
------------------
1. Upstream feature

Walkthrough
----------------
Take this serverless sample for example: https://docs.microsoft.com/en-us/azure/azure-signalr/signalr-quickstart-azure-functions-javascript?WT.mc_id=signalrquickstart-github-antchu
1. The first step **Create an Azure SignalR Service instance** can now be replaced by using the emulator
2. Install the emulator
```
dotnet tool install  -g Microsoft.Azure.SignalR.Emulator --version 1.0.0-preview1-10604 --add-source https://www.myget.org/F/azure-signalr-dev/api/v3/index.json
```
3. Run the emulator
```
asrs-emulator
```
After the emulator is successfully started, it generates the ConnectionString to be used later, for example, the ConnectionString is `Endpoint=http://localhost;Port=8088;AccessKey=ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789ABCDEFGH;Version=1.0;` as the below screenshot shows.

![image](https://user-images.githubusercontent.com/668244/91026003-bc47e800-e62c-11ea-96a6-9ac64f9bcc7e.png)

[...]

PR #1080, opened 2020-10-27
For emulator to listenAnyIp instead

vicancy:
Addresses https://github.com/Azure/azure-signalr/issues/969#issuecomment-716788379


KKhurin:
Please make listenAnyIp an option (and not the default one).

vwxyzh (approved):

PR #1081, opened 2020-10-27
Set up options and track changes in management SDK.

Y-Sindo:
### Summary of the changes 
- Move options-related class to directory "Configuration"
- set up `ServiceManagerContext` to track `ServiceManagerOptions`, set up `ServiceOptions` to track `ServiceManagerContext`
- e2e tests passed.

##### Method1: use configuration instance such as JSON file and so on. It's OK that no configuration file is found.
```C#
      services.AddSignalRServiceManager();
```

```JSON
{
    "Azure": {
        "SignalR": {
            "ApplicationName": null,
            "ConnectionCount": 1,
            "ConnectionString": "",
            "Proxy": null,
            "ServiceEndpoint": null,
            "ServiceTransportType": 0
        }
    }
}
```
##### Method2: use delegate and configuration file.
```C#
    services.AddSignalRServiceManager(o=>{
         o.ConnectionString="";
    });
```

#### `ServiceManagerBuilder`
```C#
    var manager = new ServiceManagerBuilder()
        .WithOptions(o=>o.ServiceTransportType=ServiceTransportType.Persistent)
        .WithConfiguration(new ConfigurationBuilder().AddJsonFile("config.json", reloadOnChange: true).Build())
        .Build();
````

### CODE REVIEW

File: src/Microsoft.Azure.SignalR.Management/Configuration/ServiceManagerContextSetup.cs [hunk 1]
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        public string Name => Options.DefaultName;
+
+        public void Configure(ServiceManagerContext options)
+        {
+            var serviceManagerOptions = _monitor.CurrentValue;
+            options.ServiceEndpoints = serviceManagerOptions.ServiceEndpoints ?? (serviceManagerOptions.ServiceEndpoint != null
+                    ? (new ServiceEndpoint[] { serviceManagerOptions.ServiceEndpoint })
+                    : (new ServiceEndpoint[] { new ServiceEndpoint(serviceManagerOptions.ConnectionString) }));
```
JialinXin, line 30: what if customer sets both ServiceEndpoint and ConnectionString, shall we recognize both?

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
```diff
@@ -39,16 +42,22 @@ public ServiceManagerBuilder WithCallingAssembly()
...
-
+            _services.PostConfigure<ServiceManagerOptions>(o => o.ValidateOptions());
+            _services.AddSingleton<IConfigureOptions<ServiceManagerContext>, ServiceManagerContextSetup>();
+            _serviceProvider = _services.BuildServiceProvider();
+            var context = _serviceProvider.GetRequiredService<IOptions<ServiceManagerContext>>().Value;
             var productInfo = ProductInfo.GetProductInfo(_assembly);
-            var context = new ServiceManagerContext()
-            {
-                ProductInfo = productInfo
-            };
```
JialinXin, line 48: ProductInfo is used to identify Product in UserAgent for BI use. Please help take care when switch in future version. As management SDK is also depended by function extensions.

src/Microsoft.Azure.SignalR.Management/Configuration/ServiceManagerContextSetup.cs
Y-Sindo: In `ServiceManagerOptions`, we only allow setting one of the properties. If multiple properties are set, validation function throws error.

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
Y-Sindo: I'm sorry that the code about productInfo is there but I didn't upload it.

JialinXin (approved):

src/Microsoft.Azure.SignalR.Management/DependencyInjectionExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+            services.AddSingleton<ServiceManagerContextSetup>()
+                .AddSingleton<IConfigureOptions<ServiceManagerContext>>(sp => sp.GetService<ServiceManagerContextSetup>());
+            services.AddSingleton<ServiceOptionsSetup>()
+                .AddSingleton<IConfigureOptions<ServiceOptions>>(sp => sp.GetService<ServiceOptionsSetup>());
+            if (services.Any(descriptor => descriptor.ServiceType == typeof(IOptionsChangeTokenSource<ServiceManagerOptions>)))
+            {
+                services.AddSingleton<IOptionsChangeTokenSource<ServiceManagerContext>>(sp => sp.GetService<ServiceManagerContextSetup>());
+                services.AddSingleton<IOptionsChangeTokenSource<ServiceOptions>>(sp => sp.GetService<ServiceOptionsSetup>());
+            }
+            services.WithCallingAssembly(overwrite: false);
```
vicancy, line 28: why call it by default? only Function extension wants **calling** assembly I think, others want **current** Management assembly

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
```diff
@@ -39,16 +42,22 @@ public ServiceManagerBuilder WithCallingAssembly()
         /// <returns>The instance of the <see cref="IServiceManager"/>.</returns>
         public IServiceManager Build()
         {
-            _options.ValidateOptions();
-
+            _services.PostConfigure<ServiceManagerOptions>(o => o.ValidateOptions());
```
vicancy, line 45: reuse `AddSignalRServiceManager`?

src/Microsoft.Azure.SignalR.Management/DependencyInjectionExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Reflection;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Azure.SignalR.Management
+{
+    internal static class DependencyInjectionExtensions //TODO: not ready for public use
+    {
+        /// <summary>
+        /// Add required dependencies according to the configured options <see cref="ServiceManagerOptions"/>. Please call this method after <see cref="ServiceManagerOptions"/> is configured.
```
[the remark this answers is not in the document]
Y-Sindo: if configures  ServiceManagerOptions after this call, then IOptionsChangeToken<ServiceManagerContext>, IOptionsChangeToken<ServiceOptions> wouldn't be added to service collection. Then dynamic configuration is not enabled.

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
Y-Sindo: Yes, actually in the final version it reuses it...

src/Microsoft.Azure.SignalR.Management/Microsoft.Azure.SignalR.Management.csproj
```diff
@@ -27,6 +27,10 @@
     </ItemGroup>
   </Target>
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Options.ConfigurationExtensions" Version="$(MicrosoftExtensionsOptionsConfigurationExtensionsVersion)" />
```
JialinXin, line 31: >Microsoft.Extensions.Options.ConfigurationExtensions" [](start = 31, length = 53)

Looks 3.1.9 is the latest one align with dotnetcore versions. Please check if netstandard2.0 can be applied. Otherwise, it may need to split to different frameworks like line34~41.

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
```diff
@@ -25,7 +26,7 @@ public class ServiceManagerBuilder : IServiceManagerBuilder, IDisposable
         /// <returns>The same instance of the <see cref="ServiceManagerBuilder"/> for chaining.</returns>
         public ServiceManagerBuilder WithOptions(Action<ServiceManagerOptions> configure)
         {
-            _services.Configure(configure);
+            _configure = configure;
```
vicancy: the behavior changes for `builder.WithOptions().WithOptions()...`

src/Microsoft.Azure.SignalR.Management/DependencyInjectionExtensions.cs
```diff
@@ -0,0 +1,81 @@
...
+            services.Configure(configure);
+            return services.AddSignalRServiceManagerCore();
+        }
+
+        /// <summary>
+        /// Adds the minimum essential SignalR Service Manager services to the specified services collection.
+        /// </summary>
+        /// <param name="services"></param>
+        /// <returns></returns>
+        internal static IServiceCollection AddSignalRServiceManagerCore(this IServiceCollection services)
```
vicancy, line 40: is `private` enough?

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
```diff
@@ -4,25 +4,39 @@
...
+        }
+
+        /// <summary>
+        /// Registers a configuration instance to configure <see cref="IServiceManager"/>
+        /// </summary>
+        /// <param name="config">The configuration instance.</param>
+        /// <returns>The same instance of the <see cref="ServiceManagerBuilder"/> for chaining.</returns>
+        internal ServiceManagerBuilder WithConfiguration(IConfiguration config) //todo test needed
+        {
+            _services.Configure<ServiceManagerOptions>(config.GetSection(Constants.Keys.AzureKey).GetSection(Constants.Keys.SignalRKey));
```
vicancy, line 39: what happens if no `Azure` section is found?

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
```diff
@@ -39,16 +53,21 @@ public ServiceManagerBuilder WithCallingAssembly()
         /// <returns>The instance of the <see cref="IServiceManager"/>.</returns>
         public IServiceManager Build()
         {
-            _options.ValidateOptions();
-
+            _services.AddSignalRServiceManagerCore();
```
vicancy, line 56: in general, method name with `Core` appended is a pretty private method that does not expect external calls

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
Y-Sindo: The origin version before this pull request is [Here](https://github.com/Azure/azure-signalr/blob/7f3fdbc57ae3fab8772cb0b66e672bea0e77b7ab/src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs#L25). The commit fixes the behavior change.

src/Microsoft.Azure.SignalR.Management/DependencyInjectionExtensions.cs
Y-Sindo: ServiceManagerBuilder calls this method.

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
Y-Sindo: According to doc of `GetSection` method: 

> This method will never return null. If no matching sub-section is found with the specified key, an empty Microsoft.Extensions.Configuration.IConfigurationSection will be returned.

Just like no configuration for ServiceMangerOptions, then the validation of ServiceManagerOptions will fail.

src/Microsoft.Azure.SignalR.Management/Configuration/CascadeOptionsSetup.cs
```diff
@@ -0,0 +1,48 @@
...
+    internal abstract class CascadeOptionsSetup<SourceOptions, TargetOptions> : IConfigureOptions<TargetOptions>, IOptionsChangeTokenSource<TargetOptions>
+        where SourceOptions : class
+        where TargetOptions : class
+    {
+        private protected readonly IOptionsMonitor<SourceOptions> _monitor;
+        private readonly IOptionsChangeTokenSource<SourceOptions> _changeTokenSource;
+        private readonly IChangeToken _nullChangeToken;
+
+        public CascadeOptionsSetup(IOptionsMonitor<SourceOptions> monitor, IOptionsChangeTokenSource<SourceOptions> changeTokenSource = null)
+        //Making 'tokenSource' optional avoids error when 'tokenSource' is unavailable.
```
vicancy, line 22: move above the method

src/Microsoft.Azure.SignalR.Management/Configuration/CascadeOptionsSetup.cs
```diff
@@ -0,0 +1,48 @@
...
+            }
+        }
+
+        public string Name => Options.DefaultName;
+
+        public abstract void Configure(TargetOptions options);
+
+        public IChangeToken GetChangeToken()
+        {
+            if (_changeTokenSource != null)
```
vicancy, line 38: `return _changeTokenSource?.GetChangeToken() ??  NullChangeToken.Singleton`

src/Microsoft.Azure.SignalR.Management/Configuration/ServiceOptionsSetup.cs
```diff
@@ -0,0 +1,24 @@
...
+        {
+        }
+
+        public override void Configure(ServiceOptions options)
+        {
+            var context = _monitor.CurrentValue;
+            options.ApplicationName = context.ApplicationName;
+            options.Endpoints = context.ServiceEndpoints;
+            options.Proxy = context.Proxy;
+            options.ConnectionCount = context.ConnectionCount;
```
vicancy, line 21: rebase the latest change, `ClientEndpoint`?

src/Microsoft.Azure.SignalR.Management/DependencyInjectionExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+namespace Microsoft.Azure.SignalR.Management
+{
+    internal static class DependencyInjectionExtensions //TODO: not ready for public use
+    {
+        /// <summary>
+        /// Adds the essential SignalR Service Manager services to the specified services collection and configures <see cref="ServiceManagerOptions"/> with configuration instance registered in service collection.
+        /// </summary>
+        public static IServiceCollection AddSignalRServiceManager(this IServiceCollection services)
+        {
+            services.AddSingleton<IConfigureOptions<ServiceManagerOptions>, ServiceManagerOptionsSetup>();
```
vicancy, line 19: `ServiceManagerOptionsSetup` is added twice, `new ServiceManagerOptionsSetup()` and add?

test/Microsoft.Azure.SignalR.Management.Tests/DiExtensionFacts.cs
```diff
@@ -0,0 +1,107 @@
...
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+using Newtonsoft.Json;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace Microsoft.Azure.SignalR.Management.Tests
+{
+    public class DiExtensionFacts
```
vicancy, line 17: naming convention: the name of the class to test + `Tests` or `Facts`

src/Microsoft.Azure.SignalR.Management/Configuration/ServiceOptionsSetup.cs
Y-Sindo: The latest change didn't add `ClientEndpoint` in this file.

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
```diff
@@ -4,32 +4,47 @@
...
+        }
+
+        /// <summary>
+        /// Registers a configuration instance to configure <see cref="IServiceManager"/>
+        /// </summary>
+        /// <param name="config">The configuration instance.</param>
+        /// <returns>The same instance of the <see cref="ServiceManagerBuilder"/> for chaining.</returns>
+        internal ServiceManagerBuilder WithConfiguration(IConfiguration config)
+        {
+            _services.Configure<ServiceManagerOptions>(config.GetSection(Constants.Keys.ServiceManagerOptionsSectionKey));
```
vicancy, line 39: if customer uses `ServiceManagerBuilder`, does it support hot reload?

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
```diff
@@ -4,32 +4,47 @@
...
             return this;
         }
 
         [EditorBrowsable(EditorBrowsableState.Never)]
         public ServiceManagerBuilder WithCallingAssembly()
         {
             _assembly = Assembly.GetCallingAssembly();
+            _services.WithAssembly(_assembly);
```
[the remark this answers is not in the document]
Y-Sindo: When DI pattern is fully applied, `productInfo` won't be calculated twice. Currently the one in `ServiceManagerBuilder` is for `RestClientFactory` only.

src/Microsoft.Azure.SignalR.Management/Configuration/ServiceManagerContextSetup.cs
```diff
@@ -0,0 +1,27 @@
...
+{
+    internal class ServiceManagerContextSetup : CascadeOptionsSetup<ServiceManagerOptions, ServiceManagerContext>
+    {
+        public ServiceManagerContextSetup(IOptionsMonitor<ServiceManagerOptions> monitor, IOptionsChangeTokenSource<ServiceManagerOptions> changeTokenSource = null) : base(monitor, changeTokenSource)
+        {
+        }
+
+        public override void Configure(ServiceManagerContext options)
+        {
+            var serviceManagerOptions = _monitor.CurrentValue;
```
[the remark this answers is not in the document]
Y-Sindo: Each time after `Configure` method is called, then method registered via `PostConfigure`, that is our validation method will be called.

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
Y-Sindo: Yes, the test `ConfigHotReloadTest` of `ServiceManagerBuilder` tests the hot reload .

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
```diff
@@ -39,16 +54,24 @@ public ServiceManagerBuilder WithCallingAssembly()
         /// <returns>The instance of the <see cref="IServiceManager"/>.</returns>
         public IServiceManager Build()
         {
-            _options.ValidateOptions();
-
+            _services.AddSignalRServiceManagerCore();
+            _serviceProvider = _services.BuildServiceProvider();
```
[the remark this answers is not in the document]
Y-Sindo: Check added, throw InvalidOperationException() if called twice.

src/Microsoft.Azure.SignalR.Management/DependencyInjectionExtensions.cs
Y-Sindo: Now this method is merged into AddSignalRServiceManager().

src/Microsoft.Azure.SignalR.Common/Constants.cs
```diff
@@ -9,6 +9,7 @@ internal static class Constants
     {
         public static class Keys
         {
+            public const string ServiceManagerOptionsSectionKey = "Azure:SignalR";
```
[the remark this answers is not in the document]
Y-Sindo: Move the key to `ServiceManagerOptions` as a const string named `Section`.

src/Microsoft.Azure.SignalR.Management/Configuration/ServiceManagerOptions.cs
```diff
@@ -11,6 +12,11 @@ namespace Microsoft.Azure.SignalR.Management
     /// </summary>
     public class ServiceManagerOptions
     {
+        /// <summary>
+        /// The section name used to get <see cref="ServiceManagerOptions"/> value from <see cref="IConfiguration.GetSection(string)"/>
+        /// </summary>
+        public const string Section = "Azure:SignalR";
```
vicancy, line 18: intended to expose it to customer?

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
```diff
@@ -39,16 +54,26 @@ public ServiceManagerBuilder WithCallingAssembly()
         /// <returns>The instance of the <see cref="IServiceManager"/>.</returns>
         public IServiceManager Build()
         {
-            _options.ValidateOptions();
+            if (ServiceProvider != null)
+            {
+                throw new InvalidOperationException($"Mulitple invocation of the method is not allowed.");
```
vicancy, line 59: why we have such restriction is because we want UT to test `ServiceProvider`? otherwise `ServiceProvider` can be disposed inside `Build()`?

Should the following case be a valid one?
```cs
var manager1 = builder.WithOptions(o=>o.ConnectionString="one").Build();
var manager2 = builder.WithOptions(o=>o.ConnectionString="another").Build();
```

src/Microsoft.Azure.SignalR.Management/Configuration/ServiceManagerOptions.cs
Y-Sindo: Yes, I think customer can get the right position of configuration from `Section`.

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
Y-Sindo: If `ServiceProvider` is disposed, then all the services built by service provider are disposed. We should only dispose `ServiceProvider` only when we don't need `ServiceManager` anymore.

src/Microsoft.Azure.SignalR.Management/Configuration/CascadeOptionsSetup.cs
```diff
@@ -0,0 +1,47 @@
...
+
+namespace Microsoft.Azure.SignalR.Management.Configuration
+{
+    /// <summary>
+    /// Sets up TargetOptions from ServiceManagerOptions and tracks changes .
+    /// </summary>
+    internal abstract class CascadeOptionsSetup<TargetOptions> : IConfigureOptions<TargetOptions>, IOptionsChangeTokenSource<TargetOptions>
+        where TargetOptions : class
+    {
+        private readonly ServiceManagerOptions _initialSource;
```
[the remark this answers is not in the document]
Y-Sindo: Yes, good idea.

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
Y-Sindo: Now `ServiceProvider` is disposed from `ServiceManager`

src/Microsoft.Azure.SignalR.Management/DependencyInjectionExtensions.cs
```diff
@@ -0,0 +1,59 @@
...
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public static IServiceCollection WithAssembly(this IServiceCollection services, Assembly assembly)
+        {
+            var productInfo = ProductInfo.GetProductInfo(assembly);
+            return services.Configure<ServiceManagerContext>(o => o.ProductInfo = productInfo);
+        }
+
+        private static IServiceCollection TrySetProductInfo(this IServiceCollection services)
+        {
+            var assembly = Assembly.GetExecutingAssembly();
```
vicancy, line 54: So this anyway executes even if `ProductInfo` is already set?

src/Microsoft.Azure.SignalR.Management/ServiceManager.cs
```diff
@@ -28,8 +28,10 @@ internal class ServiceManager : IServiceManager
         private readonly string _productInfo;
         private readonly ServiceManagerContext _context;
         private readonly RestClientFactory _restClientFactory;
+        private readonly ServiceProvider _serviceProvider;
 
-        internal ServiceManager(ServiceManagerContext context, RestClientFactory restClientFactory)
+        //in builder mode, serviceProvider is not null and should be disposed.
+        internal ServiceManager(ServiceManagerContext context, RestClientFactory restClientFactory, ServiceProvider serviceProvider = null)
```
vicancy, line 34: when would `ServiceProvider` be null?

src/Microsoft.Azure.SignalR.Management/Configuration/CascadeOptionsSetup.cs
```diff
@@ -0,0 +1,50 @@
...
+{
+    /// <summary>
+    /// Sets up TargetOptions from SourceOptions and tracks changes .
+    /// </summary>
+    internal abstract class CascadeOptionsSetup<TargetOptions, SourceOptions> : IConfigureOptions<TargetOptions>, IOptionsChangeTokenSource<TargetOptions>, IDisposable
+        where SourceOptions : class
+        where TargetOptions : class
+    {
+        private ConfigurationReloadToken _changeToken;
+        private IDisposable _registration;
```
vicancy, line 20: `readonly`

src/Microsoft.Azure.SignalR.Management/Configuration/CascadeOptionsSetup.cs
```diff
@@ -0,0 +1,50 @@
...
+        public CascadeOptionsSetup(IOptionsMonitor<SourceOptions> sourceMonitor)
+        {
+            _registration = sourceMonitor.OnChange(RaiseChange);
+            _sourceOptions = sourceMonitor.CurrentValue;
+            _changeToken = new ConfigurationReloadToken();
+        }
+
+        public string Name => Options.DefaultName;
+
+        public void Configure(TargetOptions target) => Convert(target, _sourceOptions);
```
vicancy, line 32: is using `sourceMonitor.CurrentValue` better?

src/Microsoft.Azure.SignalR.Management/ServiceManager.cs
```diff
@@ -141,6 +144,14 @@ public async Task<IServiceHubContext> CreateHubContextAsync(string hubName, ILog
             }
         }
 
+        public void Dispose()
+        {
+            if (_context.DisposeServiceProvider)
```
vicancy, line 149: is passing some `ownServiceProvider` parameter in .ctor doable?

src/Microsoft.Azure.SignalR.Management/Configuration/ServiceManagerOptionsSetup.cs
```diff
@@ -0,0 +1,35 @@
...
+            _configuration = configuration;
+        }
+
+        public string Name => Options.DefaultName;
+
+        public void Configure(ServiceManagerOptions options)
+        {
+            if (_configuration != null)
+            {
+                _configuration.GetSection(Constants.Keys.AzureSignalRSectionKey).Bind(options);
```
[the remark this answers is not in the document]
Y-Sindo: In here, https://github.com/Azure/azure-signalr/blob/dev/src/Microsoft.Azure.SignalR.Emulator/Startup.cs#L45, the `Configuration` variable is actually an `IConfiguration` type variable. However, in the extension methods of `ServiceCollection`, we don't have `IConfiguration`. Do you think we should add it as a parameter?

src/Microsoft.Azure.SignalR.Management/DependencyInjectionExtensions.cs
Y-Sindo: Yes, but if `ProductInfo` is not null, then the value would not be overwritten.

src/Microsoft.Azure.SignalR.Management/ServiceManager.cs
Y-Sindo: Yes, some mistakes here . Changed it to `IServiceProvider` from `ServiceProvider` and make it required .

src/Microsoft.Azure.SignalR.Management/Configuration/CascadeOptionsSetup.cs
Y-Sindo: fix

src/Microsoft.Azure.SignalR.Management/Configuration/CascadeOptionsSetup.cs
Y-Sindo: changed

vicancy (approved):

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
```diff
@@ -39,16 +54,13 @@ public ServiceManagerBuilder WithCallingAssembly()
         /// <returns>The instance of the <see cref="IServiceManager"/>.</returns>
         public IServiceManager Build()
         {
-            _options.ValidateOptions();
-
+            _services.AddSignalRServiceManager();
+            _services.Configure<ServiceManagerContext>(c => c.DisposeServiceProvider = true);
+            var serviceProvider = _services.BuildServiceProvider();
+            var context = serviceProvider.GetRequiredService<IOptions<ServiceManagerContext>>().Value;
             var productInfo = ProductInfo.GetProductInfo(_assembly);
```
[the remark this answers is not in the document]
Y-Sindo: fixed

PR #1082, opened 2020-10-27
Wanl/msglog/test

wanlwanl:
only an experiment for vendor test message log for reach release


wanlwanl:
will move to integration test

PR #1089, opened 2020-10-28
Aad auth

terencefan:
Introduce more properties into connection string.
See test cases for example.

### DIFF
--- a/samples/ChatSample/ChatSample/Startup.cs
+++ b/samples/ChatSample/ChatSample/Startup.cs
@@ -18,7 +18,7 @@ namespace ChatSample.CoreApp3
                 .AddAzureSignalR(option =>
                 {
                     option.GracefulShutdown.Mode = GracefulShutdownMode.WaitForClientsClose;
-                    option.GracefulShutdown.Timeout = TimeSpan.FromSeconds(10);
+                    option.GracefulShutdown.Timeout = TimeSpan.FromSeconds(30);
                 })
                 .AddMessagePackProtocol();
         }
--- a/src/Microsoft.Azure.SignalR.Common/Endpoints/AadAccessKey.cs
+++ b/src/Microsoft.Azure.SignalR.Common/Endpoints/AadAccessKey.cs
@@ -5,30 +5,57 @@ using System.Net.Http;
 using System.Security.Claims;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Primitives;
 using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Azure.SignalR
 {
-    internal class AadAccessKey : AccessKey
+    internal class AadAccessKey : AccessKey, IDisposable
     {
-        private readonly AuthOptions _authOptions;
+        private const int AuthorizeIntervalInMinute = 55;
+        private const int AuthorizeMaxRetryTimes = 3;
+        private const int AuthorizeRetryIntervalInSec = 3;
+
+        private static readonly TimeSpan AuthorizeInterval = TimeSpan.FromMinutes(AuthorizeIntervalInMinute);
+        private static readonly TimeSpan AuthorizeRetryInterval = TimeSpan.FromSeconds(AuthorizeRetryIntervalInSec);
+        private static readonly TimeSpan AuthorizeTimeout = TimeSpan.FromSeconds(10);
+
+        private int initialized = 0;
 
         private readonly TaskCompletionSource<bool> _authorizeTcs = new TaskCompletionSource<bool>();
 
+        private readonly TimerAwaitable _timer = new TimerAwaitable(TimeSpan.Zero, AuthorizeInterval);
+
         public bool Authorized => AuthorizeTask.IsCompleted && AuthorizeTask.Result;
+
+        public AuthOptions Options { get; }
 
         private Task<bool> AuthorizeTask => _authorizeTcs.Task;
 
-        public AadAccessKey(AuthOptions options) : base()
+        public AadAccessKey(AuthOptions options, string endpoint, int? port) : base(endpoint, port)
         {
-            _authOptions = options;
+            Options = options;
         }
 
-        public async Task AuthorizeAsync(string endpoint, int? port, string serverId, CancellationToken token = default)
+        internal async Task AuthorizeAsync(string serverId, CancellationToken token = default)
         {
             var aadToken = await GenerateAadToken();
-            await AuthorizeWithTokenAsync(endpoint, port, serverId, aadToken, token);
+            await AuthorizeWithTokenAsync(Endpoint, Port, serverId, aadToken, token);
+        }
+
+        public void Dispose()
+        {
+            ((IDisposable)_timer).Dispose();
+        }
+
+        public Task<string> GenerateAadToken()
+        {
+            if (Options is IAadTokenGenerator options)
+            {
+                return options.AcquireAccessToken();
+            }
+            throw new InvalidOperationException("This accesskey is not able to generate AccessToken, a TokenBasedAuthOptions is required.");
         }
 
         public override async Task<string> GenerateAccessToken(
@@ -41,13 +68,42 @@ namespace Microsoft.Azure.SignalR
             return await base.GenerateAccessToken(audience, claims, lifetime, algorithm);
         }
 
-        public Task<string> GenerateAadToken()
+        public async Task UpdateAccessKeyAsync(IServerNameProvider provider, ILoggerFactory loggerFactory)
         {
-            if (_authOptions is IAadTokenGenerator options)
+            if (Interlocked.CompareExchange(ref initialized, 1, 0) == 1)
             {
-                return options.AcquireAccessToken();
+                return;
             }
-            throw new InvalidOperationException("This accesskey is not able to generate AccessToken, a TokenBasedAuthOptions is required.");
+
+            _timer.Start();
+
+            var logger = loggerFactory.CreateLogger<AadAccessKey>();
+
+            while (await _timer)
+            {
+                var isAuthorized = false;
+                for (int i = 0; i < AuthorizeMaxRetryTimes; i++)
+                {
+                    var source = new CancellationTokenSource(AuthorizeTimeout);
+                    try
+                    {
+                        await AuthorizeAsync(provider.GetName(), source.Token);
+                        Log.SucceedAuthorizeAccessKey(logger, Endpoint);
+                        isAuthorized = true;
+                        break;
+                    }
+                    catch (Exception e)
+                    {
+                        Log.FailedAuthorizeAccessKey(logger, Endpoint, e);
+                        await Task.Delay(AuthorizeRetryInterval);
+                    }
+                }
+
+                if (!isAuthorized)
+                {
+                    Log.ErrorAuthorizeAccessKey(logger, Endpoint);
+                }
+            }
         }
 
         private async Task AuthorizeWithTokenAsync(string endpoint, int? port, string serverId, string accessToken, CancellationToken token = default)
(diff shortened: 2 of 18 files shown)


src/Microsoft.Azure.SignalR.Common/Endpoints/AadAccessKey.cs
```diff
@@ -31,25 +31,24 @@ public async Task AuthorizeAsync(string endpoint, int? port, string serverId, Ca
...
+            if (Options is IAadTokenGenerator options)
             {
                 return options.AcquireAccessToken();
             }
             throw new InvalidOperationException("This accesskey is not able to generate AccessToken, a TokenBasedAuthOptions is required.");
         }
 
+        public override async Task<string> GenerateAccessToken(
+                    string audience,
+            IEnumerable<Claim> claims,
```
vicancy, line 45: format

src/Microsoft.Azure.SignalR.Common/Utilities/ConnectionStringParser.cs
```diff
@@ -97,7 +115,50 @@ internal static (string endpoint, string accessKey, string version, int? port, s
                 }
             }
 
-            return (dict[EndpointProperty].TrimEnd('/'), dict[AccessKeyProperty], version, port, clientEndpoint);
+            AccessKey accessKey;
+            if (dict.ContainsKey(AuthTypeProperty) && "aad".Equals(dict[AuthTypeProperty], StringComparison.OrdinalIgnoreCase))
```
vicancy, line 119: use `enum` and `enum.TryParse` for authType

vicancy: How many authTypes do we have, currently in my mind:
```cs
msi,
aad,
keybased,
anonymous, (available for websocket)
```

thoughts?

ArchangelSDY: I think `msi` should be part of `aad` since it doesn't make sense beyond AAD scope.

src/Microsoft.Azure.SignalR.Common/Utilities/ConnectionStringParser.cs
```diff
@@ -29,12 +49,56 @@ internal static class ConnectionStringParser
...
+                var options = new AadApplicationOptions(dict[ClientIdProperty], dict[TenantIdProperty]);
+                if (dict.ContainsKey(ClientSecretProperty))
+                {
+                    return new AadAccessKey(options.WithClientSecret(dict[ClientSecretProperty]));
+                }
+                else if (dict.ContainsKey(ClientCertProperty))
+                {
+                    if (!File.Exists(dict[ClientCertProperty]))
+                    {
+                        throw new ArgumentNullException(FileNotExists, nameof(connectionString));
```
vicancy, line 70: `FileNotFoundException`?

src/Microsoft.Azure.SignalR.Common/Utilities/ConnectionStringParser.cs
```diff
@@ -97,7 +161,13 @@ internal static (string endpoint, string accessKey, string version, int? port, s
                 }
             }
 
-            return (dict[EndpointProperty].TrimEnd('/'), dict[AccessKeyProperty], version, port, clientEndpoint);
+            dict.TryGetValue(AuthTypeProperty, out string type);
+            AccessKey accessKey = type switch
```
vicancy, line 165: case insensitive

src/Microsoft.Azure.SignalR/EndpointProvider/ServiceEndpointProvider.cs
```diff
@@ -53,43 +43,9 @@ internal class ServiceEndpointProvider : IServiceEndpointProvider, IDisposable
...
-                        logger.LogWarning(e, $"Failed to authorize AccessKey, will retry in {AuthorizeRetryIntervalInSec} seconds.");
-                        await Task.Delay(AuthorizeRetryInterval);
-                    }
-                }
-
-                if (!isAuthorized)
-                {
-                    logger.LogError($"AccessKey authorized failed more than {AuthorizeMaxRetryTimes} times.");
-                }
+                _ = key.UpdateAccessKeyAsync(provider, endpoint, loggerFactory ?? NullLoggerFactory.Instance);
```
vicancy, line 48: shall the accessKey itself control when to start updating the key?

vicancy: In current design, provider is created for every request, it should be super lightweighted

terencefan: it's already super lightweight, only the first call will be executed, the method is protected by a TaskCompletionSource.

samples/ChatSample/ChatSample/Startup.cs
```diff
@@ -18,7 +18,14 @@ public void ConfigureServices(IServiceCollection services)
...
                     option.GracefulShutdown.Mode = GracefulShutdownMode.WaitForClientsClose;
-                    option.GracefulShutdown.Timeout = TimeSpan.FromSeconds(10);
+                    option.GracefulShutdown.Timeout = TimeSpan.FromSeconds(30);
+
+                    var serviceName = "aadtestcanary";
+                    var clientId = "604aca3a-0b10-4a04-8a7f-dfb59137a607";
+                    var clientSecret = "5J.~gpbA1fpLg08PmZR4Gx8H8_BBS9_jCL";
+                    var tenantId = "72f988bf-86f1-41af-91ab-2d7cd011db47";
+
+                    option.ConnectionString = $"Endpoint=https://{serviceName}.service.signalr.net;AuthType=aad;clientId={clientId};clientSecret={clientSecret};tenantId={tenantId}";
```
vicancy: this breaks the chat sample

src/Microsoft.Azure.SignalR.Common/Endpoints/AadAccessKey.cs
```diff
@@ -41,13 +68,43 @@ public async Task AuthorizeAsync(string endpoint, int? port, string serverId, Ca
             return await base.GenerateAccessToken(audience, claims, lifetime, algorithm);
         }
 
-        public Task<string> GenerateAadToken()
+        public async Task UpdateAccessKeyAsync(IServerNameProvider provider, ServiceEndpoint endpoint, ILoggerFactory loggerFactory)
         {
-            if (_authOptions is IAadTokenGenerator options)
+            if (!_initializeTcs.TrySetResult(true))
```
vicancy, line 73: a bool is enough?

test/Microsoft.Azure.SignalR.Common.Tests/Auth/AzureActiveDirectoryTests.cs
```diff
@@ -13,9 +13,9 @@ public class AzureActiveDirectoryTests
     {
         private const string IssuerEndpoint = "https://sts.windows.net/";
 
-        private const string TestClientId = "";
-        private const string TestClientSecret = "";
-        private const string TestTenantId = "";
+        private const string TestClientId = "c55771ea-a4b2-4d0f-a4db-a4cb62403e52";
```
vicancy, line 16: any security concerns?

src/Microsoft.Azure.SignalR.Common/Endpoints/AadAccessKey.cs
```diff
@@ -41,13 +68,43 @@ public async Task AuthorizeAsync(string endpoint, int? port, string serverId, Ca
             return await base.GenerateAccessToken(audience, claims, lifetime, algorithm);
         }
 
-        public Task<string> GenerateAadToken()
+        public async Task UpdateAccessKeyAsync(IServerNameProvider provider, ServiceEndpoint endpoint, ILoggerFactory loggerFactory)
```
vicancy, line 71: So what if the caller `UpdateAccessKeyAsync(endpoint1)` then `UpdateAccessKeyAsync(endpoint2)`, `endpoint2` never get authorized?

terencefan: They don't need if they shared the same `AccessKey` instance.

src/Microsoft.Azure.SignalR.Common/Endpoints/AadAccessKey.cs
terencefan: I'm not sure if the function will be called twice or more.
If the `AccessKey` could be shared among different endpoints then it's necessary to use a `TaskCompletionSource` to ensure the idempotency.

src/Microsoft.Azure.SignalR.Common/Endpoints/AadAccessKey.cs
```diff
@@ -41,13 +68,43 @@ public async Task AuthorizeAsync(string endpoint, int? port, string serverId, Ca
...
+            {
+                var isAuthorized = false;
+                for (int i = 0; i < AuthorizeMaxRetryTimes; i++)
+                {
+                    logger.LogInformation($"Try authorizing AccessKey...({i})");
+                    var source = new CancellationTokenSource(AuthorizeTimeout);
+                    try
+                    {
+                        await AuthorizeAsync(endpoint.Endpoint, endpoint.Port, provider.GetName(), source.Token);
+                        logger.LogInformation("AccessKey has been authorized successfully.");
```
vicancy, line 92: For logs, please follow such coding convention https://github.com/Azure/azure-signalr/blob/b5a82860e9af41d23f582a6c87522fb30318596f/src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpointManagerBase.cs#L333

src/Microsoft.Azure.SignalR.Common/Endpoints/AccessKey.cs
```diff
@@ -14,6 +14,10 @@ internal class AccessKey
 
         public string Value { get; protected set; }
 
+        public string Endpoint { get; internal set; }
```
vicancy, line 17: remove the property set to into ctor to make it immutable?

src/Microsoft.Azure.SignalR.Common/Utilities/ConnectionStringParser.cs
```diff
@@ -29,12 +47,56 @@ internal static class ConnectionStringParser
         private static readonly char[] PropertySeparator = { ';' };
         private static readonly char[] KeyValueSeparator = { '=' };
 
-        internal static (string endpoint, string accessKey, string version, int? port, string clientEndpoint) Parse(string connectionString)
+        private static AccessKey BuildAadAccessKey(Dictionary<string, string> dict)
```
vicancy, line 50: private methods below public/internal ones

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpoint.cs
```diff
@@ -13,7 +13,7 @@ public class ServiceEndpoint
 
         public virtual string Name { get; internal set; }
 
-        public string Endpoint { get; }
+        public string Endpoint => AccessKey?.Endpoint;
```
vicancy, line 16: AccessKey should never be null?

src/Microsoft.Azure.SignalR.Common/Utilities/ConnectionStringParser.cs
```diff
@@ -29,12 +47,56 @@ internal static class ConnectionStringParser
...
+
+        private static AccessKey BuildAccessKey(Dictionary<string, string> dict)
+        {
+            if (!dict.ContainsKey(AccessKeyProperty))
+            {
+                throw new ArgumentException(MissingAccessKeyProperty, AccessKeyProperty);
+            }
+            return new AccessKey(dict[AccessKeyProperty]);
+        }
+
```
vicancy, line 92: extra empty line

src/Microsoft.Azure.SignalR.Common/Utilities/ConnectionStringParser.cs
```diff
@@ -29,12 +47,56 @@ internal static class ConnectionStringParser
...
+            }
+            else
+            {
+                return new AadAccessKey(new AadManagedIdentityOptions());
+            }
+        }
+
+        private static AccessKey BuildAccessKey(Dictionary<string, string> dict)
+        {
+            if (!dict.ContainsKey(AccessKeyProperty))
```
vicancy, line 86: use `dict.TryGetValue` to avoid get twice

src/Microsoft.Azure.SignalR.Common/Endpoints/AccessKey.cs
```diff
@@ -11,16 +11,20 @@ namespace Microsoft.Azure.SignalR
     internal class AccessKey
     {
         public string Id { get; protected set; }
```
vicancy, line 13: also remove set for `Id` and `Value`?

terencefan: The value for `Id` and `Value` in `AadAccessKey` (which is inherited from `AccessKey`) are updated from the `GetAccessKey` API results. So it needs to be settable and the level should be **protected**;

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpoint.cs
terencefan: I'm afraid we don't prevent this at the compile level.
And I think for a dynamic property, it is much reasonable to acquire a `null` value rather than get a *NullReferenceException* while something unexpected happens.

vicancy (approved):

PR #1101, opened 2020-11-11
Upgrade to 5.0.100

JialinXin:
Upgrade core version.


vicancy:
Nice... Also add net5 target framework for our SDK?

src/Microsoft.Azure.SignalR.Common/Microsoft.Azure.SignalR.Common.csproj
```diff
@@ -5,7 +5,7 @@
   </PropertyGroup>
   
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;netcoreapp3.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netcoreapp3.1;net5.0</TargetFrameworks>
```
vicancy, line 8: shall we still keep 3.0? otherwise what if there are some security bugs in our old packages targeting 3.0? We are different from SignalR in that SignalR is part of ASPNETCORE framework but we are a standalone package.

JialinXin: Suggest under these cases we should push customer to upgrade to LTS 3.1 with latest SDK version. Customer has to upgrade SDK and not much difference from upgrade to .net core 3.1 at same time.

Besides from our BI data, most users are keep up with LTS version. So suggest to align with LTS version for a better management.

JialinXin: Let's maintain 2.0,3.0 and 5.0

test/Microsoft.Azure.SignalR.IntegrationTests/MessageOrderTests.cs
```diff
@@ -98,35 +99,35 @@ public async Task OutgoingMessagesUseSameServiceConnection()
                     // finally, get ready to verify the order of messages
                     int actualCallbackNum = (int)invMsg.Arguments[0];
 
-                    if (MessageOrderFixed)
-                    {
-                        // this check works for both primary and secondary connections
-                        Assert.Equal(actualCallbackNum, msgCount);
-                    }
+                    // TODO: unreable code
```
vicancy, line 102: ?

JialinXin: Fix some build warnings. The part of code is not working now.

JialinXin: reverted

vicancy (approved):

PR #1105, opened 2020-11-13
Use relative path in ServiceEndpoint key parsing

Y-Sindo:
 - Use relative path in ServiceEndpoint key parsing
 - Extract `ServiceEndpoint` key parsing from `ServiceOptionsSetup` to `IConfigurationExtension` in preparation for multi-endpoint configuration in Azure Function extension.


src/Microsoft.Azure.SignalR.Common/Endpoints/IConfigurationExtension.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Extensions.Configuration;
+
+namespace Microsoft.Azure.SignalR.Common.Endpoints
+{
+    public static class IConfigurationExtension
```
vicancy, line 10: internal

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpoint.cs
```diff
@@ -139,44 +136,24 @@ public override bool Equals(object obj)
 
         internal static (string, EndpointType) ParseKey(string key)
         {
-            if (key == Constants.Keys.ConnectionStringDefaultKey || key == Constants.Keys.ConnectionStringSecondaryKey)
+            if (string.IsNullOrWhiteSpace(key))
```
vicancy, line 139: `IsNullOrEmpty`

src/Microsoft.Azure.SignalR.Common/Endpoints/IConfigurationExtension.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        public static (string ConnectionString, ServiceEndpoint[]) GetSignalRServiceEndpoints(this IConfiguration configuration, string sectionKey)
```
vicancy, line 12: what does the return value mean? one connectionstring but mulitple endpoints? is this `connectionString` another `service endpoint`? Is it a valid case that both have value?

src/Microsoft.Azure.SignalR.Common/Endpoints/IConfigurationExtension.cs
Y-Sindo: This will used by function extension?

src/Microsoft.Azure.SignalR.Common/Endpoints/IConfigurationExtension.cs
Y-Sindo: `ConnectionString` is the value with key `Azure:SignalR:ConnectionString`, `ServiceEndpoint[]` are endpoints with key `Azure:SignalR:ConnectionString:...` whose name or `ServiceEndpointType` is explicitly defined.

src/Microsoft.Azure.SignalR.Common/Endpoints/IConfigurationExtension.cs
Y-Sindo: `[EditorBrowsable(EditorBrowsableState.Never)]` added.

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpoint.cs
Y-Sindo: fixed.

src/Microsoft.Azure.SignalR.Common/Endpoints/IConfigurationExtension.cs
vicancy: so connectionString is another ServiceEndpoint with some default properties?

src/Microsoft.Azure.SignalR.Common/Endpoints/IConfigurationExtension.cs
vicancy: is it possible to internal visible to the function extension instead?

src/Microsoft.Azure.SignalR.Common/Endpoints/IConfigurationExtension.cs
Y-Sindo: This is a interface from original [`ServiceOptionsSetup`](https://github.com/Azure/azure-signalr/blob/0ff0e066af962927f84ef8e86aa045a0f6acbcee/src/Microsoft.Azure.SignalR/ServiceOptionsSetup.cs#L60), following this interface makes things easier. I think why sperate `Azure:SignalR:ConnectionString` and others service endpoint is that `Azure:SignalR:ConnectionString` is on the higher configuration lever than other keys, and they are parse in different ways.

src/Microsoft.Azure.SignalR.Common/Endpoints/IConfigurationExtension.cs
Y-Sindo: Finished.

test/Microsoft.Azure.SignalR.Tests/ServiceOptionsSetupFacts.cs
```diff
@@ -0,0 +1,64 @@
...
+            var defaultConnectionString = "Endpoint=http://default;AccessKey=ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789;Port=8080;Version=1.0";
+            configuration[$"{Constants.Keys.ConnectionStringDefaultKey}"] = defaultConnectionString;
+            foreach(var key in EndpointDict.Keys){
+                configuration[ConfigurationPath.Combine(Constants.Keys.ConnectionStringDefaultKey, key)] = FakeConnectionString;
+            }
+
+            var setup = new ServiceOptionsSetup(configuration);
+            var options = new ServiceOptions();
+            setup.Configure(options);
+
```
vicancy, line 58: remove the redundant empty line

test/Microsoft.Azure.SignalR.Tests/ServiceOptionsSetupFacts.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            Assert.Equal(defaultConnectionString, options.ConnectionString);
+            Assert.Equal(EndpointDict.AsEnumerable().Count(), options.Endpoints.Length);
```
vicancy, line 61: `EndpointDict.Length`

vicancy (approved):

PR #1124, opened 2020-11-23
Throw exc while key doesn't have read accesskey permit.

terencefan:
1. Throw `AzureSignalRAccessKeyNotAuthorizedException` while the given Accesskey doesn't have permission to generate `AccessToken` for clients.
2. Add exception handler for both `ASP.NET Core SignalR` and `ASP.NET SignalR`.

### DIFF
--- a/src/Microsoft.Azure.SignalR.AspNet/Middleware/NegotiateMiddleware.cs
+++ b/src/Microsoft.Azure.SignalR.AspNet/Middleware/NegotiateMiddleware.cs
@@ -173,8 +173,16 @@ namespace Microsoft.Azure.SignalR.AspNet
             }
 
             var url = provider.GetClientEndpoint(null, originalPath, queryString);
-
-            return GenerateClientAccessTokenAsync(provider, context, url, claims);
+            try
+            {
+                return GenerateClientAccessTokenAsync(provider, context, url, claims);
+            }
+            catch (AzureSignalRAccessTokenNotAuthorizedException e)
+            {
+                Log.NegotiateFailed(_logger, e.Message);
+                context.Response.StatusCode = 500;
+                return context.Response.End("");
+            }
         }
 
         private async Task GenerateClientAccessTokenAsync(
--- a/src/Microsoft.Azure.SignalR.Common/Endpoints/AadAccessKey.cs
+++ b/src/Microsoft.Azure.SignalR.Common/Endpoints/AadAccessKey.cs
@@ -5,6 +5,7 @@ using System.Net.Http;
 using System.Security.Claims;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Azure.SignalR.Common;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Primitives;
 using Newtonsoft.Json.Linq;
@@ -21,27 +22,19 @@ namespace Microsoft.Azure.SignalR
         private static readonly TimeSpan AuthorizeRetryInterval = TimeSpan.FromSeconds(AuthorizeRetryIntervalInSec);
         private static readonly TimeSpan AuthorizeTimeout = TimeSpan.FromSeconds(10);
 
-        private int initialized = 0;
-
-        private readonly TaskCompletionSource<bool> _authorizeTcs = new TaskCompletionSource<bool>();
-
+        private readonly TaskCompletionSource<object> _initializedTcs = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
         private readonly TimerAwaitable _timer = new TimerAwaitable(TimeSpan.Zero, AuthorizeInterval);
-
-        public bool Authorized => AuthorizeTask.IsCompleted && AuthorizeTask.Result;
+        private volatile bool _isAuthorized = false;
+        private int initialized = 0;
+        public bool Authorized => InitializedTask.IsCompleted && _isAuthorized;
 
         public AuthOptions Options { get; }
 
-        private Task<bool> AuthorizeTask => _authorizeTcs.Task;
+        private Task<object> InitializedTask => _initializedTcs.Task;
 
         public AadAccessKey(AuthOptions options, string endpoint, int? port) : base(endpoint, port)
         {
             Options = options;
-        }
-
-        internal async Task AuthorizeAsync(string serverId, CancellationToken token = default)
-        {
-            var aadToken = await GenerateAadToken();
-            await AuthorizeWithTokenAsync(Endpoint, Port, serverId, aadToken, token);
         }
 
         public void Dispose()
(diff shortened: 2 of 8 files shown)


src/Microsoft.Azure.SignalR.Common/Endpoints/AadAccessKey.cs
```diff
@@ -64,46 +67,54 @@ public Task<string> GenerateAadToken()
...
+            await InitializedTask;
+            if (!Authorized)
+            {
+                throw new AzureSignalRAccessTokenNotAuthorizedException();
+            }
             return await base.GenerateAccessToken(audience, claims, lifetime, algorithm);
         }
 
-        public async Task UpdateAccessKeyAsync(IServerNameProvider provider, ILoggerFactory loggerFactory)
+        private async Task UpdateAccessKeyAsyncInner(IServerNameProvider provider, ILoggerFactory loggerFactory)
```
vicancy, line 78: naming: append `Core` as convention

vicancy: private method below public ones

src/Microsoft.Azure.SignalR.Common/Endpoints/AadAccessKey.cs
```diff
@@ -125,7 +136,7 @@ private async Task AuthorizeWithTokenAsync(string endpoint, int? port, string se
                 cancellationToken: token);
         }
 
-        private async Task<bool> HandleHttpResponseAsync(HttpResponseMessage response)
+        private async Task<bool> HandleHttpResponseAsyncInner(HttpResponseMessage response)
```
vicancy, line 139: `HandleHttpResponseAsyncCore`

src/Microsoft.Azure.SignalR.Common/Exceptions/AzureSignalRAccessTokenNotAuthorizedException.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.SignalR.Common
+{
+    /// <summary>
+    /// The exception throws when AccessKey is not authorized.
+    /// </summary>
+    class AzureSignalRAccessTokenNotAuthorizedException : AzureSignalRException
```
vicancy, line 9: add explicit `internal`

src/Microsoft.Azure.SignalR.Common/Exceptions/AzureSignalRAccessTokenNotAuthorizedException.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+         /// <summary>
+        /// Initializes a new instance of the <see cref="AzureSignalRAccessTokenTooLongException"/> class.
```
vicancy, line 12: typo AzureSignalRAccessTokenTooLongException

src/Microsoft.Azure.SignalR.Common/Endpoints/AadAccessKey.cs
```diff
@@ -23,15 +24,17 @@ internal class AadAccessKey : AccessKey, IDisposable
 
         private int initialized = 0;
 
-        private readonly TaskCompletionSource<bool> _authorizeTcs = new TaskCompletionSource<bool>();
+        private readonly TaskCompletionSource<bool> _initializedTcs = new TaskCompletionSource<bool>();
```
vicancy, line 27: use `new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously)`

vicancy: nit: looks like `<object>` and `TrySetResult(null)` is enough

src/Microsoft.Azure.SignalR.Common/Endpoints/AadAccessKey.cs
```diff
@@ -74,36 +71,13 @@ public async Task UpdateAccessKeyAsync(IServerNameProvider provider, ILoggerFact
             {
                 return;
             }
+            await Task.Run(() => UpdateAccessKeyAsyncCore(provider, loggerFactory));
```
vicancy, line 74: we don't need to `await` the result? I think for this public method, it is only a trigger, we never expect the caller to await for the result

src/Microsoft.Azure.SignalR.Common/Endpoints/AadAccessKey.cs
```diff
@@ -137,18 +121,50 @@ private async Task<bool> HandleHttpResponseAsync(HttpResponseMessage response)
...
+                    var source = new CancellationTokenSource(AuthorizeTimeout);
+                    try
+                    {
+                        await AuthorizeAsync(provider.GetName(), source.Token);
+                        Log.SucceedAuthorizeAccessKey(logger, Endpoint);
+                        break;
+                    }
+                    catch (Exception e)
+                    {
+                        _isAuthorized = false;
```
vicancy, line 154: should this `false` set move outside of the retry logic, considering the case when the previous auth is succeeded?

terencefan: If the previous auth is succeeded, we will simply break the for...loop and there will be no retry afterward.

src/Microsoft.Azure.SignalR/Utilities/ServiceRouteHelper.cs
```diff
@@ -55,6 +55,12 @@ internal class ServiceRouteHelper
                 await context.Response.WriteAsync(ex.Message);
                 return;
             }
+            catch (AzureSignalRAccessTokenNotAuthorizedException ex)
```
vicancy, line 58: add ut for these?

src/Microsoft.Azure.SignalR.AspNet/Middleware/NegotiateMiddleware.cs
```diff
@@ -173,8 +173,16 @@ private Task ProcessNegotiationRequest(IOwinContext owinContext, HostContext con
             }
 
             var url = provider.GetClientEndpoint(null, originalPath, queryString);
-
-            return GenerateClientAccessTokenAsync(provider, context, url, claims);
+            try
+            {
+                return GenerateClientAccessTokenAsync(provider, context, url, claims);
+            }
+            catch (AzureSignalRAccessTokenNotAuthorizedException e)
```
vicancy, line 180: add ut for it?

vicancy (approved):

PR #1126, opened 2020-11-24
ConnectionInitializedTask in a connection container finishes when any connection initialized

Y-Sindo:


vicancy (approved):

vicancy:
So it is possible that ConnectionInitializedTask returns false that the connection failed to establish

Y-Sindo:
> So it is possible that ConnectionInitializedTask returns false that the connection failed to establish

I think yes. https://github.com/Azure/azure-signalr/blob/27fefb444df49c85fb8124c94810a081e6d4cd22/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs#L185

wanlwanl (approved):

vicancy:
So the correct logic should sound like: WhenAny return `true` or WhenAll return `false`?

PR #1128, opened 2020-11-24
support negotiate response options for serverless

wanlwanl:
 - Detail 1
 - Detail 2

### DIFF
--- a/src/Microsoft.Azure.SignalR.Management/Configuration/NegotiationResponseOptions.cs
+++ b/src/Microsoft.Azure.SignalR.Management/Configuration/NegotiationResponseOptions.cs
@@ -1,0 +1,23 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Microsoft.Azure.SignalR.Management
+{
+    /// <summary>
+    /// Configuarable options for negotiation response options for redirecting client to SignalR service.
+    /// </summary>
+    public class NegotiationResponseOptions
+    {
+        /// <summary>
+        /// Gets or sets the flag whether the negotation response contains diagnostic client marker.
+        /// </summary>
+        public bool IsDiagnosticClient { get; set; } = false;
+
+        /// <summary>
+        /// The interval used by the server to timeout incoming handshake requests in SignalR service by clients. The default timeout is 15 seconds. The valid timeout is between 1 second and 30 seconds.
+        /// </summary>
+        public TimeSpan? HandshakeTimeout { get; set; } = null;
+    }
+}
--- a/src/Microsoft.Azure.SignalR.Management/Configuration/ServiceManagerContext.cs
+++ b/src/Microsoft.Azure.SignalR.Management/Configuration/ServiceManagerContext.cs
@@ -22,6 +22,8 @@ namespace Microsoft.Azure.SignalR.Management
 
         public bool DisposeServiceProvider { get; set; } = false;
 
+        public NegotiationResponseOptions NegotiationResponseOptions { get; set; } = new NegotiationResponseOptions();
+
         internal void ValidateOptions()
         {
             if (ServiceEndpoints.Length == 0)
--- a/src/Microsoft.Azure.SignalR.Management/Configuration/ServiceManagerOptions.cs
+++ b/src/Microsoft.Azure.SignalR.Management/Configuration/ServiceManagerOptions.cs
@@ -43,5 +43,10 @@ namespace Microsoft.Azure.SignalR.Management
         /// Gets or sets the transport type to Azure SignalR Service. Default value is Transient.
         /// </summary>
         public ServiceTransportType ServiceTransportType { get; set; } = ServiceTransportType.Transient;
+
+        /// <summary>
+        /// Gets or sets negotiation response options for redirecting client to SignalR service.
+        /// </summary>
+        public NegotiationResponseOptions NegotiationResponseOptions { get; set; } = new NegotiationResponseOptions();
     }
 }
--- a/src/Microsoft.Azure.SignalR.Common/Utilities/HandshakeTimeoutHelper.cs
+++ b/src/Microsoft.Azure.SignalR.Common/Utilities/HandshakeTimeoutHelper.cs
@@ -1,0 +1,68 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using Microsoft.Extensions.Logging;
+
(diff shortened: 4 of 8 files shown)


src/Microsoft.Azure.SignalR.Management/Configuration/ServiceManagerContext.cs
```diff
@@ -22,6 +22,8 @@ internal class ServiceManagerContext
 
         public bool DisposeServiceProvider { get; set; } = false;
 
+        public NegotiationResponseOptions NegotiationResponseOptions { get; set; } = new NegotiationResponseOptions();
```
vicancy, line 25: context is the inner converted options, it should not have the default value if it is converted from the customer faced `ServiceManagerOptions`, @Y-Sindo shall we consider adding some coding guidance in the comment here?

src/Microsoft.Azure.SignalR.Management/Configuration/ServiceManagerOptions.cs
```diff
@@ -43,5 +43,10 @@ public class ServiceManagerOptions : IServiceEndpointOptions
         /// Gets or sets the transport type to Azure SignalR Service. Default value is Transient.
         /// </summary>
         public ServiceTransportType ServiceTransportType { get; set; } = ServiceTransportType.Transient;
+
+        /// <summary>
+        /// Gets or sets negotiation response options for redirecting client to SignalR service.
+        /// </summary>
+        public NegotiationResponseOptions NegotiationResponseOptions { get; set; } = new NegotiationResponseOptions();
```
vicancy, line 50: I feel that we don't need this inner wrapped option, flattened options are enough. Otherwise, customer needs to understand `negotiate`, `redirect` these technical details at the very beginning.

wanlwanl: The `HttpConnectionOptions` used in `HubConnectionBuilder` already has an option `public bool SkipNegotiation { get; set; }`, looks like they don't mind the term `negotiation`. I think these options will be configured by advanced users. For begineers, they can used the default value. 

I thinks I can add more in comment for `NegotiationResponseOptions` - it will ber only used in `GenerateClientAccessToken` to pass client options to ASRS.

What about I rename the `NegotiationResponseOptions` to `ClientOptions`? And comments avoid involving tech terms? e.g. IsDiagnosticClient - "Gets or sets the flag whether marking SignalR client as a diagnostic client."

Y-Sindo: I think it's OK to have a hierarchy in the options to make the structure more clear. But the field name can be more succinct, because the field name is also the key for field in configuration files.  I think `ClientOptions` can be rename to `Client`. `IsDiagnosticClient` can be rename to `Diagnostic`.

src/Microsoft.Azure.SignalR.Management/Configuration/NegotiationResponseOptions.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        /// <summary>
+        /// Gets or sets the flag whether the negotation response contains diagnostic client marker.
+        /// </summary>
+        public bool IsDiagnosticClient { get; set; } = false;
+
+        /// <summary>
+        /// The interval used by the server to timeout incoming handshake requests in SignalR service by clients. The default timeout is 15 seconds. The valid timeout is between 1 second and 30 seconds.
+        /// </summary>
+        public TimeSpan? HandshakeTimeout { get; set; } = null;
```
Y-Sindo, line 21: Can `TimeSpan` type be resolved by the built-in `Bind` method? See [here ](https://github.com/Azure/azure-signalr/blob/a53d7b481aedcdc8fb5bbd93a04ad620968f8c9f/src/Microsoft.Azure.SignalR.Management/Configuration/ServiceManagerOptionsSetup.cs#L26). May need docs about configuring `TimeSpan` with string.

src/Microsoft.Azure.SignalR.Management/Configuration/ServiceManagerOptions.cs
wanlwanl: `Client` looks weird. I think `ClientConfiguration` like `public Action<ClientWebSocketOptions> WebSocketConfiguration { get; set; }` is better.
`Diagnostic` is too vague.

src/Microsoft.Azure.SignalR.Management/Configuration/NegotiationResponseOptions.cs
wanlwanl: See here: https://blog.ploeh.dk/2019/11/25/timespan-configuration-values-in-net-core/
built-in. doc can be in another pr for these options.

src/Microsoft.Azure.SignalR.Management/Configuration/ServiceManagerOptions.cs
Y-Sindo: All fields in `ServiceManagerOptions` are for configuration. So `configuration` is duplicate.

Y-Sindo: `public Action<ClientWebSocketOptions> WebSocketConfiguration { get; set; }` . If `WebSocketConfiguration` is not used as configuration key, then it is OK. But it is too prolix as a configuration key. Imagine users have to configure `Azure:SignalR:ClientConfiguration:IsDiagnosticClient=true`

wanlwanl: Let's call it `Client`, `IsDiagnosticClient` and `HandshakeTimeout`

src/Microsoft.Azure.SignalR.Management/Configuration/NegotiationResponseOptions.cs
```diff
@@ -0,0 +1,23 @@
...
+    /// </summary>
+    public class NegotiationResponseOptions
+    {
+        /// <summary>
+        /// Gets or sets the flag whether the negotation response contains diagnostic client marker.
+        /// </summary>
+        public bool IsDiagnosticClient { get; set; } = false;
+
+        /// <summary>
+        /// The interval used by the server to timeout incoming handshake requests in SignalR service by clients. The default timeout is 15 seconds. The valid timeout is between 1 second and 30 seconds.
```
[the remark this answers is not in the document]
wanlwanl: What if cx change the hanshakeTimeout to an invalid? does validation also check it?

wanlwanl:
DiagnosticClientFilter with HttpContext ctx is better. support later in the new service manager. 
Current workaround is to add claim manually.

PR #1129, opened 2020-11-24
prepare to obsolete service manager

Y-Sindo:
 - move service manager to obsolete directory 
 - rename ServiceManagerContext to ContextOptions
 - always dispose `ServiceProvider` inside `ServiceManager` when dispose `ServiceManager`


src/Microsoft.Azure.SignalR.Management/Obsolete/ServiceManager.cs
```diff
@@ -9,51 +9,33 @@
...
         public Task<IServiceHubContext> CreateHubContextAsync(string hubName, ILoggerFactory loggerFactory = null, CancellationToken cancellationToken = default) =>
             _serviceHubContextFactory.CreateAsync(hubName, loggerFactory, cancellationToken);
 
         public void Dispose()
         {
-            if (_disposeServiceProvider)
-            {
-                (_serviceProvider as IDisposable).Dispose();
-            }
+            (_serviceProvider as IDisposable).Dispose();
```
vwxyzh, line 38: `(_serviceProvider as IDisposable)?.Dispose();`

Y-Sindo: fixed.

Y-Sindo: I think if `ServiceProvider` is not a `IDisposable`, then memory leak occurs. So `ServiceProvider` must be `IDisposable`.

vicancy (approved):

PR #1133, opened 2020-12-01
add protocol in OpenConnectionMessage.

vwxyzh:
### Summary of the changes
add protocol field in OpenConnectionMessage.


src/Microsoft.Azure.SignalR.Protocols/IHasProtocol.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Microsoft.Azure.SignalR.Protocol
```
wanlwanl, line 1: license

wanlwanl (approved):

vicancy (approved):

PR #1134, opened 2020-12-01
add docs for service context options

Y-Sindo:


src/Microsoft.Azure.SignalR.Management/Configuration/ContextOptions.cs
```diff
@@ -12,14 +12,25 @@ internal class ContextOptions
         //Users not allowed to configure it
         internal string ProductInfo { get; set; }
 
+        /// <summary>
+        /// Gets or sets the service endpoints.
+        /// </summary>
         public ServiceEndpoint[] ServiceEndpoints { get; set; }
 
+        /// <summary>
+        /// Gets or sets the ApplicationName which will be prefixed to each hub name
```
vicancy, line 21: `which will be prefixed to each hub name` this is the implementation detail, hide it?
> Gets or sets the ApplicationName, it can be useful when you'd like to share the same Azure SignalR service for different applications with the same Hub name

src/Microsoft.Azure.SignalR.Management/Configuration/ContextOptions.cs
```diff
@@ -12,14 +12,25 @@ internal class ContextOptions
...
+        /// </summary>
         public string ApplicationName { get; set; }
 
+        /// <summary>
+        /// Gets or sets the total number of connections from SDK to Azure SignalR Service. Default value is 1.
+        /// </summary>
         public int ConnectionCount { get; set; }
```
vicancy, line 28: Do we want our users to customize it? Give it a value, e.g. `3`?

src/Microsoft.Azure.SignalR.Management/Configuration/ContextOptions.cs
```diff
@@ -12,14 +12,25 @@ internal class ContextOptions
...
 
+        /// <summary>
+        /// Gets or sets the proxy used when ServiceManager will attempt to connect to Azure SignalR Service.
```
vicancy, line 31: `will attempt` => `attempts`

src/Microsoft.Azure.SignalR.Management/Configuration/ContextOptions.cs
Y-Sindo: I'm going to turn `ConnectionCount` into internal in this PR https://github.com/Azure/azure-signalr/pull/1135

KKhurin: A bit more complicated, but maybe `bool ConnectionNumberAutoconfig` - that implies SDK will try to chose an optimum number (~ number of pods?). This way neither us nor customers will have to guess the magic number. But if someone has some specific idea about the number of required connections then they can use this property instead?

src/Microsoft.Azure.SignalR.Management/Configuration/ContextOptions.cs
wanlwanl: `ServiceManager` => ` instance of <see cref="IServiceManager"/>`

src/Microsoft.Azure.SignalR.Management/Configuration/ContextOptions.cs
Y-Sindo: @vwxyzh has a plan to introduce a mechanism to automatically manager connection count. It seems OK to just give it a value before that mechanism is implemented.

KKhurin: Just a question - after this automatic mechanism is implemented, does this value override the automatically selected one? Sets the higher limit? Sets the lower limit? If so, maybe the name should imply that the actual number of connections could be different?

Y-Sindo: I guess `ConnectionCount` will be ignored after the automatic mechanism is implemented. Now the `ConnectionCount` is internal and won't confuse users. As we still use it internally, I think it's better to stay consistent with the name in server SDK and `IServiceManager`?

vicancy (approved):

PR #1138, opened 2020-12-04
remove useless protocol serialization for a certain connection.

vwxyzh:
reduce the message size for sending to a certain client connection by removing useless bytes for other protocol.

### DIFF
--- a/src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManager.cs
+++ b/src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManager.cs
@@ -63,7 +63,7 @@ namespace Microsoft.Azure.SignalR
 
             if (_clientConnectionManager.ClientConnections.TryGetValue(connectionId, out var serviceConnectionContext))
             {
-                var message = CreateMesage(connectionId, methodName, args, serviceConnectionContext);
+                var message = CreateMessage(connectionId, methodName, args, serviceConnectionContext);
                 var messageWithTracingId = (IMessageWithTracingId)message;
                 try
                 {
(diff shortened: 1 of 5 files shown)


src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManager.cs
```diff
@@ -63,27 +63,55 @@ public override async Task SendConnectionAsync(string connectionId, string metho
...
-                    MessageLog.StartToSendMessageToConnections(Logger, message);
-                }
-
-                try
-                {
-                    // Write directly to this connection
-                    await serviceConnectionContext.ServiceConnection.WriteAsync(message);
-                    
+                    var message = new ConnectionDataMessage(connectionId, SerializeProtocol(serviceConnectionContext.Protocol, methodName, args)).WithTracingId();
                     if (message.TracingId != null)
```
vicancy, line 69: only this line needs to be inside the `if` check?

src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManager.cs
```diff
@@ -63,27 +63,55 @@ public override async Task SendConnectionAsync(string connectionId, string metho
...
                     {
-                        MessageLog.SucceededToSendMessage(Logger, message);
+                        MessageLog.StartToSendMessageToConnection(Logger, message);
+                    }
```
vicancy, line 72: looks like we only need to `if` check these lines?

vicancy (approved):
:shipit:

vicancy (approved):

PR #1141, opened 2020-12-08
Various livescale-related fixes

KKhurin:
### Various fixes to unblock the additional message order tests
 - fixing Start/Stop for CustomizedPingTimer
 - fixing container stop logic (won't try restarting or accept on demand connections after stop or offline)
 - misc fixes

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
+++ b/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
@@ -50,6 +50,8 @@ namespace Microsoft.Azure.SignalR
 
         private volatile ServiceConnectionStatus _status;
 
+        private bool _isOffline = false;
+
 
         // <serversTag, latestTimestamp>
         private volatile Tuple<string, long> _serversTagContext = DefaultServersTagContext;
@@ -98,6 +100,8 @@ namespace Microsoft.Azure.SignalR
                 }
             }
         }
+
+        protected bool ReadyForNewConnections => !_isOffline && !_terminated;
 
         protected ServiceConnectionContainerBase(IServiceConnectionFactory serviceConnectionFactory,
                                                  int minConnectionCount,
@@ -171,7 +175,7 @@ namespace Microsoft.Azure.SignalR
         /// <returns></returns>
         protected async Task StartCoreAsync(IServiceConnection connection, string target = null)
         {
-            if (_terminated)
+            if (!ReadyForNewConnections)
             {
                 return;
             }
@@ -255,6 +259,7 @@ namespace Microsoft.Azure.SignalR
 
         public virtual Task OfflineAsync(GracefulShutdownMode mode)
         {
+            _isOffline = true;
             return Task.WhenAll(ServiceConnections.Select(c => RemoveConnectionAsync(c, mode)));
         }
 
@@ -304,12 +309,17 @@ namespace Microsoft.Azure.SignalR
             serviceConnection.ConnectionStatusChanged -= OnConnectionStatusChanged;
 
             var index = ServiceConnections.IndexOf(serviceConnection);
+
             if (index != -1)
             {
-                // first FixedConnectionCount connections are "fixed" and always try to restart
+                // always try to restart first FixedConnectionCount connections
                 if (index < FixedConnectionCount)
                 {
-                    await RestartFixedServiceConnectionCoreAsync(index);
+                    // unless the container is offline or stopped
+                    if (ReadyForNewConnections)
+                    {
+                        await RestartFixedServiceConnectionCoreAsync(index);
+                    }
                 }
                 // the rest are "on demand" and are only created upon request
                 else
@@ -570,16 +580,15 @@ namespace Microsoft.Azure.SignalR
                 _dueTime = dueTime;
                 _intervalTime = intervalTime;
                 _defaultPingTicks = intervalTime.Seconds * Stopwatch.Frequency;
-
-                _timer = Init();
             }
 
             public bool Start()
             {
                 if (Interlocked.Increment(ref _counter) == 1)
                 {
-                    _timer.Start();
-                    _ = PingAsync(_timer);
+                    var timer = _timer = Init();
+                    timer.Start();
+                    _ = PingAsync(timer);
                     return true;
                 }
                 return false;
@@ -598,14 +607,15 @@ namespace Microsoft.Azure.SignalR
                     }
                     if (Interlocked.Decrement(ref _counter) == 0)
                     {
-                        _timer.Stop();
+                        _timer?.Stop();
+                        _timer = null;
                     }
                 }
             }
 
             public void Dispose()
             {
-                _timer.Stop();
+                _timer?.Stop();
             }
 
             private TimerAwaitable Init()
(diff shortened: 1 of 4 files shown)


src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -598,14 +607,15 @@ public void Stop()
                     }
                     if (Interlocked.Decrement(ref _counter) == 0)
                     {
-                        _timer.Stop();
+                        _timer?.Stop();
```
JialinXin, line 610: `_counter` already ensure `_timer = Init()`, no need to check null here?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -598,14 +607,15 @@ public void Stop()
...
+                        _timer = null;
                     }
                 }
             }
 
             public void Dispose()
             {
-                _timer.Stop();
+                _timer?.Stop();
```
JialinXin, line 618: also set null after stop?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -50,6 +50,8 @@ internal abstract class ServiceConnectionContainerBase : IServiceConnectionConta
 
         private volatile ServiceConnectionStatus _status;
 
+        private bool _isOffline = false;
```
JialinXin, line 53: Shall we use single flag to make this status check? Like merge _isOffline and _terminated? Looks both terminate all connections and hold on for new ones.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
KKhurin: yeah, pretty much being too defensive..

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
KKhurin: if we null it out here, then we should keep the null check in Stop() to avoid race condition between Dispose and Stop, right?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
KKhurin: They indeed serve the same purpose but at different stages of shutdown. Perhaps the only reason leave them separate is for debugging purpose? Having one flag is cleaner though.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
JialinXin: I'm wondering whether the race condition between Dispose and Stop could happen. Just notice TimeAwaitable has a `Dispose` method, or shall we call `_timer.Dispose()` here?

KKhurin: From looking at the `TimerAwaitable` implementation, it seems the proper way is to call `_timer.Stop()` and only then dispose it. The `using(timer)` I added inside `PingAsync` does just that. Thus calling `_timer.Stop` in `CustomizedPingTimer.Dispose` is following the same pattern. Although if nobody calls `CustomizedPingTimer.Start` then we'd never call `_timer.Dispose`... And calling Dispose more than once doesn't have any adverse effects. However calling `Dispose` and then `Stop` on the timer does throw...

Which brings me to the race between `Dispose` and then `Stop`. I am also not 100% sure what scenario will cause this. As far as I am aware nobody right now is ever disposing `MultiEndpointServiceConnectionContainer` instances (this is the only other route of getting the containers disposed). But it seems like a good practice to make Dispose as safe to call as possible.

KKhurin: > I'm wondering whether the race condition between Dispose and Stop could happen. Just notice TimeAwaitable has a `Dispose` method, or shall we call `_timer.Dispose()` here?

BTW, it is explicitly implemented as `void IDisposable.Dispose()` which implies that "nobody _should_ ever call it, other than from using()". From further looking into its code it seems that disposing it is not necessary if nobody called `Start`.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -620,22 +630,25 @@ private TimerAwaitable Init()
 
             private async Task PingAsync(TimerAwaitable timer)
             {
-                while (await timer)
+                using (timer)
```
JialinXin, line 633: `using` timer will lead to dispose error, that when finally remove endpoint, we'll dispose the container and call `Container.Dispose()`, which will call timer.Dispose() and will throw stop a disposed item.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
JialinXin: Yes, but currently we split the Start/Stop operation and not able to handle cleanup by a single using. How about call `((IDisposable)_timer).Dispose();`?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -304,12 +309,17 @@ protected virtual async Task OnConnectionComplete(IServiceConnection serviceConn
...
+
             if (index != -1)
             {
-                // first FixedConnectionCount connections are "fixed" and always try to restart
+                // always try to restart first FixedConnectionCount connections
                 if (index < FixedConnectionCount)
                 {
-                    await RestartFixedServiceConnectionCoreAsync(index);
+                    // unless the container is offline or stopped
+                    if (ReadyForNewConnections)
```
JialinXin, line 319: too much `if`, how about `if (index < FixedConnectionCount && !_terminated)`

src/Microsoft.Azure.SignalR.Common/ServiceConnections/StrongServiceConnectionContainer.cs
```diff
@@ -20,11 +20,14 @@ internal class StrongServiceConnectionContainer : ServiceConnectionContainerBase
         public override async Task HandlePingAsync(PingMessage pingMessage)
         {
             await base.HandlePingAsync(pingMessage);
-            if (RuntimeServicePingMessage.TryGetRebalance(pingMessage, out var target) && !string.IsNullOrEmpty(target))
+            if (ReadyForNewConnections)
```
JialinXin, line 23: instead of adding the check here, we should let RT not sending rebalance request when it's going offline.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
KKhurin: Yeah, initially I did exactly what you proposed. This (in case of `_terminated` == true ) will result in calling `RemoveOnDemandConnection` in else statement and eventually emptying ServiceConnections. This is ok, but I was worrying about our usage of [Task.WhenAny on ServiceConnections](https://github.com/Azure/azure-signalr/blob/dev/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs#L222) which will throw when the collection is finally empty. Even though right now I don't see a scenario when someone would check `ConnectionInitializedTask` during shutdown stage I thought I will save the potential trouble and keep the old behavior of not removing stopped connections from `ServiceConnections`.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/StrongServiceConnectionContainer.cs
KKhurin: Should be both - SDK not creating new connections when in shutting down stage and RT not sending rebalance requests. Since there are inevitable (and sometimes unpredictable) delays of incoming messages the actual rebalance ping could have been sent before the shutdown started.

KKhurin:
I split these changes to separate PRs where I have addressed some of the feedback. 
Closing this one.

Issue #1152, opened 2020-12-19
Method WriteMessageAsync return Request instead of Response


mikhail-khalizev:
This line returns request:

https://github.com/Azure/azure-signalr/blob/b5a82860e9af41d23f582a6c87522fb30318596f/src/Microsoft.Azure.SignalR.Emulator/HubEmulator/HttpServerlessMessageHandler.cs#L63

There should be something like:

```c#
return new ReadOnlySequence<byte>(await response.Content.ReadAsByteArrayAsync());
```

PR #1153, opened 2020-12-21
Fix upstream response and add size check

vicancy:
Fix https://github.com/Azure/azure-signalr/issues/1152

### DIFF
--- a/src/Microsoft.Azure.SignalR.Emulator/HubEmulator/HttpServerlessMessageHandler.cs
+++ b/src/Microsoft.Azure.SignalR.Emulator/HubEmulator/HttpServerlessMessageHandler.cs
@@ -23,6 +23,8 @@ namespace Microsoft.Azure.SignalR.Emulator.HubEmulator
 {
     internal class HttpServerlessMessageHandler<THub> : IUpstreamMessageHandler where THub: Hub
     {
+        // We don't support response large than 16M
+        private const int MaxAllowedResponseLength = 16 * 1024 * 1024;
         private static readonly byte[] OpenConnectionPayload = Encoding.UTF8.GetBytes(JsonConvert.SerializeObject(
             new ServerlessProtocol.OpenConnectionMessage { Type = ServerlessProtocol.Constants.OpenConnectionMessageType }
         ));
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.SignalR.Emulator/HubEmulator/HttpServerlessMessageHandler.cs
```diff
@@ -60,7 +60,37 @@ public async Task<ReadOnlySequence<byte>> WriteMessageAsync(HubConnectionContext
...
+                        {
+                            // same as no content.
+                            await connectionContext.WriteAsync(CompletionMessage.Empty(message.InvocationId));
+                        }
+                        else if (contentLength > 16 * 1024 * 1024)
+                        {
+                            // We don't support response large than 16M, fast fail.
+                            await connectionContext.WriteAsync(CompletionMessage.WithError(message.InvocationId, $"Invocation failed, response too large."));
+                        }
+                        else
```
vwxyzh, line 74: what about `null`?

vicancy: which is null?

vicancy: Added LimitedStream

src/Microsoft.Azure.SignalR.Emulator/HubEmulator/HttpServerlessMessageHandler.cs
```diff
@@ -60,7 +61,38 @@ public async Task<ReadOnlySequence<byte>> WriteMessageAsync(HubConnectionContext
...
                     if (response.StatusCode == HttpStatusCode.OK)
                     {
-                        return message.Payload;
+                        var contentLength = response.Content.Headers.ContentLength;
+                        if (contentLength == 0)
+                        {
+                            // same as no content.
+                            await connectionContext.WriteAsync(CompletionMessage.Empty(message.InvocationId));
+                        }
+                        else if (contentLength > 16 * 1024 * 1024)
```
JialinXin, line 70: >16 * 1024 * 1024 [](start = 49, length = 16)

make this max length constants?

JialinXin (approved):
:shipit:

PR #1155, opened 2020-12-25
fix IConfigurationExtensions bug

Y-Sindo:
* fix `IConfigurationExtensions` bug which fail to get all endpoints from a section

* better naming


src/Microsoft.Azure.SignalR.Common/Endpoints/IConfigurationExtension.cs
```diff
@@ -15,14 +15,9 @@ internal static class IConfigurationExtension
         /// <remarks>
         /// The SignalR service endpoint whose key is exactly the section name is not extracted.
         /// </remarks>
-        public static ServiceEndpoint[] GetSignalRServiceEndpoints(this IConfiguration configuration, string sectionName)
+        public static IEnumerable<ServiceEndpoint> GetSignalREndpointsFromSectionChildren(this IConfiguration configuration, string sectionName)
```
JialinXin, line 18: Simplify method name and it's confusing using same input/output but different method names. Suggest to reference previous way: https://github.com/Azure/azure-signalr/blob/v1.6.0/src/Microsoft.Azure.SignalR/ServiceOptionsSetup.cs#L60-L81

Y-Sindo: Sometimes we want all the endpoints as a `IEnumerable`, if using `(string, ServiceEndoint[])`, then we have to merge the result with one more step. How about simplify the method names to `GetAllEndpoints` and `GetSuffixedEndpoints`? "SuffixedEndpoints" means endpoints whose keys have suffixes?

JialinXin: Should they handle together as when loading the settings we don't distinguish them? It's internal logic to detect suffixed list or single connection string. Return `(string, IEnumerable<ServiceEndpoint>)` to save the merge?

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpointUtility.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+
+namespace Microsoft.Azure.SignalR.Common
```
JialinXin, line 6: >Microsoft.Azure.SignalR [](start = 10, length = 23)

use namespace `Microsoft.Azure.SignalR`.

src/Microsoft.Azure.SignalR.Common/Endpoints/IConfigurationExtension.cs
```diff
@@ -3,51 +3,25 @@
...
-                yield return new ServiceEndpoint(connectionString);
-            }
-
-            var section = configuration.GetSection(sectionName);
-            GetEndpoints(section);
+            var suffixedEndpoints = section.AsEnumerable(true)
+                                           .Where(entry => !string.IsNullOrEmpty(entry.Value))
+                                           .Select(entry => new ServiceEndpoint(entry.Key, entry.Value));
+            var connectionString = includeNoSuffixEndpoint ? configuration[sectionName] : null;
+            return ServiceEndpointUtility.Merge(connectionString, suffixedEndpoints);
```
JialinXin, line 24: >return ServiceEndpointUtility.Merge(connectionString, suffixedEndpoints); [](start = 12, length = 73)

return `includeNoSuffixEndpoint ? ServiceEndpointUtility.Merge(configuration[sectionName], suffixedEndpoints) : suffixedEndpoints;` ?

JialinXin (approved):

PR #1157, opened 2020-12-27
Prevent service connection auto-restart after container shutdown

KKhurin:
### Summary 
 - prevent auto-restarting service connections after container shutdown
 - update unit test


JialinXin (approved):
:shipit:

PR #1167, opened 2021-01-05
Message order fix and tests

KKhurin:
### Guaranteed message order fix and tests for mainstream scenarios
 - added extensive tests for the message order for various scenarios
 - addressed feedback for previous PR #1040
 - fixed memleak in hot reload scenario and added targeted test for it

Note: the current fix does not address some of the corner cases such as guaranteed message order in the face of concurrent writes. Even though the guarantee the SDK can provide in this case is already fairly weak (as it all depends on the order of the messages as they arrive from concurrent threads) the remaining corner cases in the SDK code might introduce additional inconsistencies (such as using several service connections for a short period of time). These corner case issues will be resolved in future fixes.


test/Microsoft.Azure.SignalR.IntegrationTests/Infrastructure/MessageOrderTestHub.cs
```diff
@@ -0,0 +1,199 @@
...
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.SignalR;
+
+namespace Microsoft.Azure.SignalR.IntegrationTests.Infrastructure
+{
+    // Each test uses a separate hub
+    public class UseSameServiceConnectionHub : Hub
```
vicancy, line 12: as the convention, per class per file

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -472,32 +472,61 @@ private void OnConnectionStatusChanged(StatusChange obj)
             }
         }
 
-
-
-        private Task WriteToRandomAvailableConnection(ServiceMessage serviceMessage)
+        private async Task WriteToScopedOrRandomAvailableConnection(ServiceMessage serviceMessage)
```
vicancy, line 475: the product code LGTM

vicancy (approved):

PR #1169, opened 2021-01-06
only wrap some timeout or cancelled error for negotiate error

Y-Sindo:


JialinXin (approved):

src/Microsoft.Azure.SignalR.Management/NegotiateProcessor.cs
```diff
@@ -54,7 +54,7 @@ public async Task<NegotiationResponse> NegotiateAsync(string hubName, HttpContex
                     AccessToken = tokenTask.Result
                 };
             }
-            catch (Exception e)
+            catch (Exception e) when (e is TaskCanceledException || e is TimeoutException)
```
vicancy, line 57: use `OperationCanceledException` instead as a general practice

vicancy (approved):

PR #1171, opened 2021-01-06
Make endpoint disposable

terencefan:
 - Detail 1
 - Detail 2

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpointManagerBase.cs
+++ b/src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpointManagerBase.cs
@@ -230,6 +230,9 @@ namespace Microsoft.Azure.SignalR
 
             // Set complete
             endpoint.CompleteScale();
+
+            // Dispose endpoint
+            endpoint.Dispose();
         }
 
         private void UpdateEndpoints(Dictionary<ServiceEndpoint, ServiceEndpoint> updatedEndpoints,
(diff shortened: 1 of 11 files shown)


src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpointManagerBase.cs
```diff
@@ -230,6 +230,9 @@ private async Task RemoveHubServiceEndpointAsync(HubServiceEndpoint endpoint, Ca
 
             // Set complete
             endpoint.CompleteScale();
+
+            // Dispose endpoint
```
Y-Sindo, line 234: As endpoints are shared between hubs, will it matter when an endpoint is disposed in one hub, while requests come in another hub?

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpointManagerBase.cs
```diff
@@ -230,6 +230,9 @@ private async Task RemoveHubServiceEndpointAsync(HubServiceEndpoint endpoint, Ca
...
+            endpoint.Dispose();
```
vicancy, line 235: wrap into try..finally.. block?

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpointManagerBase.cs
Y-Sindo: How about sharing a global timer between AAD access key?

PR #1172, opened 2021-01-07
Ensure service hub context created from service manager independent f…

Y-Sindo:
…rom each other

 - Use separate service collection for each hub again
 - Change the `ServiceHubContext` creation for `IServiceContext` to not implemented, as it needs redesign.

### Some considerations
1. If `MultiendpointServiceConnectionContainer` is reused, then `ServiceHubContext` s with same hub name are not independent from each other. If one of them is disposed, then other won't work. 
2. What's more, if `IServiceEndpointManager` is an singleton, then multiple `ServiceHubContext` s with same hub name also have problems, as `IServiceEndpointManager` allows only one `IServiceConnectionContainer` for each `HubServiceEndpoint`.

### DIFF
--- a/src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/ServiceHubContextFactory.cs
+++ b/src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/ServiceHubContextFactory.cs
@@ -1,33 +1,35 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.SignalR;
 using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Azure.SignalR.Management
 {
     internal class ServiceHubContextFactory
     {
         private readonly ServiceHubLifetimeManagerFactory _managerFactory;
+        private readonly IServiceProvider _serviceProvider;
 
-        public ServiceHubContextFactory(ServiceHubLifetimeManagerFactory managerFactory)
+        public ServiceHubContextFactory(ServiceHubLifetimeManagerFactory managerFactory,IServiceProvider serviceProvider)
         {
             _managerFactory = managerFactory;
+            _serviceProvider = serviceProvider;
         }
 
-        public async Task<IServiceHubContext> CreateAsync(string hubName, ILoggerFactory loggerFactory = null, CancellationToken cancellationToken = default)
+        public Task<IServiceHubContext> CreateAsync(string hubName)
         {
-            var manager = await _managerFactory.CreateAsync(hubName, cancellationToken, loggerFactory);
+            var manager = _managerFactory.Create(hubName);
             var servicesPerHub = new ServiceCollection();
             servicesPerHub.AddSignalRCore();
             servicesPerHub.AddSingleton((HubLifetimeManager<Hub>)manager);
             var serviceProviderPerHub = servicesPerHub.BuildServiceProvider();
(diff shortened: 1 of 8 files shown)


src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/ServiceHubContextFactory.cs
```diff
@@ -1,33 +1,35 @@
...
+using System;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.SignalR;
 using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Azure.SignalR.Management
 {
     internal class ServiceHubContextFactory
```
Y-Sindo, line 12: This class is not needed anymore. To minimize changes in this PR, this class is not removed at this time.

src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/ServiceHubLifetimeManagerFactory.cs
```diff
@@ -4,40 +4,33 @@
...
-        public ServiceHubLifetimeManagerFactory(IServiceProvider sp, IOptions<ContextOptions> context, MultiEndpointConnectionContainerFactory connectionContainerFactory)
+        public ServiceHubLifetimeManagerFactory(IServiceProvider sp, IOptions<ContextOptions> context)
         {
             _serviceProvider = sp;
-            _connectionContainerFactory = connectionContainerFactory;
             _context = context.Value;
         }
 
-        public async Task<IServiceHubLifetimeManager> CreateAsync(string hubName, CancellationToken cancellationToken, ILoggerFactory loggerFactoryPerHub = null)
+        public  IServiceHubLifetimeManager Create(string hubName)
```
vicancy, line 24: extra space

vicancy (approved):

src/Microsoft.Azure.SignalR.Management/Obsolete/ServiceManager.cs
```diff
@@ -16,22 +19,39 @@ namespace Microsoft.Azure.SignalR.Management
...
         private readonly IServiceEndpointProvider _endpointProvider;
 
-        public ServiceManager(RestClientFactory restClientFactory, ServiceHubContextFactory serviceHubContextFactory, IServiceEndpointManager endpointManager, IServiceProvider serviceProvider)
+        public ServiceManager(IServiceCollection serviceDescriptors)
         {
-            _restClientFactory = restClientFactory;
-            _serviceHubContextFactory = serviceHubContextFactory;
-            _serviceProvider = serviceProvider;
+            _services = serviceDescriptors;
+            _serviceProvider = _services.BuildServiceProvider(); ;
```
wanlwanl, line 30: `; ;`

src/Microsoft.Azure.SignalR.Management/Obsolete/ServiceManager.cs
```diff
@@ -16,22 +19,39 @@ namespace Microsoft.Azure.SignalR.Management
...
+            _restClientFactory = _serviceProvider.GetService<RestClientFactory>();
```
wanlwanl, line 31: is it possible to be null? how about GetRequiredService`
Same to next line

Y-Sindo: Just think that if null, then error will be thrown when we need them. We don't always need `RestClientFactory` during some test scenarios.

src/Microsoft.Azure.SignalR.Management/Obsolete/ServiceManager.cs
Y-Sindo: Might be fixed in another pr.

PR #1175, opened 2021-01-08
add diagnostic client claim to management sdk

wanlwanl:


src/Microsoft.Azure.SignalR.Management/ServiceContext.cs
```diff
@@ -31,9 +31,19 @@ public Task<IServiceHubContext> CreateHubContextAsync(string hubName, Cancellati
...
+            if (claims != null)
+            {
+                customClaims.AddRange(claims);
+            }
+            if (isDiagnosticClient)
+            {
+                customClaims.Add(new Claim(Constants.ClaimType.DiagnosticClient, "true"));
+            }
+
+            return _negotiateProcessor.GetClientEndpointAsync(hubName, httpContext, userId, customClaims, lifetime, cancellationToken);
```
JialinXin, line 46: Looks you can pass in `isDiagnosticClient` where eventually call `ClaimsUtility.BuildJwtClaims` that respect this parameter.

wanlwanl: you are right...

JialinXin (approved):

PR #1182, opened 2021-01-12
Enable negotiation for multiple endpoints inside ServiceHubContext

Y-Sindo:
* Allow connectionstring to be null when service endpoints not null.
* Remove unused code.
* Move `IServiceManager` ... files out of obsolete directory.


src/Microsoft.Azure.SignalR.Management/IInternalServiceHubContext.cs
```diff
@@ -11,29 +11,20 @@
...
         /// <param name="lifetime">The lifetime of the token. The default value is one hour.</param>
         /// <param name="isDiagnosticClient">The flag whether the client to be connected is a diagnostic client.</param>
         /// <param name="cancellationToken">Cancellation token for aborting the operation. If null, the <see cref="HttpContext.RequestAborted"/> of <paramref name="httpContext"/> will be used. </param>
         /// <returns>Client endpoint and access token to Azure SignalR Service.</returns>
-        Task<NegotiationResponse> NegotiateAsync(string hubName, HttpContext httpContext = null, string userId = null, IList<Claim> claims = null, TimeSpan? lifetime = null, bool isDiagnosticClient = false, CancellationToken cancellationToken = default);
+        Task<NegotiationResponse> NegotiateAsync(HttpContext httpContext = null, string userId = null, IList<Claim> claims = null, TimeSpan? lifetime = null,bool isDiagnosticClient = false, CancellationToken cancellationToken = default);
```
wanlwanl, line 26: `null,bool` space

src/Microsoft.Azure.SignalR.Management/IInternalServiceHubContext.cs
```diff
@@ -11,29 +11,20 @@
 
 namespace Microsoft.Azure.SignalR.Management
 {
-    public interface IServiceContext : IDisposable
+    internal interface IInternalServiceHubContext : IServiceHubContext
```
wanlwanl, line 14: you keep it here, still has a plan to use it in the future?

src/Microsoft.Azure.SignalR.Management/Obsolete/ServiceManager.cs
```diff
@@ -19,6 +18,7 @@ namespace Microsoft.Azure.SignalR.Management
 {
     internal class ServiceManager : IServiceManager
     {
+        private const string EmptyConnectionStringMsg = "Connection string is null or empty or only contains whitespace.";
```
wanlwanl, line 21: Msg => Message

src/Microsoft.Azure.SignalR.Management/Obsolete/ServiceManager.cs
```diff
@@ -75,11 +83,16 @@ public string GenerateClientAccessToken(string hubName, string userId = null, IL
 
         public string GetClientEndpoint(string hubName)
         {
-            return _endpointProvider.GetClientEndpoint(hubName, null, null);
+            return _endpointProvider == null ? throw new InvalidOperationException
+                (EmptyConnectionStringMsg) : _endpointProvider.GetClientEndpoint(hubName, null, null);
```
wanlwanl, line 87: _endpointProvider?.GetClientEndpoint() ?? throw new InvalidOperationException(EmptyConnectionStringMsg)

src/Microsoft.Azure.SignalR.Management/Obsolete/ServiceManager.cs
```diff
@@ -31,8 +31,12 @@ public ServiceManager(IServiceCollection serviceDescriptors)
             _serviceProvider = _services.BuildServiceProvider(); ;
             _restClientFactory = _serviceProvider.GetRequiredService<RestClientFactory>();
             var endpointManager = _serviceProvider.GetRequiredService<IServiceEndpointManager>();
-            _endpoint = new ServiceEndpoint(_serviceProvider.GetRequiredService<IOptions<ServiceManagerOptions>>().Value.ConnectionString);
-            _endpointProvider = endpointManager.GetEndpointProvider(_endpoint);
+            var connectionString = _serviceProvider.GetRequiredService<IOptions<ServiceManagerOptions>>().Value.ConnectionString;
+            if (!string.IsNullOrWhiteSpace(connectionString))
+            {
+                _endpoint = new ServiceEndpoint(connectionString);
+                _endpointProvider = endpointManager.GetEndpointProvider(_endpoint);
```
wanlwanl, line 38: how about check null here for _endpoint  and _endpointProvider ? then no need to check later

src/Microsoft.Azure.SignalR.Management/IInternalServiceHubContext.cs
Y-Sindo: Emmm, actually it is a newly created file, but git associates it with an old deleted file. The new space is added by formatter. I think it's better to keep it, otherwise we have to deal with the space every time we format the file.

src/Microsoft.Azure.SignalR.Management/IInternalServiceHubContext.cs
Y-Sindo: Actually it is a new file. It is different from `IServiceContext`. Use it to avoid add new internal methods to `IServiceHubContext` and disable users to implement `IServiceHubContext` for test. Some methods might be moved to `IServiceHubContext` and made public if ready.

wanlwanl: why IInternalService**Hub**Context  has method `NegotiateAsync` and `GetServiceEndpoints`? what's the plan?

src/Microsoft.Azure.SignalR.Management/Obsolete/ServiceManager.cs
Y-Sindo: If check null here, then `ConnectionString = null && ServiceEndpoints!=null` will be invalid, which should be valid.

wanlwanl: is ServiceEndpoints use here?

Y-Sindo: No

wanlwanl (approved):

PR #1185, opened 2021-01-14
Create service manager from service provder instead of new

Y-Sindo:
To get the service provider in function binding

### DIFF
--- a/src/Microsoft.Azure.SignalR.Management/DependencyInjectionExtensions.cs
+++ b/src/Microsoft.Azure.SignalR.Management/DependencyInjectionExtensions.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System.Collections.Generic;
 using System.ComponentModel;
+using System.Linq;
 using System.Net.Http;
 using System.Reflection;
 using System.Threading.Tasks;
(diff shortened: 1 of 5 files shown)


src/Microsoft.Azure.SignalR.Management/DependencyInjectionExtensions.cs
```diff
@@ -40,6 +42,7 @@ public static IServiceCollection AddSignalRServiceManager(this IServiceCollectio
 
         private static IServiceCollection AddSignalRServiceCore(this IServiceCollection services)
         {
+            services.AddSingleton<IServiceManager, ServiceManager>();
```
wanlwanl, line 45: looks like it's all about mng sdk. merge to AddSignalRServiceManager?

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
```diff
@@ -63,8 +65,10 @@ public ServiceManagerBuilder WithCallingAssembly()
         /// <returns>The instance of the <see cref="IServiceManager"/>.</returns>
         public IServiceManager Build()
         {
-            _services.AddSignalRServiceManager();
-            return new ServiceManager(_services);
+            return _services.AddSignalRServiceManager()
+                .AddSingleton(_services.ToList() as IReadOnlyCollection<ServiceDescriptor>)
```
wanlwanl, line 69: since ServiceManager is useless for multi-endpoint scenario, how about exclude ServiceManager, and use a new builder, then the process to build the context should be simple
 e.g. 
``` c#
var context = MultiEndpointHubContextBuilder.WithConfiguration(params).Build();
// start
await context.StartAsync();
// negotiate
var res = await context.NegotiateAsync(params);
// send message
await context.Clients.X.SendAsync(params);
// stop
await context.StopAsync();
```

wanlwanl: @vicancy @vwxyzh what do you think?
I think the service manger is totally useless in muti-endpoint scenario. so directly build a hub context is better

wanlwanl (approved):

PR #1187, opened 2021-01-15
Allow creating service hub context with a subset of service endpoints

Y-Sindo:
Design for function binding


test/Microsoft.Azure.SignalR.E2ETests/Management/ServiceHubContextE2EFacts.cs
```diff
@@ -294,9 +294,57 @@ internal async Task ConnectionJoinLeaveGroupTest(ServiceTransportType serviceTra
...
+                for (var i = 0; i < endpoints.Length; i++)
+                {
+                    var j = i;
+                    connections[j].On(MethodName, (string message) =>
+                    {
+                        receivedFlags[j] = true;
+                    });
+                }
+
+                var subHubContext = (hubContext as IInternalServiceHubContext).WithEndpoints(endpoints.Take(1));
```
wanlwanl, line 332: looks like `WithEnpoint()` equals to build a new serviceManager then create hub context.
Since it's internal in asrs function extension, why not simply do that?

Y-Sindo: It is not equal to build a new ServiceManager, because it doesn't create new connections.

wanlwanl: Then you limit the endpoints has to be the subset of the endpoints.
For the old version, in binding attribute, cx can set any connection string key no matter which one is for glabal usage.
I think we can keep it too.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
```diff
@@ -31,7 +31,7 @@ internal class MultiEndpointServiceConnectionContainer : IMultiEndpointServiceCo
             IServiceEndpointManager endpointManager,
             IMessageRouter router,
             ILoggerFactory loggerFactory,
-            TimeSpan? scaleTimeout = null)
+            TimeSpan? scaleTimeout = null) : base(loggerFactory)
```
vicancy, line 34: is it enough to have another ctor for `MultiEndpointServiceConnectionContainer` that supports `endpoints` parameter?

Y-Sindo: No, `MultiEndpointServiceConnectionContainer` will dispose its connection containers while we don't want them to be disposed.

wanlwanl: @vicancy I prefer inherit multiendpointcontainer in function extension side, and override the getRountedEndpoints method, since the management sdk is already internal visible to function extension side. it minimal the changes in server sdk, and I don't see any meaning to public withEndpoint method, internal only. the new DirectMultiEnpointContainer similar to view in database, instead of a concret table (connection container here)

src/Microsoft.Azure.SignalR.Common/ServiceConnections/RawMultiEndpointServiceConnectionContainer.cs
```diff
@@ -0,0 +1,167 @@
...
+{
+    /// <summary>
+    /// A service connection container which sends message to multiple service endpoints.
+    /// </summary>
+    internal class RawMultiEndpointServiceConnectionContainer : IMultiEndpointServiceConnectionContainer
+    {
+        private readonly ILogger _logger;
+        internal IEnumerable<HubServiceEndpoint> TargetEndpoints { get; }
+
+        public RawMultiEndpointServiceConnectionContainer(IEnumerable<HubServiceEndpoint> targetEndpoints, ILoggerFactory loggerFactory)
```
vicancy, line 23: is `IReadOnlyCollection` better indicating the input is immutable?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/RawMultiEndpointServiceConnectionContainer.cs
```diff
@@ -0,0 +1,167 @@
...
+
+            if (routed.Length == 1)
+            {
+                return routed[0];
+            }
+
+            return Task.WhenAll(routed);
+        }
+
+        public Task StartAsync() => Task.CompletedTask;
```
vicancy, line 103: I feel implementing `StartAsync` and `StopAsync` makes it a real container, any reason not have it?

vicancy: Is it just some multiple endpoint message writer?

Y-Sindo: Yes, it is just some multiple endpoint message writer

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointMessageWriter.cs
```diff
@@ -0,0 +1,167 @@
...
+{
+    /// <summary>
+    /// A service connection container which sends message to multiple service endpoints.
+    /// </summary>
+    internal class MultiEndpointMessageWriter : IMultiEndpointServiceConnectionContainer
+    {
+        private readonly ILogger _logger;
+        internal IReadOnlyCollection<HubServiceEndpoint> TargetEndpoints { get; }
+
+        public MultiEndpointMessageWriter(IEnumerable<HubServiceEndpoint> targetEndpoints, ILoggerFactory loggerFactory)
```
vicancy, line 23: parameter to use `IReadOnlyCollection` instead?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointMessageWriter.cs
```diff
@@ -0,0 +1,167 @@
...
+        {
+            TargetEndpoints = targetEndpoints.ToList();
+            _logger = loggerFactory.CreateLogger<MultiEndpointMessageWriter>();
+        }
+
+        public Task ConnectionInitializedTask => Task.WhenAll(from endpoint in TargetEndpoints select endpoint.ConnectionContainer.ConnectionInitializedTask);
```
vicancy, line 29: `endpoint` can be null

src/Microsoft.Azure.SignalR.Management/ServiceHubContext.cs
```diff
@@ -51,5 +55,27 @@ public async Task DisposeAsync()
             await _lifetimeManager.DisposeAsync();
             (ServiceProvider as IDisposable)?.Dispose();
         }
+
+        IInternalServiceHubContext IInternalServiceHubContext.WithEndpoints(IEnumerable<ServiceEndpoint> endpoints)
+        {
+            var targetEndpoints = _endpointManager.GetEndpoints(_hubName).Intersect(endpoints, EqualityComparer<ServiceEndpoint>.Default).Select(e => e as HubServiceEndpoint).ToList();
```
vicancy, line 61: endpoint is not neccessorily a `HubServiceEndpoint`, actually user can never create a `HubServiceEndpoint`, is `WithEndpoint` to be exposed to the customer?

Y-Sindo: The parameter 'endpoints' doesn't need to be `HubServiceEndpoint`. In the `ServiceHubContextFacts`, a test to demonstrate such situation where `ServiceEndpoint` is passed. As `_endpointManager.GetEndpoints(_hubName)` returns `IEnumerable<HubServiceEndpoint`, the intersect result will be `IEnumerable<HubServiceEndpoint>`. The `EqualityComparer` is `EqualityComparer<ServiceEndpoint>.Default`, which will use the `ServiceEndpoint.Equals(...)` and `ServiceEndpoints.GetHashCode()` to comparer.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointMessageWriter.cs
Y-Sindo: I think `IEnumerable` as parameter type would reduce the caller effort. No matter what type the parameter is, the ctor takes the responsibility to convert it to read only collection.

vicancy: So `IReadOnlyCollection` as the parameter indicates the caller that the endpoints are immutable. And the caller decides when to evaluate it to an immutable list.

src/Microsoft.Azure.SignalR.Management/ServiceHubContext.cs
vicancy: And `e as HubServiceEndpoint` will be always null for them?

Y-Sindo: No, it always works. ` _endpointManager.GetEndpoints(_hubName).Intersect(...)`   returns endpoints in endpoint manager.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointMessageWriter.cs
```diff
@@ -0,0 +1,167 @@
...
+    /// A service connection container which sends message to multiple service endpoints.
+    /// </summary>
+    internal class MultiEndpointMessageWriter : IMultiEndpointServiceConnectionContainer
+    {
+        private readonly ILogger _logger;
+        internal IReadOnlyCollection<HubServiceEndpoint> TargetEndpoints { get; }
+
+        public MultiEndpointMessageWriter(IReadOnlyCollection<HubServiceEndpoint> targetEndpoints, ILoggerFactory loggerFactory)
+        {
+            TargetEndpoints = targetEndpoints?.ToList();
```
vicancy, line 25: ~~ToList()~~

vicancy (approved):

PR #1192, opened 2021-01-25
Add emulator doc

vicancy:
Fixes https://github.com/Azure/azure-signalr/issues/969


docs/emulator.md
```diff
@@ -0,0 +1,54 @@
...
+
+3. Run the emulator `asrs-emulator` to list all the available commands:
+![image](https://user-images.githubusercontent.com/668244/93064883-a7e69080-f6aa-11ea-925a-89149e32c233.png)
+
+4. Init the default upstream settings using:
+```
+asrs-emulator upstream init
+```
+
+It inits a default `settings.json` into the current folder, with a default upstream `UrlTemplate` as `http://localhost:7071/runtime/webhooks/signalr`, which is the URL for SignalR's **local** function trigger. When the file is modified, its change will be hot loaded into the emulator.
```
chenkennt, line 37: More information about how to update this file?

docs/emulator.md
```diff
@@ -0,0 +1,54 @@
...
+
+5. Start the emulator
+```
+asrs-emulator start
+```
+
+After the emulator is successfully started, it generates the ConnectionString to be used later, for example, the ConnectionString is `Endpoint=http://localhost;Port=8888;AccessKey=ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789ABCDEFGH;Version=1.0;` as the below screenshot shows.
```
chenkennt, line 44: Can access key and port be customized?

vicancy: port can be customized, AccessKey not

chenkennt (approved):

PR #1196, opened 2021-01-28
Extended service manager options to configure JSON serialization

Andy9FromSpace:
The REST client currently uses the default serialization settings, but it would be very helpful to change the settings (e.g. to enable camel case for property serialization).

### DIFF
--- a/src/Microsoft.Azure.SignalR.Management/Configuration/ServiceManagerOptions.cs
+++ b/src/Microsoft.Azure.SignalR.Management/Configuration/ServiceManagerOptions.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using Newtonsoft.Json;
 using System;
 using System.Net;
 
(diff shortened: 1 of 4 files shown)


Andy9FromSpace:
The test "Microsoft.Azure.SignalR.Common.Tests.CustomizedTimerTests.BasicStartStopTest" that failed was successful when I ran it locally. It does some timer stuff, so I don't think it has anything to do with my changes.

src/Microsoft.Azure.SignalR.Management/Configuration/ServiceManagerOptions.cs
```diff
@@ -41,8 +42,13 @@ public class ServiceManagerOptions
         /// </summary>
         public ServiceTransportType ServiceTransportType { get; set; } = ServiceTransportType.Transient;
 
-        internal string ProductInfo { get; set; }
+        /// <summary>
+        /// Gets the json serializer settings that will be used to serialize content sent to Azure SignalR Service.
+        /// </summary>
+        public JsonSerializerSettings JsonSerializerSettings { get; } = new JsonSerializerSettings();
```
vicancy, line 48: also allow `set` following the convention of other properties?

vicancy (approved):

src/Microsoft.Azure.SignalR.Management/Configuration/ServiceManagerOptions.cs
Andy9FromSpace: I didn't add a setter because there shouldn't be a need to set a new instance. I did it the same way like in https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.mvcnewtonsoftjsonoptions.serializersettings as developers should be used to that.

And what about the failed test? Do you agree that it doesn't have anything to do with my changes?

Y-Sindo: The test failed randomly, already rerun and passed.

PR #1199, opened 2021-02-04
fix handshake bug

wanlwanl:
 - Detail 1
 - Detail 2

### DIFF
--- a/src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManager.cs
+++ b/src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManager.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


vicancy:
Could you first create an issue and link this PR to the issue, this can provide a detailed reference to the user what we fixed in the release

src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManager.cs
```diff
@@ -88,24 +89,21 @@ public override async Task SendConnectionAsync(string connectionId, string metho
...
+            IDictionary<string, ReadOnlyMemory<byte>> payloads = new Dictionary<string, ReadOnlyMemory<byte>>();
             if (serviceConnectionContext.Protocol != null)
             {
-                var message = new ConnectionDataMessage(connectionId, SerializeProtocol(serviceConnectionContext.Protocol, methodName, args)).WithTracingId();
-                if (message.TracingId != null)
-                {
-                    MessageLog.StartToSendMessageToConnection(Logger, message);
-                }
-                return message;
+                payloads.Add(serviceConnectionContext.Protocol, SerializeProtocol(serviceConnectionContext.Protocol, methodName, args));
```
vicancy, line 95: move the init in L92 into here? Also add a comment describing why we use `MultiConnectionDataMessage`?

vwxyzh (approved):

BrennanConroy:
Test(s)?

PR #1201, opened 2021-02-05
Allow to choose SignalR hub protocol in management SDK

Y-Sindo:
 - Add services from `ServiceManager` to the `ServiceHubContext` created from `ServiceHubContext.WithEndpoints()` , rather than `AddSignalR().Services`, so that we can specify `IHubProtocol` instead of the default protocol added by `AddSignalR()` when create `ServiceManager`.

### Reason
Azure Function runtime v3 uses .Net framework version higher than Net Core 3.0, which uses System.Text.Json instead of Newtonsoft.Json as the default Json protocol of SignalR. To make our function extension compatible with Azure Function v2 and v3, function extension will use Newtonsoft.Json as the SignalR protocol.

### DIFF
--- a/src/Microsoft.Azure.SignalR.Management/ServiceHubContext.cs
+++ b/src/Microsoft.Azure.SignalR.Management/ServiceHubContext.cs
@@ -10,7 +10,9 @@ using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Connections;
 using Microsoft.AspNetCore.SignalR;
+using Microsoft.AspNetCore.SignalR.Protocol;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.SignalR.Management/ServiceHubContext.cs
```diff
@@ -77,9 +83,8 @@ IInternalServiceHubContext IInternalServiceHubContext.WithEndpoints(IEnumerable<
                 //add required service instances
                 .AddSingleton(ServiceProvider.GetRequiredService<IOptions<ServiceManagerOptions>>())
                 .AddSingleton(_negotiateProcessor)
-                .AddSingleton(_endpointManager)
-
-                .BuildServiceProvider()
+                .AddSingleton(_endpointManager);
+                return services.BuildServiceProvider()
```
JialinXin, line 87: Add an empty line before `return` and remove extra spaces before `return`. Also remove extra empty lines in the `ServiceConnection()` call.

JialinXin (approved):

PR #1208, opened 2021-02-09
add service event message in service protocol.

vwxyzh:
Allow service send some warning messages back to hub server, for example:
* there is no connection with a certain certain id.
* there is no client for a certain user.
* there is no client in a certain group.

### DIFF
--- a/src/Microsoft.Azure.SignalR.Protocols/ServiceMessage.cs
+++ b/src/Microsoft.Azure.SignalR.Protocols/ServiceMessage.cs
@@ -227,43 +227,41 @@ namespace Microsoft.Azure.SignalR.Protocol
     /// <summary>
     /// A service warning message
     /// </summary>
-    public class ServiceWarningMessage : ExtensibleServiceMessage
+    public class ServiceEventMessage : ExtensibleServiceMessage
     {
         /// <summary>
-        /// Gets or sets the type of warning object.
-        /// <list type="bullet">
-        /// <item>connection</item>
-        /// <item>user</item>
-        /// <item>group</item>
-        /// </list>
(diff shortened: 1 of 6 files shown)


src/Microsoft.Azure.SignalR.Protocols/ServiceMessage.cs
```diff
@@ -224,6 +224,49 @@ public ServiceErrorMessage(string errorMessage)
...
 
+    /// <summary>
+    /// A service warning message
+    /// </summary>
+    public class ServiceWarningMessage : ExtensibleServiceMessage
+    {
+        /// <summary>
+        /// Gets or sets the type of warning object.
+        /// <list type="bullet">
+        /// <item>connection</item>
```
vicancy, line 235: shall the user care about the case? the case here is all lower case while the case for WarningKind is capitalized.

src/Microsoft.Azure.SignalR.Protocols/ServiceMessage.cs
```diff
@@ -224,6 +224,49 @@ public ServiceErrorMessage(string errorMessage)
         }
     }
 
+    /// <summary>
+    /// A service warning message
+    /// </summary>
+    public class ServiceWarningMessage : ExtensibleServiceMessage
```
vicancy, line 230: shall we consider it some kind of `service event` message?

src/Microsoft.Azure.SignalR.Protocols/ServiceMessage.cs
vwxyzh: use enum now.

vicancy (approved):

Issue #1115, opened 2020-11-18
Capability to create a Hub that can only be used to check ASRS health and clients should not be terminated on that client


junedans:
We are trying to build a negotiation server that will only implement custom negotiate router to help select endpoints to route clients to. Currently we can achieve this by setting the connectionCount option to 0 and removing all the business code from the negotiation server.

However, the negotiation server does not have any ability to check if a endpoint is online or not and the only workaround is quering via REST API. It would be great if the negotiation server can directly check ASRS online status via the SDK API itself.

Y-Sindo:
Fortunately, the capability of providing endpoints online status will come with the multiple endpoints feature we are working on currently. Pull request that enabling providing endpoints online status under persistent mode is supposed to be made in a few weeks.

Y-Sindo:
Under persistent mode, the endpoints online status will depend on the WebSocket connection status, so we suggest you to set the connection count to above 1 to make the status more stable.

Y-Sindo:
Currently REST API is the only solution to check ASRS online status. Please track https://github.com/Azure/azure-functions-signalrservice-extension/issues/145 to get update.

Y-Sindo:
For transient mode, the support will be a bit late, because we have to finish all tasks in https://github.com/Azure/azure-functions-signalrservice-extension/issues/145 first.

PR #1223, opened 2021-02-22
Make negotiation interface for multiple endpoints public

Y-Sindo:
close https://github.com/Azure/azure-signalr/issues/1115

### DIFF
--- a/src/Microsoft.Azure.SignalR.Management/IServiceHubContext.cs
+++ b/src/Microsoft.Azure.SignalR.Management/IServiceHubContext.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http.Connections;
 using Microsoft.AspNetCore.SignalR;
@@ -16,7 +17,7 @@ namespace Microsoft.Azure.SignalR.Management
         /// Performs a negotiation operation asynchronously that routes a client to a Azure SignalR instance.
         /// </summary>
         /// <returns>A negotiation response object that contains an endpoint url and an access token for the client to connect to the Azure SignalR instance. </returns>
-        Task<NegotiationResponse> NegotiateAsync(NegotiationOptions negotiationOptions = null);
+        Task<NegotiationResponse> NegotiateAsync(NegotiationOptions negotiationOptions = null, CancellationToken cancellationToken = default);
 
         /// <summary>
         /// Gets a user group manager instance which implements <see cref="IUserGroupManager"/> that can be used to add and remove users to named groups.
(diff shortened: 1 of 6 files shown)


vwxyzh:
breaking change?

src/Microsoft.Azure.SignalR.Management/IServiceHubContext.cs
```diff
@@ -12,13 +17,25 @@ namespace Microsoft.Azure.SignalR.Management
...
+        /// <param name="userId">The user ID. If null, the identity name in <see cref="HttpContext.User" /> of <paramref name="httpContext"/> will be used.</param>
+        /// <param name="claims">The claim list to be put into access token. If null, the claims in <see cref="HttpContext.User"/> of <paramref name="httpContext"/> will be used.</param>
+        /// <param name="lifetime">The lifetime of the token. The default value is one hour.</param>
+        /// <param name="isDiagnosticClient">The flag whether the client to be connected is a diagnostic client.</param>
+        /// <param name="cancellationToken">Cancellation token for aborting the operation. If null, the <see cref="HttpContext.RequestAborted"/> of <paramref name="httpContext"/> will be used. </param>
+        /// <returns>Client endpoint and access token to Azure SignalR Service.</returns>
+        Task<NegotiationResponse> NegotiateAsync(HttpContext httpContext = null, string userId = null, IList<Claim> claims = null, TimeSpan? lifetime = null, bool isDiagnosticClient = false, CancellationToken cancellationToken = default);
```
vicancy, line 30: move the parameters to a new class e.g. `NegotiateOptions` to avoid later breaking change?

src/Microsoft.Azure.SignalR.Management/NegotiationOptions.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        /// <summary>
+        /// Gets or sets the flag indicates whether the client is a diagnostic client.
+        /// </summary>
+        public bool IsDiagnosticClient { get; set; } = false;
+
+        /// <summary>
+        /// Notifies when the negotiation operation should be cancelled. If null, the <see cref="HttpContext.RequestAborted"/> of <see cref="HttpContext"/> will be used.
+        /// </summary>
+        public CancellationToken CancellationToken { get; set; }
```
vicancy, line 43: this should be a parameter for the async method. Generally, every async method has a cancellation token to interpret the process

vicancy (approved):

src/Microsoft.Azure.SignalR.Management/IServiceHubContext.cs
```diff
@@ -12,13 +13,19 @@ namespace Microsoft.Azure.SignalR.Management
     /// </summary>
     public interface IServiceHubContext : IHubContext<Hub>
     {
+        /// <summary>
+        /// Performs a negotiation operation asynchronously that routes a client to a Azure SignalR instance.
+        /// </summary>
+        /// <returns>A negotiation response object that contains an endpoint url and an access token for the client to connect to the Azure SignalR instance. </returns>
+        Task<NegotiationResponse> NegotiateAsync(NegotiationOptions negotiationOptions = null, CancellationToken cancellationToken = default);
```
davidfowl, line 20: Breaking change

Y-Sindo: Do you think to use abstract class instead of interface to get more flexibility is a good idea? We want to add a new method but don't want to maintain two versions of SDK.

davidfowl: Yep, it is if we plan to change it changes often.

PR #1228, opened 2021-02-26
fix reconnect issue when switching from self-hosted signalr to azure …

bjqian:
…signalr

block all endpoints except negotiate and js proxy, return 404


src/Microsoft.Azure.SignalR.AspNet/Middleware/NegotiateMiddleware.cs
```diff
@@ -78,8 +78,18 @@ public override Task Invoke(IOwinContext owinContext)
...
+            var normalized = context.Request.LocalPath.TrimEnd('/');
+
+            if (normalized.EndsWith("/hubs", StringComparison.OrdinalIgnoreCase) ||
+                normalized.EndsWith("/js", StringComparison.OrdinalIgnoreCase))
+            {
+                return Next.Invoke(owinContext);
+            }
 
-            return Next.Invoke(owinContext);
+            context.Response.StatusCode = 404;
```
vicancy, line 91: any unknown path return 404? I think we should only handle signalr related URLs.. What if the customer uses the same app server to act as a normal web API?

src/Microsoft.Azure.SignalR.AspNet/Middleware/NegotiateMiddleware.cs
```diff
@@ -78,8 +78,18 @@ public override Task Invoke(IOwinContext owinContext)
             {
                 return ProcessNegotiationRequest(owinContext, context);
```
vicancy, line 79: Could you also create an issue describing the problem?

src/Microsoft.Azure.SignalR.AspNet/Middleware/NegotiateMiddleware.cs
bjqian: For AspNet, it seems there is always a mapping for path: 
https://github.com/Azure/azure-signalr/blob/4e234d39729bd16eb750ecc0d9859c1b2aa91847/src/Microsoft.Azure.SignalR.AspNet/OwinExtensions.cs#L62
So I think this middleware will only work for signalr path.

src/Microsoft.Azure.SignalR.AspNet/Middleware/NegotiateMiddleware.cs
bjqian: Add an issue:
https://github.com/Azure/azure-signalr/issues/1236

vicancy (approved):

PR #1234, opened 2021-03-03
Fix bug that user join leave group fails under multiple endpoints

Y-Sindo:


vicancy:
is it possible to add some ut?

vicancy (approved):

PR #1235, opened 2021-03-09
Add always reconnect logic

vicancy:
The default retry policy for aspnet.core signalr using `withAutomaticReconnect` is :
> SignalR will try and reconnect immediately, then 2 seconds later, then again after 10 seconds and then after 30 seconds. At that time, if the connection could still not be reestablished, it will stop trying to reconnect:

When using Azure SignalR, we would suggest clients always reconnect to the service as a common practice.

### DIFF
--- a/samples/ChatSample/ChatSample.Net50/wwwroot/index.html
+++ b/samples/ChatSample/ChatSample.Net50/wwwroot/index.html
@@ -153,7 +153,7 @@
 
             function onConnectionError(error) {
                 if (error && error.message) {
-                    console.error(error.message);
+                    console.warn(error.message);
                     var modal = document.getElementById('myModal');
                 } else {
                     var modal = document.getElementById('closeModal');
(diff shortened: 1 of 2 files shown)


samples/ChatSample/ChatSample.Net50/wwwroot/index.html
```diff
@@ -162,19 +162,27 @@ <h2 class="text-center" style="margin-top: 0; padding-top: 30px; padding-bottom:
...
             var connection = new signalR.HubConnectionBuilder()
-                        .withUrl('/chat')
-                        .withAutomaticReconnect()
-                        .withHubProtocol(new signalR.protocols.msgpack.MessagePackHubProtocol())
-                        .build();
-                            
+                .withUrl('/chat')
+                .withAutomaticReconnect({
+                    nextRetryDelayInMilliseconds: function (retryContext) {
+                        return getRandom(2000, 5000); // always retry
```
chenkennt, line 173: Add some exponential backoff?

vicancy: shall we keep the sample simple?

PR #1243, opened 2021-03-11
support user-assigned managed identity

terencefan:
- Add support for user-assigned managed identity

#### Usage

To choose one of the user-assigned managed identity:

```text
Endpoint={endpoint};AuthType=aad;ClientId={ClientId of user-assigned identity}
```

> There is no need to provide **TenantId** according to this [link](https://docs.microsoft.com/en-us/dotnet/api/overview/azure/service-to-service-authentication)

Doc changes are on the way : )

Fixes:

Issue #1237

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/Auth/AadManagedIdentityOptions.cs
+++ b/src/Microsoft.Azure.SignalR.Common/Auth/AadManagedIdentityOptions.cs
@@ -5,12 +5,27 @@ namespace Microsoft.Azure.SignalR
 {
     internal class AadManagedIdentityOptions : AuthOptions, IAadTokenGenerator
     {
+        private readonly AzureServiceTokenProvider _azureServiceTokenProvider;
+
         internal override string AuthType => "ManagedIdentity";
+
+        internal ManagedIdentityType ManagedIdentityType { get; }
+
+        public AadManagedIdentityOptions()
+        {
+            _azureServiceTokenProvider = new AzureServiceTokenProvider("RunAs=App", AzureActiveDirectoryInstance);
+            ManagedIdentityType = ManagedIdentityType.System;
+        }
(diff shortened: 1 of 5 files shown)


src/Microsoft.Azure.SignalR.Common/Auth/AadManagedIdentityOptions.cs
```diff
@@ -3,14 +3,38 @@
...
+
     internal class AadManagedIdentityOptions : AuthOptions, IAadTokenGenerator
     {
+        private readonly AzureServiceTokenProvider _azureServiceTokenProvider;
+
         internal override string AuthType => "ManagedIdentity";
 
+        internal ManagedIdentityType ManagedIdentityType { get; }
+
+        public AadManagedIdentityOptions(string clientId = null)
```
vicancy, line 21: separate into 2 .ctor

src/Microsoft.Azure.SignalR.Common/Auth/AadManagedIdentityOptions.cs
```diff
@@ -3,14 +3,36 @@
 
 namespace Microsoft.Azure.SignalR
 {
+    internal enum ManagedIdentityType
```
vicancy, line 6: move to another class

vicancy (approved):

enijburg:
Hi Guys, previously the "Endpoint=https://nnn.service.signalr.net;AuthType=aad;" connection string worked for us in combination with IAM role 'SignalR App Server (Preview)' for AAD users and identities. Around 10 days ago we started to get 401 errors from clients connecting after negotiating with the Hub. if I change the connection string to the primary key from azure portal it works. Can this be related to this change>?

Thanks in advance

PR #1274, opened 2021-04-29
add new serivce events: BufferFull

vwxyzh:
When hub servers try to send too many messages, service side will throttle the message sending, it will keep sending messages in a reasonable rate.
We add the `OverloadProtecting` event to notify app server, app server can hook it, and decrease the rate of message sending to avoid OOM in app server side.

### DIFF
--- a/src/Microsoft.Azure.SignalR.Protocols/enums.cs
+++ b/src/Microsoft.Azure.SignalR.Protocols/enums.cs
@@ -45,8 +45,8 @@
         /// </summary>
         NotExisted,
         /// <summary>
-        /// The overload-protecting event.
+        /// The buffer-full event.
         /// </summary>
-        OverloadProtecting,
+        BufferFull,
     }
 }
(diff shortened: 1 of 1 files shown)


vicancy (approved):

src/Microsoft.Azure.SignalR.Protocols/enums.cs
```diff
@@ -40,5 +44,9 @@ public enum ServiceEventKind
         /// The id is not existed.
         /// </summary>
         NotExisted,
+        /// <summary>
+        /// The buffer-full event. Normally it caused by sending messages rapidly.
```
vicancy, line 48:
```suggestion
        /// The buffer-full event. When the server is sending too many messages at the same time, the service would back-pressure the messages to the server-side and also trigger this `BufferFull` event.
```

vicancy (approved):

PR #1283, opened 2021-05-24
Remove unused code

Y-Sindo:
SignalR Function extensions project doesn't use it anymore. `AddSignalRServiceManager()` sets a default product info.
`ServiceHubContextBuilder.WithCallingAssembly()` is never used


vwxyzh:
what about a customer is using old function extension but he try to upgrade our management sdk to new version?

vwxyzh (approved):

PR #1286, opened 2021-06-02
add service messages

wanlwanl:
`CheckConnectionExistenceWithAckMessage` for [Check if the connection with the given connectionId exists
](https://github.com/Azure/azure-signalr/blob/dev/docs/swagger/v1.md#get-check-if-the-connection-with-the-given-connectionid-exists)
`CheckAnyConnectionInGroupWithAckMessage` for [Check if there are any client connections inside the given group
](https://github.com/Azure/azure-signalr/blob/dev/docs/swagger/v1.md#get-check-if-there-are-any-client-connections-inside-the-given-group)
`CheckAnyConnectionInUserWithAckMessage` for [Check if there are any client connections connected for the given user
](https://github.com/Azure/azure-signalr/blob/dev/docs/swagger/v1.md#get-check-if-there-are-any-client-connections-connected-for-the-given-user)

### DIFF
--- a/src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
+++ b/src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
@@ -43,7 +43,7 @@ namespace Microsoft.Azure.SignalR.Protocol
         /// <summary>
         /// Initializes a new instance of the <see cref="LeaveGroupWithAckMessage"/> class.
         /// </summary>
-        /// <param name="userId">The connection Id.</param>
+        /// <param name="userId">The user Id.</param>
         /// <param name="groupName">The group name, from which the connection will leave.</param>
         /// <param name="ackId">The ack Id</param>
         /// <param name="tracingId">The tracing Id of the message.</param>
--- a/src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
+++ b/src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
@@ -99,9 +99,9 @@ namespace Microsoft.Azure.SignalR.Protocol
     }
 
     /// <summary>
-    /// A waiting for ack check-any-connection-in-user message.
+    /// A waiting for ack check-user-existence message.
     /// </summary>
-    public class CheckAnyConnectionInUserWithAckMessage : CheckWithAckMessage
+    public class CheckUserExistenceWithAckMessage : CheckWithAckMessage
     {
         /// <summary>
         /// Gets or sets the user Id.
@@ -114,7 +114,7 @@ namespace Microsoft.Azure.SignalR.Protocol
         /// <param name="userId">The user Id.</param>
         /// <param name="ackId">The ack Id</param>
         /// <param name="tracingId">The tracing Id of the message.</param>
-        public CheckAnyConnectionInUserWithAckMessage(string userId, int ackId = 0, ulong? tracingId = null) : base(ackId, tracingId)
+        public CheckUserExistenceWithAckMessage(string userId, int ackId = 0, ulong? tracingId = null) : base(ackId, tracingId)
         {
             UserId = userId;
         }
--- a/src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
+++ b/src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
@@ -41,11 +41,11 @@ namespace Microsoft.Azure.SignalR.Protocol
         public string GroupName { get; set; }
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="LeaveGroupWithAckMessage"/> class.
+        /// Initializes a new instance of the <see cref="CheckUserInGroupWithAckMessage"/> class.
         /// </summary>
         /// <param name="userId">The user Id.</param>
-        /// <param name="groupName">The group name, from which the connection will leave.</param>
-        /// <param name="ackId">The ack Id</param>
+        /// <param name="groupName">The group name to check whether the user is in or not.</param>
+        /// <param name="ackId">The ack Id.</param>
         /// <param name="tracingId">The tracing Id of the message.</param>
         public CheckUserInGroupWithAckMessage(string userId, string groupName, int ackId = 0, ulong? tracingId = null) : base(ackId, tracingId)
         {
--- a/src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
+++ b/src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
@@ -1,0 +1,122 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.SignalR.Protocol
+{
+    /// <summary>
+    /// Base class of check-with-ack messages between Azure SignalR Service and SDK.
+    /// </summary>
+    public abstract class CheckWithAckMessage : ExtensibleServiceMessage, IAckableMessage, IMessageWithTracingId
+    {
+        /// <summary>
+        /// Gets or sets the ack id.
+        /// </summary>
+        public int AckId { get; set; }
+
+        /// <summary>
+        /// Gets or sets the tracing Id
+        /// </summary>
+        public ulong? TracingId { get; set; }
+
+        protected CheckWithAckMessage(int ackId, ulong? tracingId)
+        {
+            AckId = ackId;
+            TracingId = tracingId;
+        }
+    }
+
+    /// <summary>
+    /// A waiting for ack check-user-in-group message.
+    /// </summary>
+    public class CheckUserInGroupWithAckMessage : CheckWithAckMessage
+    {
+        /// <summary>
+        /// Gets or sets the user Id.
+        /// </summary>
+        public string UserId { get; set; }
+
+        /// <summary>
+        /// Gets or sets the group name.
+        /// </summary>
+        public string GroupName { get; set; }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="LeaveGroupWithAckMessage"/> class.
+        /// </summary>
+        /// <param name="userId">The connection Id.</param>
+        /// <param name="groupName">The group name, from which the connection will leave.</param>
+        /// <param name="ackId">The ack Id</param>
+        /// <param name="tracingId">The tracing Id of the message.</param>
+        public CheckUserInGroupWithAckMessage(string userId, string groupName, int ackId = 0, ulong? tracingId = null) : base(ackId, tracingId)
+        {
+            UserId = userId;
+            GroupName = groupName;
+        }
+    }
+
+    /// <summary>
+    /// A waiting for ack check-any-connection-in-group message.
+    /// </summary>
+    public class CheckAnyConnectionInGroupWithAckMessage : CheckWithAckMessage
+    {
+        /// <summary>
+        /// Gets or sets the group name.
+        /// </summary>
+        public string GroupName { get; set; }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CheckAnyConnectionInGroupWithAckMessage"/> class.
+        /// </summary>
+        /// <param name="groupName">The group name, from which the connection will leave.</param>
+        /// <param name="ackId">The ack Id</param>
+        /// <param name="tracingId">The tracing Id of the message.</param>
+        public CheckAnyConnectionInGroupWithAckMessage(string groupName, int ackId = 0, ulong? tracingId = null) : base(ackId, tracingId)
+        {
+            GroupName = groupName;
+        }
+    }
+
+    /// <summary>
+    /// A waiting for ack check-connection-existence message.
+    /// </summary>
+    public class CheckConnectionExistenceWithAckMessage : CheckWithAckMessage
+    {
+        /// <summary>
+        /// Gets or sets the connection Id.
+        /// </summary>
+        public string ConnectionId { get; set; }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CheckConnectionExistenceWithAckMessage"/> class.
+        /// </summary>
+        /// <param name="connectionId">The connection Id.</param>
+        /// <param name="ackId">The ack Id</param>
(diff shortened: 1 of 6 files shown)


src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        /// <summary>
+        /// Gets or sets the group name.
+        /// </summary>
+        public string GroupName { get; set; }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="LeaveGroupWithAckMessage"/> class.
+        /// </summary>
+        /// <param name="userId">The connection Id.</param>
```
Y-Sindo, line 46: Why the name is `userId` and the description is "connection Id"?

src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// <param name="groupName">The group name, from which the connection will leave.</param>
```
Y-Sindo, line 47: "from which the connection will leave" is a little confusing, could you provide more context?

src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        protected CheckWithAckMessage(int ackId, ulong? tracingId)
+        {
+            AckId = ackId;
+            TracingId = tracingId;
+        }
+    }
+
+    /// <summary>
+    /// A waiting for ack check-user-in-group message.
```
Y-Sindo, line 29: ~~waiting for~~?

src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
```diff
@@ -0,0 +1,122 @@
...
+        public CheckConnectionExistenceWithAckMessage(string connectionId, int ackId = 0, ulong? tracingId = null) : base(ackId, tracingId)
+        {
+            ConnectionId = connectionId;
+        }
+    }
+
+    /// <summary>
+    /// A waiting for ack check-connection-existence-as-user message.
+    /// </summary>
+    public class CheckAnyConnectionInUserWithAckMessage : CheckWithAckMessage
```
Y-Sindo, line 104: `InUser` -> `ForUser` ?

src/Microsoft.Azure.SignalR.Protocols/GroupMessage.cs
```diff
@@ -247,45 +247,4 @@ public LeaveGroupWithAckMessage(string connectionId, string groupName, int ackId
             TracingId = tracingId;
         }
     }
-
-    /// <summary>
-    /// A waiting for ack check-user-in-group message.
-    /// </summary>
-    public class CheckUserInGroupWithAckMessage : ExtensibleServiceMessage, IAckableMessage, IMessageWithTracingId
```
Y-Sindo, line 254: Breaking change?

src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
```diff
@@ -0,0 +1,122 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.SignalR.Protocol
+{
+    /// <summary>
+    /// Base class of check-with-ack messages between Azure SignalR Service and SDK.
+    /// </summary>
+    public abstract class CheckWithAckMessage : ExtensibleServiceMessage, IAckableMessage, IMessageWithTracingId
```
Y-Sindo, line 9: Given that all the `Check*Message` must have ack, no need to add `WithAck` to the name?

src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
wanlwanl: nice catch, just move the old one https://github.com/Azure/azure-signalr/blob/dev/src/Microsoft.Azure.SignalR.Protocols/GroupMessage.cs#L279 to here

src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
wanlwanl: the grammer is right - connection will leave from the `group`

src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
wanlwanl: https://github.com/Azure/azure-signalr/blob/dev/src/Microsoft.Azure.SignalR.Protocols/GroupMessage.cs#L252
old comment

I think it should be `waiting-for-ack`

src/Microsoft.Azure.SignalR.Protocols/GroupMessage.cs
wanlwanl: why? think the message instance <-> bytes is enough

src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
wanlwanl: Since we have non ackable and ackable message, messages implement `IAckableMessage` has the suffix `WithAckMessage`, so I follow the naming convention

src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
Y-Sindo: I am not meaning the grammar, but the context of using the message. Is the message used only when a connection is going to leave a group? If so, I think pointing out the usage scenario in the comments would be better for code readers to understand.

src/Microsoft.Azure.SignalR.Protocols/GroupMessage.cs
Y-Sindo: As this is a public class, users might already depend on the class.

src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
Y-Sindo: I am also wondering the usage scenario of `CheckWithAckMessage`s. Do I need to provide a user interface for SDK users to send this kind of messages?

src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
Y-Sindo: or `OfUser`

src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
Y-Sindo: I think `WithAck` is necessary when we have the same kind messages with ack and without ack, such as `JoinGroupWithAckMessage` and `JoinGroupMessage`. For `Check*Message`, we only have messages with ack, so there is no need to append `WithAck` to the class name, which seems redundant.

src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
Y-Sindo: According to @vwxyzh , when to use `CheckWithAckMessage` depends on users, so this kind of messages is universal, and no need to add `from which the connection will leave.`

wanlwanl: I found the comments are simplyd copied by `LeaveGroupWithAckMessage` by previous author, need to update.

src/Microsoft.Azure.SignalR.Protocols/GroupMessage.cs
vwxyzh: i do not think it is a breaking change.

src/Microsoft.Azure.SignalR.Protocols/ServiceProtocol.cs
```diff
@@ -797,6 +839,45 @@ private static CheckUserInGroupWithAckMessage CreateCheckUserInGroupWithAckMessa
             return result;
         }
 
+        private static CheckAnyConnectionInGroupWithAckMessage CreateCheckAnyConnectionInGroupWithAckMessage(ref MessagePackReader reader, int arrayLength)
+        {
+            var groupName = ReadString(ref reader, "groupName");
+            var ackId = ReadInt32(ref reader, "ackId");
+            
+            var result = new CheckAnyConnectionInGroupWithAckMessage(groupName, ackId);
+            if (arrayLength >= 4)
```
vwxyzh, line 848: remove if.

src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
wanlwanl: I still think `in` is better,
* Check Any Connection **In** User
* Check Any Connection **Of** User
* Check Any Connection **For** User
What do you think? @KKhurin @vicancy

src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
vicancy: Let's use `CheckUserExists/CheckGroupExists/CheckConnectionExists` to be consistent with REST API definitions

src/Microsoft.Azure.SignalR.Protocols/ServiceProtocol.cs
```diff
@@ -87,6 +87,12 @@ private static ServiceMessage ParseMessage(ref MessagePackReader reader)
                     return CreateLeaveGroupWithAckMessage(ref reader, arrayLength);
                 case ServiceProtocolConstants.CheckUserInGroupWithAckMessageType:
                     return CreateCheckUserInGroupWithAckMessage(ref reader, arrayLength);
+                case ServiceProtocolConstants.CheckAnyConnectionInGroupWithAckMessageType:
```
vicancy, line 90: >CheckAnyConnectionInGroupWithAckMessageType [](http://example.com/codeflow?start=46&length=43)

also rename

src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
KKhurin: If we didn't already have `*GroupWithAckMessage` I'd propose `AckableCheckMessage` or `AckableQueryMessage`. 

How about have the `ack` in for the base class and not for the derived ones?

src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
```diff
@@ -0,0 +1,122 @@
...
+    /// </summary>
+    public class CheckUserExistenceWithAckMessage : CheckWithAckMessage
+    {
+        /// <summary>
+        /// Gets or sets the user Id.
+        /// </summary>
+        public string UserId { get; set; }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CheckConnectionExistenceWithAckMessage"/> class.
```
Y-Sindo, line 112: `CheckConnectionExistenceWithAckMessage`

test/Microsoft.Azure.SignalR.Protocols.Tests/ServiceMessageEqualityComparer.cs
```diff
@@ -211,6 +217,27 @@ private bool CheckUserInGroupWithAckMessageEqual(CheckUserInGroupWithAckMessage
                    x.TracingId == y.TracingId;
         }
 
+        private bool CheckAnyConnectionInGroupWithAckMessageEqual(CheckAnyConnectionInGroupWithAckMessage x, CheckAnyConnectionInGroupWithAckMessage y)
+        {
+            return StringEqual(x.GroupName, y.GroupName) &&
+                   x.AckId == y.AckId &&
+                   x.TracingId == y.TracingId;
```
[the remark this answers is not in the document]
vwxyzh: here just verifies the serialization and deserialization.

src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
wanlwanl: review an out-of-date version?

src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
wanlwanl: I totally agree we can have a better naming if the `ackable` message is added at 1st time, but we already have messages with suffix `WithAckMessage`, e.g. `JoinGroupWithAckMessage`, I think following the naming convention is better.

src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
wanlwanl: it's out of date

src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
```diff
@@ -0,0 +1,122 @@
...
+        {
+            UserId = userId;
+            GroupName = groupName;
+        }
+    }
+
+    /// <summary>
+    /// A waiting for ack check-any-connection-in-group message.
+    /// </summary>
+    public class CheckAnyConnectionInGroupWithAckMessage : CheckWithAckMessage
```
Y-Sindo, line 60: How about rename it to `CheckGroupExistence`?

src/Microsoft.Azure.SignalR.Protocols/CheckWithAckMessage.cs
wanlwanl: actually, out rest api doc doesn't have the consistency for now, e.g. `Check if there are any client connections connected for the given user`, maybe we can also update it in the future

vicancy (approved):
:shipit:

PR #1289, opened 2021-06-08
Fix incorrect array size header for ServiceEventMessage

BrennanConroy:


vwxyzh (approved):

vwxyzh:
Should we merge to dev branch first?

PR #1290, opened 2021-06-08
Add close connection interface to Mgm SDK and implement in transient mode

Y-Sindo:


src/Microsoft.Azure.SignalR.Management/ServiceHubContext.cs
```diff
@@ -25,6 +25,16 @@ internal abstract class ServiceHubContext : IServiceHubContext
...
 
+        /// <summary>
+        /// Close a connection asynchronously.
+        /// </summary>
+        /// <param name="connectionId">The connection to close</param>
+        /// <param name="reason">The reason to close the connection. </param>
+        /// <param name="cancellationToken"></param>
+        /// <returns>The created <see cref="System.Threading.Tasks.Task{TResult}">Task</see> that represents the asynchronous operation.</returns>
+        /// <remarks>To get the <paramref name="reason"/> from connection closed event, client should set <see cref="NegotiationOptions.EnableDetailedErrors"/> during negotiation.</remarks>
+        public virtual Task CloseConnectionAsync(string connectionId, string reason = null, CancellationToken cancellationToken = default) => null;
```
vwxyzh, line 36: `null` -> `Task.CompletedTask` or throw exception.

src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
```diff
@@ -118,9 +118,14 @@ public Task<bool> IsUserInGroup(string userId, string groupName, CancellationTok
             return WriteAckableMessageAsync(message);
         }
 
+        public Task CloseConnectionAsync(string connectionId, string reason, CancellationToken cancellationToken)
+        {
+            throw new NotImplementedException();
```
vicancy, line 123: let's not enable this until both mode support it

src/Microsoft.Azure.SignalR.Management/ServiceHubContext.cs
vicancy, line 36: how about putting this into an `IServiceManager` as similar to `IGroupManager` and `IUserGroupManager`?

src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
Y-Sindo: This is not enabled until `ServiceHubContext` is public

src/Microsoft.Azure.SignalR.Management/ServiceHubContext.cs
Y-Sindo: There are another three methods to be added: check connection exists, check user exists, check group exists. Should we avoid interface regarding to extensibility? And the naming seems to be a problem...

src/Microsoft.Azure.SignalR.Management/IServiceHubLifetimeManager.cs
```diff
@@ -1,12 +1,15 @@
...
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace Microsoft.Azure.SignalR.Management
 {
     internal interface IServiceHubLifetimeManager : IHubLifetimeManager, IUserGroupHubLifetimeManager
     {
+        Task CloseConnectionAsync(string connectionId, CancellationToken cancellationToken);
+
         Task DisposeAsync();
```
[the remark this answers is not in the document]
Y-Sindo: I searched the dotnet code base and found that most of the `DisposeAsync()` don't have `CancellationToken`

src/Microsoft.Azure.SignalR.Management/ServiceHubContext.cs
vicancy: hm... check group exists fits more into `IGroupManager` though...

vicancy (approved):
:shipit:

PR #1299, opened 2021-06-11
add message ford user join/leave group with ack message

wanlwanl:


vwxyzh (approved):

src/Microsoft.Azure.SignalR.Protocols/ServiceProtocol.cs
```diff
@@ -388,6 +398,26 @@ private static void WriteUserLeaveGroupMessage(ref MessagePackWriter writer, Use
...
+            writer.Write(ServiceProtocolConstants.UserJoinGroupWithAckMessageType);
+            writer.Write(message.UserId);
+            writer.Write(message.GroupName);
+            writer.Write(message.AckId);
+            message.WriteExtensionMembers(ref writer);
+        }
+
+        private static void WriteUserLeaveGroupWithAckMessage(ref MessagePackWriter writer, UserLeaveGroupWithAckMessage message)
+        {
+            writer.WriteArrayHeader(5);
```
wanlwanl, line 413: should be 4, since no interface is implemented? @vwxyzh 
how about return if count == 0? https://github.com/Azure/azure-signalr/blob/dev/src/Microsoft.Azure.SignalR.Protocols/ServiceMessage.cs#L45

PR #1303, opened 2021-06-15
Add a unified way to customize JSON serialization for Newtonsoft

Y-Sindo:
`_serviceProvider.GetService<IOptions<NewtonsoftServiceHubProtocolOptions>>()` is always not null, so it is wrong to use it to determine whether `ServiceHubContextBuilder.AddNewtonsoftJsonHubProtocol()` is called.

### DIFF
--- a/src/Microsoft.Azure.SignalR.Management/Serialization/NewtonsoftServiceHubProtocolOptions.cs
+++ b/src/Microsoft.Azure.SignalR.Management/Serialization/NewtonsoftServiceHubProtocolOptions.cs
@@ -6,7 +6,7 @@ using Newtonsoft.Json.Serialization;
 
 namespace Microsoft.Azure.SignalR.Management
 {
-    public class NewtonsoftServiceHubProtocolOptions
+    internal class NewtonsoftServiceHubProtocolOptions
     {
         /// <summary>
         /// Gets or sets the settings used to serialize invocation arguments and return values.
--- a/src/Microsoft.Azure.SignalR.Management/Serialization/NewtonsoftServiceHubProtocolOptions.cs
+++ b/src/Microsoft.Azure.SignalR.Management/Serialization/NewtonsoftServiceHubProtocolOptions.cs
@@ -11,7 +11,7 @@ namespace Microsoft.Azure.SignalR.Management
         /// <summary>
         /// Gets or sets the settings used to serialize invocation arguments and return values.
         /// </summary>
-        public JsonSerializerSettings PayloadSerializerSettings { get; } = new()
+        public JsonSerializerSettings PayloadSerializerSettings { get; set; } = new()
         {
             ContractResolver = new CamelCasePropertyNamesContractResolver()
         };
--- a/src/Microsoft.Azure.SignalR.Management/Serialization/NewtonsoftServiceHubProtocolOptions.cs
+++ b/src/Microsoft.Azure.SignalR.Management/Serialization/NewtonsoftServiceHubProtocolOptions.cs
@@ -11,7 +11,7 @@ namespace Microsoft.Azure.SignalR.Management
         /// <summary>
         /// Gets or sets the settings used to serialize invocation arguments and return values.
         /// </summary>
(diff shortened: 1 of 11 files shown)


vwxyzh:
what about `System.Text.Json`?

Y-Sindo:
> what about `System.Text.Json`?

I have a plan to add `System.Text.Json`, but the current problem is that `RestClient` only support `Newtonsoft`, and is not convenient to depend on new options class in Mgm SDK as `RestClient` is under `Common` namespace.

vwxyzh (approved):

src/Microsoft.Azure.SignalR.Management/Serialization/NewtonsoftServiceHubProtocolOptions.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace Microsoft.Azure.SignalR.Management
+{
+    public class NewtonsoftServiceHubProtocolOptions
+    {
+        /// <summary>
+        /// Gets or sets the settings used to serialize invocation arguments and return values.
+        /// </summary>
+        public JsonSerializerSettings PayloadSerializerSettings = new()
+        {
+            ContractResolver = new CamelCasePropertyNamesContractResolver()
```
vicancy, line 16: why put this as the default resolver?

src/Microsoft.Azure.SignalR.Management/Serialization/SerializationDependencyInjectionExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    internal static class SerializationDependencyInjectionExtensions
+    {
+        public static IServiceCollection AddNewtonsoftHubProtocol(this IServiceCollection services, Action<IOptions<NewtonsoftServiceHubProtocolOptions>> configure)
+        {
+            services.Configure(configure);
+
+            // for persistent transport type only:
+#if NETCOREAPP3_0_OR_GREATER
+            services.AddSingleton<IHubProtocol, NewtonsoftJsonHubProtocol>();
```
vicancy, line 19: looks pretty cool

src/Microsoft.Azure.SignalR.Management/Serialization/NewtonsoftServiceHubProtocolOptions.cs
Y-Sindo: To keep consistent with SignalR. Got some customers complains about this.

src/Microsoft.Azure.SignalR.Management/Serialization/NewtonsoftServiceHubProtocolOptions.cs
```diff
@@ -0,0 +1,19 @@
...
+using Newtonsoft.Json.Serialization;
+
+namespace Microsoft.Azure.SignalR.Management
+{
+    public class NewtonsoftServiceHubProtocolOptions
+    {
+        /// <summary>
+        /// Gets or sets the settings used to serialize invocation arguments and return values.
+        /// </summary>
+        public JsonSerializerSettings PayloadSerializerSettings = new()
```
vicancy, line 14: for public one use property instead of field

src/Microsoft.Azure.SignalR.Management/Serialization/NewtonsoftServiceHubProtocolOptions.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Newtonsoft.Json;
+using Newtonsoft.Json.Serialization;
+
+namespace Microsoft.Azure.SignalR.Management
+{
+    public class NewtonsoftServiceHubProtocolOptions
```
vicancy, line 9: do you want this to be internal first?

vicancy (approved):

vicancy (approved):

PR #1304, opened 2021-06-16
Implements `CloseConnection` in persistent mode

Y-Sindo:


vwxyzh (approved):

src/Microsoft.Azure.SignalR.Common/Logging/MessageLog.cs
```diff
@@ -142,6 +142,18 @@ internal static class MessageLog
...
+        private static readonly Action<ILogger, ulong?, string, Exception> _startToCloseConnection =
+                LoggerMessage.Define<ulong?, string>(
+                    LogLevel.Information,
+                    new EventId(140, "StartToCloseConnection"),
+                    "Start to send message {tracingId} to close connection {connectionId}.");
+
+        private static readonly Action<ILogger, ulong?, string, string, Exception> _startToCloseConnectionWithReason =
+                LoggerMessage.Define<ulong?, string, string>(
+                    LogLevel.Information,
+                    new EventId(141, "StartToCloseConnectionWithReason"),
```
wanlwanl, line 154: id increase from 130, then 140, then 141? 
use 150 instead of 141?

src/Microsoft.Azure.SignalR.Common/Logging/MessageLog.cs
```diff
@@ -142,6 +142,18 @@ internal static class MessageLog
                 new EventId(130, "StartToCheckIfUserInGroup"),
                 StartToCheckIfUserInGroupTemplate);
 
+        private static readonly Action<ILogger, ulong?, string, Exception> _startToCloseConnection =
+                LoggerMessage.Define<ulong?, string>(
+                    LogLevel.Information,
+                    new EventId(140, "StartToCloseConnection"),
```
wanlwanl, line 148: better to use the same event name "StartToCloseConnection", need to keep the event name less. is no reason, call it sth like "normal close".

vicancy (approved):

PR #1308, opened 2021-06-18
Fix: Dispose ServiceProvider

Y-Sindo:


vicancy (approved):

src/Microsoft.Azure.SignalR.Management/ServiceHubContextBuilder.cs
```diff
@@ -87,8 +87,8 @@ internal ServiceHubContextBuilder ConfigureServices(Action<IServiceCollection> c
         public async Task<ServiceHubContext> CreateAsync(string hubName, CancellationToken cancellationToken)
         {
             //add requried services
-            var transportType = _services.BuildServiceProvider()
-                .GetRequiredService<IOptions<ServiceManagerOptions>>().Value.ServiceTransportType;
+            using var serviceProvider = _services.BuildServiceProvider();
```
vwxyzh, line 90: we should avoid create a service provider.

davidfowl: Why do we create one?

Y-Sindo: We want to add and initialize services according to the `ServiceTransportType`. Currently `IServiceConnectionContainer` should be added and initialized for persistent mode, but not for transient mode. 

https://github.com/Azure/azure-signalr/blob/0acc57386d7cef314df07e8302b10e86cff24f86/src/Microsoft.Azure.SignalR.Management/DependencyInjectionExtensions.cs#L41-L56

https://github.com/Azure/azure-signalr/blob/0acc57386d7cef314df07e8302b10e86cff24f86/src/Microsoft.Azure.SignalR.Management/ServiceHubContextBuilder.cs#L101-L106

Y-Sindo: `ServiceTransportType` not only come from the configure action, 
https://github.com/Azure/azure-signalr/blob/0acc57386d7cef314df07e8302b10e86cff24f86/src/Microsoft.Azure.SignalR.Management/ServiceHubContextBuilder.cs#L39-L43

but also the `Configuration`:
https://github.com/Azure/azure-signalr/blob/0acc57386d7cef314df07e8302b10e86cff24f86/src/Microsoft.Azure.SignalR.Management/Configuration/ServiceManagerOptionsSetup.cs#L23-L27

So creating a `ServiceProvider` seems to be a clean way to get the final options value, based on the assumption that building a service provider is not expensive if we only required one `IOptions<ServiceManagerOptions>` service. Please correct me if the assumption is wrong.

Y-Sindo: Will do refactoring later to avoid creating `ServiceProvider` twice.

PR #1313, opened 2021-06-21
Add `WithNewtonsoft` to `ServiceManagerBuilder`

Y-Sindo:


vwxyzh (approved):

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
```diff
@@ -51,6 +51,19 @@ public ServiceManagerBuilder WithRouter(IEndpointRouter router)
...
+        internal ServiceManagerBuilder WithNewtonsoftJsonHubProtocol(Action<NewtonsoftServiceHubProtocolOptions> configure)
+        {
+            _services.AddNewtonsoftHubProtocol(configure);
+            return this;
+        }
+
+        // todo: make public
+        internal ServiceManagerBuilder WithNewtonsoftJsonHubProtocol()
+        {
+            return WithNewtonsoftJsonHubProtocol(o => { });
```
vicancy, line 64: `AddNewtonsoftHubProtocol` to check `configure = null` instead?

Y-Sindo: From [Framework Design Guidelines](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/member-overloading)
> DO use member overloading rather than defining members with default arguments.

vicancy: I mean `WithNewtonsoftJsonHubProtocol(null);` as anyway user can pass `null` into the above method overload

Y-Sindo: I think we could support `WithNewtonsoftJsonHubProtocol(null)`, but `WithNewtonsoftJsonHubProtocol()` seems more elegant, and just like dotnet core https://github.com/dotnet/aspnetcore/blob/main/src/SignalR/common/Protocols.NewtonsoftJson/src/NewtonsoftJsonProtocolDependencyInjectionExtensions.cs#L25

vicancy: I think this is a pattern about nullable, either allow null so you can pass in null directly or not allow null and you need to always pass in an object

PR #1321, opened 2021-06-24
add Async postfixes.

terencefan:
 - Detail 1
 - Detail 2

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/Endpoints/AccessKey.cs
+++ b/src/Microsoft.Azure.SignalR.Common/Endpoints/AccessKey.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Security.Claims;
+using System.Threading;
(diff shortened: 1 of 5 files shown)


src/Microsoft.Azure.SignalR.Common/Endpoints/AccessKey.cs
```diff
@@ -29,11 +30,12 @@ protected AccessKey(string endpoint, int? port)
...
         }
 
-        public virtual Task<string> GenerateAccessToken(
+        public virtual Task<string> GenerateAccessTokenAsync(
             string audience,
             IEnumerable<Claim> claims,
             TimeSpan lifetime,
-            AccessTokenAlgorithm algorithm)
+            AccessTokenAlgorithm algorithm,
+            CancellationToken ctoken = default)
```
vicancy, line 38: looks like this `ctoken` is not used?

vicancy (approved):

Issue #1318, opened 2021-06-22
Make DefaultCloseTimeoutMilliseconds configurable


ajbeaven:
I operate a chat app that involves maintaining a list of active users. When a user connects they are added to this list of active users, and when they disconnect they are removed. All clients are notified when a user is added to or removed from this list. 

Here's some simplified code showing the disconnect:

```
public override async Task OnDisconnectedAsync(Exception exception) {
    var currentUserId = GetCurrentUserId();
    activeUsers.RemoveAll(u => u.UserId = currentUserId);

    await Task.Delay(10000);
    if(!activeUsers.Any(u => u.UserId == currentUserId)) {
        await ChatService.MarkUserOffline(currentUserId);
        await Clients.All.SendAsync("UserOffline", currentUserId);
    }

    await base.OnDisconnectedAsync(exception);
}
```

It's a bit more involved than this as the server side needs to deal with many complexities surrounding handling of online users (eg. chat requests sent to users that aren't active when they're caught moving between pages), but this should give you a good idea of the general gist.

You'll note that there is a Task.Delay here. Before this delay was added, when a user refreshed the page or moved to a different page, clients would receive a UserOffline message so the user was immediately removed from the list of online users. They reappeared shortly after though when the new page was loaded and the client reconnected. To avoid this "flickering" the delay was added to only notify clients that the user left if they hadn't reconnected after a 10 second delay. A 10 second delay should be enough to load even the slowest pages on the site and have the client reconnect whereas a delay of less than 5 seconds would in some cases be too short.

Unfortunately ASRS seems to have a problem with this delay because it throws the following exception
```
Microsoft.Azure.SignalR.Common.AzureSignalRException: Cancelled running application task, probably caused by time out.
  ?, in async Task ServiceConnection.ProcessIncomingMessageAsync(ClientConnectionContext connection)
  ?, in async Task ServiceConnection.ProcessClientConnectionAsync(ClientConnectionContext connection)
```

It looks like this occurs because ServiceConnection.cs forcibly closes the disconnect task after 5 seconds:

```
namespace Microsoft.Azure.SignalR
{
    internal partial class ServiceConnection : ServiceConnectionBase
    {
        private const int DefaultCloseTimeoutMilliseconds = 5000;
        ….

        private async Task ProcessIncomingMessageAsync(ClientConnectionContext connection)
        {
            // Wait for the application task to complete
            // application task can end when exception, or Context.Abort() from hub
[...]

PR #1322, opened 2021-06-24
Expose clientTimeout option

vicancy:
To support scenarios described here: https://github.com/Azure/azure-signalr/issues/1318

### DIFF
--- a/src/Microsoft.Azure.SignalR/ServerConnections/ClientConnectionContext.cs
+++ b/src/Microsoft.Azure.SignalR/ServerConnections/ClientConnectionContext.cs
@@ -225,30 +225,30 @@ namespace Microsoft.Azure.SignalR
         /// <summary>
         /// Cancel the outgoing process
         /// </summary>
-        public void CancelOutgoing(int millisecondsDelay = 0)
+        public void CancelOutgoing(TimeSpan? delay = null)
         {
-            if (millisecondsDelay <= 0)
+            if (!delay.HasValue)
             {
                 _abortOutgoingCts.Cancel();
             }
             else
             {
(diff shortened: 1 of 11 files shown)


vwxyzh (approved):

src/Microsoft.Azure.SignalR/ServerConnections/ClientConnectionContext.cs
```diff
@@ -225,30 +225,30 @@ public void TickHeartbeat()
         /// <summary>
         /// Cancel the outgoing process
         /// </summary>
-        public void CancelOutgoing(int millisecondsDelay = 0)
+        public void CancelOutgoing(TimeSpan? delay = null)
         {
-            if (millisecondsDelay <= 0)
+            if (!delay.HasValue)
```
Y-Sindo, line 230: How about inverting the `if` to make the logic more straight-forward?

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnectionFactory.cs
```diff
@@ -29,7 +28,8 @@ internal class ServiceConnectionFactory : IServiceConnectionFactory
             ConnectionDelegate connectionDelegate,
             IClientConnectionFactory clientConnectionFactory,
             IServerNameProvider nameProvider,
-            IServiceEventHandler serviceEventHandler)
+            IServiceEventHandler serviceEventHandler,
+            ServiceOptions options)
```
Y-Sindo, line 32: Change it to IOptions<ServiceOptions> so that no need to add a WeakConnectionFactory?

vicancy:
Looks

PR #1324, opened 2021-06-24
Remove unused param "serverId"

terencefan:


samples/ChatSample/ChatSample.NetCore31/Startup.cs
```diff
@@ -10,6 +10,24 @@ namespace ChatSample.CoreApp3
 {
     public class Startup
     {
+        // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
+        public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
```
vicancy, line 14: why move? revert?

vicancy (approved):

vicancy (approved):

PR #1331, opened 2021-06-28
update message tracing id generator: keep consistent to RT

wanlwanl:
update message tracing id generator: keep consistent to RT

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/ServiceMessages/MessageWithTracingIdHelper.cs
+++ b/src/Microsoft.Azure.SignalR.Common/ServiceMessages/MessageWithTracingIdHelper.cs
@@ -13,19 +13,14 @@ namespace Microsoft.Azure.SignalR
 
         // message tracing id is constructed in the format:
         // from most significant digit to least significant digit:
-        // 1 hex digits:
-        //      1st lowest bit: isFromTracingClient
-        //      2nd lowest bit: 
-        //          direction:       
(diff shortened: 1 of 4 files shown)


src/Microsoft.Azure.SignalR.Common/ServiceMessages/MessageWithTracingIdHelper.cs
```diff
@@ -13,19 +13,14 @@ internal static class MessageWithTracingIdHelper
...
-        public static ulong Generate(bool isFromTracingClient)
+        // Usage:
+        // - Messaging log category is enabled
+        // - EnableMessageTracing is true for Management SDK
+        public static ulong Generate()
         {
-            ulong tracingClientMask = isFromTracingClient ? (ulong)0x1000_0000_0000_0000 : 0;
-            var id = (Prefix | tracingClientMask) + ((ulong)Interlocked.Increment(ref _index));
-            return id;
+            return Prefix + (ulong)Interlocked.Increment(ref _index);
```
wanlwanl, line 23: since standard2.0 has no `Increment(ref uint x)`, still need to keep `_index` as int32

vwxyzh: convert to `uint` is more make sense.

vwxyzh (approved):

PR #1332, opened 2021-06-29
Server connection does not block on client connection lifetime

vicancy:
To solve the issue described here https://github.com/Azure/azure-signalr/issues/1318, to be consistent with self-host SignalR, our SDK should not timeout on long-run client hub logic.

Why we have this timeout is because the client connection lifetime impacts the auto-reconnect of the server connection. When a server connection drops, it first cleans up all the existing client connections, waiting for the client connections to complete their lifetime, and then restart another server connection. If there is some malformed code inside one client hub, it can impact the whole stability of the server connection if there is no timeout.

Server connection, can be considered as the transport layer to carry traffics for some mapped client connections. Server connection drops, similar to network issues, should not terminate the client hub application logic. If the client hub logic depends on the transport layer, the client hub logic can determine if it throws or continue. Considering this, we change the server-connection cleanup logic to not blocked by client lifetime.

### DIFF
--- a/test/Microsoft.Azure.SignalR.Tests/ServiceConnectionTests.cs
+++ b/test/Microsoft.Azure.SignalR.Tests/ServiceConnectionTests.cs
@@ -247,7 +247,7 @@ namespace Microsoft.Azure.SignalR.Tests
                 Assert.Equal(ServiceConnectionStatus.Disconnected, connection.Status);
 
                 // since the service connection ends, the client connection is cleaned up from the collection...
-                Assert.Single(ccm.ClientConnections);
+                Assert.Empty(ccm.ClientConnections);
             }
         }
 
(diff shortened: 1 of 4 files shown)


JialinXin (approved):

test/Microsoft.Azure.SignalR.Tests/ServiceConnectionTests.cs
```diff
@@ -606,7 +610,7 @@ public Task OnClientDisconnectedAsyncForTest(CloseConnectionMessage message)
             }
         }
 
-        private TestServiceConnection MockServiceConnection(IConnectionFactory serviceConnectionFactory = null,
+        private TestServiceConnection MockServiceConnection(CancellationTokenSource cts, IConnectionFactory serviceConnectionFactory = null,
```
vwxyzh, line 613: format

vwxyzh (approved):

PR #1342, opened 2021-07-06
work-around fix for connection migration

terencefan:
The code is for intercepting `CloseMessage` to prevent the message from reaching the client before it has been logically migrated. (There is a short time period that the client connection belongs to 2 servers so we cannot simply ignore messages from the old server)

It was accidentally applied for all scenarios while doing code refactor.
#814 

And has finally been removed in the following PR to fix another issue.
#986 

Since it is a little tricky to have this logic in the server SDK, we decided to move this interceptor into our service.
So it's a work-around fix.


vicancy:
Could you describe the issue in the PR a little bit more? This helps us track the issue in the future.

vicancy (approved):

PR #1359, opened 2021-07-20
Copy functions from `ServiceHubContextBuilder`  to `ServiceManagerBuilder`

Y-Sindo:
Prepare to merge two builder into one to reduce user migration cost.

### DIFF
--- a/src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
+++ b/src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
@@ -25,6 +25,7 @@ namespace Microsoft.Azure.SignalR.Management
     public class ServiceManagerBuilder : IServiceManagerBuilder
     {
         private readonly IServiceCollection _services;
+        private Action<IServiceCollection> _configureAction;
 
         internal ServiceManagerBuilder(IServiceCollection services)
         {
@@ -35,8 +36,6 @@ namespace Microsoft.Azure.SignalR.Management
         {
             _services.AddSignalRServiceManager();
         }
-
-        private Action<IServiceCollection> _configureAction;
 
         /// <summary>
         /// Registers an action used to configure <see cref="IServiceManager"/>.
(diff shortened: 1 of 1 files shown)


src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
```diff
@@ -84,16 +100,56 @@ public ServiceManagerBuilder WithCallingAssembly()
...
         /// Builds <see cref="IServiceManager"/> instances.
         /// </summary>
         /// <returns>The instance of the <see cref="IServiceManager"/>.</returns>
         public IServiceManager Build()
         {
-            return _services.AddSignalRServiceManager()
-                .AddSingleton(_services.ToList() as IReadOnlyCollection<ServiceDescriptor>)
-                .BuildServiceProvider()
+            var serviceCollection = new ServiceCollection().Add(_services)
+                .AddSingleton(_services.ToList() as IReadOnlyCollection<ServiceDescriptor>);
```
vicancy, line 116: `services` is added twice?

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
```diff
@@ -84,16 +100,56 @@ public ServiceManagerBuilder WithCallingAssembly()
...
+            _configureAction?.Invoke(serviceCollection);
+            return serviceCollection.BuildServiceProvider()
                 .GetRequiredService<IServiceManager>();
         }
+
+        /// <summary>
+        /// Builds <see cref="ServiceHubContext"/> instances.
+        /// </summary>
+        /// <returns>The instance of the <see cref="IServiceManager"/>.</returns>
+        public async Task<ServiceHubContext> CreateHubContextAsync(string hubName, CancellationToken cancellationToken)
```
vicancy, line 126: `public` or `internal` first?

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
```diff
@@ -84,16 +99,56 @@ public ServiceManagerBuilder WithCallingAssembly()
...
+            //initialize
+            var connectionContainer = serviceHubContext.ServiceProvider.GetService<IServiceConnectionContainer>();
+            if (connectionContainer != null)
+            {
+                await connectionContainer.ConnectionInitializedTask.OrTimeout(cancellationToken);
+            }
+
+            if (cancellationToken.IsCancellationRequested)
+            {
+                await serviceHubContext?.DisposeAsync();
```
vicancy, line 148: ConnectionInitializedTask can throw, wrap this into finally?

Y-Sindo: I think we should dispose only if exception thrown, so wrap them into catch.

vicancy: do you want to also throw when `cancellationToken.IsCancellationRequested`

vicancy (approved):

PR #1360, opened 2021-07-20
Refactor `ServiceHubContext` initialization

Y-Sindo:
Make the initialization logic unified in the `ServiceHubContextBuilder` no matter what service transport type is

### DIFF
--- a/src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/ConnectionService.cs
+++ b/src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/ConnectionService.cs
@@ -1,0 +1,31 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Hosting;
+
+namespace Microsoft.Azure.SignalR.Management
+{
+    internal class ConnectionService : IHostedService
+    {
+        private readonly IServiceConnectionContainer _connectionContainer;
+
+        public ConnectionService(IServiceConnectionContainer connectionContainer)
+        {
+            _connectionContainer = connectionContainer;
+        }
+
+        public async Task StartAsync(CancellationToken token)
+        {
+            _ = _connectionContainer.StartAsync();
+            await _connectionContainer.ConnectionInitializedTask.OrTimeout(token, TimeSpan.FromMinutes(1), "establishing service connections");
+        }
+
+        public Task StopAsync(CancellationToken _)
+        {
+            return _connectionContainer.StopAsync();
+        }
+    }
+}
--- a/src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/ConnectionService.cs
+++ b/src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/ConnectionService.cs
@@ -17,10 +17,10 @@ namespace Microsoft.Azure.SignalR.Management
             _connectionContainer = connectionContainer;
         }
 
-        public async Task StartAsync(CancellationToken token)
+        public Task StartAsync(CancellationToken token)
         {
             _ = _connectionContainer.StartAsync();
-            await _connectionContainer.ConnectionInitializedTask.OrTimeout(token, TimeSpan.FromMinutes(1), "establishing service connections");
+            return _connectionContainer.ConnectionInitializedTask.OrTimeout(token, TimeSpan.FromMinutes(1), "establishing service connections");
         }
 
         public Task StopAsync(CancellationToken _)
--- a/src/Microsoft.Azure.SignalR.Management/ServiceHubContextBuilder.cs
+++ b/src/Microsoft.Azure.SignalR.Management/ServiceHubContextBuilder.cs
@@ -9,87 +9,22 @@ using System.Threading.Tasks;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
-using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.Options;
+using Microsoft.Extensions.Hosting;
 
 namespace Microsoft.Azure.SignalR.Management
 {
-    public class ServiceHubContextBuilder
+    internal class ServiceHubContextBuilder
     {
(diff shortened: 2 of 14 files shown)


src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/IInitializationTask.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.SignalR.Management
+{
+    internal interface IInitializationTask
```
vicancy, line 9: `IServiceConnectionInitializer`?

src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/PersistentInitializationTask.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        public PersistentInitializationTask(IServiceConnectionContainer connectionContainer)
+        {
+            _connectionContainer = connectionContainer;
+        }
+
+        public async Task StartAsync(CancellationToken token)
+        {
+            _ = _connectionContainer.StartAsync();
+            await _connectionContainer.ConnectionInitializedTask.OrTimeout(token);
```
vicancy, line 21: what if Timeout? What will user see?

src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/IInitializationTask.cs
Y-Sindo: For transient mode, there is no service connection, but a health checker.

src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/PersistentInitializationTask.cs
Y-Sindo: After "Improve timeout error message" commit, user will get a `TimeoutException` with error message "Timeout occurred for establishing service connections after {timeout}."

src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/IInitializationTask.cs
vicancy: then `IInitializer`

src/Microsoft.Azure.SignalR.Management/ServiceManagerImpl.cs
```diff
@@ -36,44 +34,20 @@ public ServiceManagerImpl(IReadOnlyCollection<ServiceDescriptor> services, IServ
...
-            catch
-            {
-                if (serviceHubContext is not null)
-                {
-                    await serviceHubContext.DisposeAsync();
-                }
-                throw;
-            }
+            var builder = new ServiceHubContextBuilder(_services);
+            var serviceHubContext = await builder.CreateAsync(hubName, cancellationToken);
```
vicancy, line 49: nit: direct return instead of `await`?

src/Microsoft.Azure.SignalR.Management/ServiceHubContextImpl.cs
```diff
@@ -55,8 +54,9 @@ public override Task<NegotiationResponse> NegotiateAsync(NegotiationOptions opti
 
         public override async Task DisposeAsync()
         {
-            await _lifetimeManager.DisposeAsync();
-            (ServiceProvider as IDisposable)?.Dispose();
+            var host = ServiceProvider.GetRequiredService<IHost>();
+            await host.StopAsync();
+            host.Dispose();
```
vicancy, line 59: `using` host?

src/Microsoft.Azure.SignalR.Management/ServiceHubContextBuilder.cs
```diff
@@ -9,121 +9,53 @@
...
-
-            if (cancellationToken.IsCancellationRequested)
+            catch
             {
-                await serviceHubContext?.DisposeAsync();
-                return null;
+                if (host != null)
+                {
+                    await host.StopAsync();
+                    host.Dispose();
```
vicancy, line 55: `using` host?

Y-Sindo: If no error is thrown, then the host should keep alive. So we cannot `using` host, otherwise the host will always be disposed.

vicancy: using host in catch since StopAsync can throw?

src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/ConnectionService.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        public ConnectionService(IServiceConnectionContainer connectionContainer)
+        {
+            _connectionContainer = connectionContainer;
+        }
+
+        public async Task StartAsync(CancellationToken token)
+        {
+            _ = _connectionContainer.StartAsync();
+            await _connectionContainer.ConnectionInitializedTask.OrTimeout(token, TimeSpan.FromMinutes(1), "establishing service connections");
```
vicancy, line 23: directly return

src/Microsoft.Azure.SignalR.Management/ServiceHubContextBuilder.cs
Y-Sindo: We can only use `using` when declare a variable? I added a finally block.

src/Microsoft.Azure.SignalR.Management/ServiceHubContextBuilder.cs
```diff
@@ -49,13 +50,20 @@ internal async Task<ServiceHubContext> CreateAsync(string hubName, CancellationT
             }
             catch
             {
+                shouldDispose = true;
                 if (host != null)
                 {
                     await host.StopAsync();
```
vicancy, line 56: using here?
```cs
using (host) {
    await host.StopAsync();
}
```

Y-Sindo: Got it. `using host;` is forbidden, but `using(host)` is Ok.

vicancy (approved):

PR #1363, opened 2021-07-23
Add method to build ServiceManager class from builder

Y-Sindo:


src/Microsoft.Azure.SignalR.Management/ServiceManager.cs
```diff
@@ -11,10 +11,12 @@
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 using Microsoft.Rest;
 
 namespace Microsoft.Azure.SignalR.Management
 {
+    //todo public
     internal class ServiceManager : IServiceManager
```
vicancy, line 20: or public an abstract class

vicancy (approved):

PR #1365, opened 2021-07-26
Allow to configure `HttpTransportType` for server SDK

### LINKED ISSUE #1325: Detect and disable ServerSentEvents transport when using Azure SignalR
### The issue:
Some of the customers’ clients go through proxies that don’t support WebSocket and SSE. A client experiences like below:
1.	Client starts with WebSocket, the proxy swallows “upgrade: WebSocket” header, service returns 401
2.	Client fallback to SSE, the proxy just never returns the SSE response to the client, and as a result, the service never gets SignalR handshake request and aborts the connection after 15s timeout
3.	Client fails to fall back to long polling correctly as described here https://github.com/dotnet/aspnetcore/issues/20500

It is actually also hard for our customers to identify the issue because the issue only happens to customers' customers when they are going through those proxies.

### Current mitigation
You can see that the WebSocket try fails fast, however, SSE has a 15s delay and it’s hard for the service to detect that failure. Currently, the service has mitigation to disable SSE transport type for an IP if clients from that IP get handshake canceled 3 times within 30 minutes. This is not ideal though since for mobile cases, IP can change. Also, this mitigation cannot prevent clients from having the issue.

### Solution

- [ ] 1. Respect SignalR's `HttpConnectionDispatcherOptions` option if the transport type is configured

- [x] 2. Improve the service's side detection logic: Disable SSE for one IP if it HandshakeCancelled once

Y-Sindo:
To fixes #1325 
At first, I try to take SignalR's `HttpConnectionDispatcherOptions` option if the transport type is configured, but found that we can't get the `HttpConnectionDispatcherOptions`, so I make another options.

Usage:
```cs
  .AddAzureSignalR(o =>
  {
      o.ConnectionString = DefaultConnectionString;
      o.TransportTypeDetector = context => ...;
  })

```
If user set the `TransportTypeDetector`, then a claim will be added to the access token.

CliamType: `asrs.u.htt`   Value: the int number of the `HttpTransportType` flag.

If user doesn't set the `TransportDetector`, then no related claim.

### DIFF
--- a/src/Microsoft.Azure.SignalR/HubHost/NegotiateHandler.cs
+++ b/src/Microsoft.Azure.SignalR/HubHost/NegotiateHandler.cs
@@ -7,7 +7,6 @@ using System.Linq;
 using System.Net;
 using System.Security.Claims;
 using System.Threading.Tasks;
-using Microsoft.AspNetCore.Connections;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Connections;
 using Microsoft.AspNetCore.Localization;
@@ -19,27 +18,6 @@ namespace Microsoft.Azure.SignalR
 {
     internal class NegotiateHandler<THub> where THub : Hub
     {
-        //AvailableTransport copy from https://github.com/dotnet/aspnetcore/tree/2be49d930a5fb53e781abd175c3b2a8f8b7827d4/src/SignalR/common/Http.Connections/src/Internal/HttpConnectionDispatcher.cs
-        private static readonly AvailableTransport _webSocketAvailableTransport =
-            new()
-            {
-                Transport = nameof(HttpTransportType.WebSockets),
-                TransferFormats = new List<string> { nameof(TransferFormat.Text), nameof(TransferFormat.Binary) }
-            };
-
-        private static readonly AvailableTransport _serverSentEventsAvailableTransport =
-            new()
-            {
-                Transport = nameof(HttpTransportType.ServerSentEvents),
-                TransferFormats = new List<string> { nameof(TransferFormat.Text) }
-            };
-
-        private static readonly AvailableTransport _longPollingAvailableTransport =
-            new()
-            {
-                Transport = nameof(HttpTransportType.LongPolling),
-                TransferFormats = new List<string> { nameof(TransferFormat.Text), nameof(TransferFormat.Binary) }
-            };
         private readonly IUserIdProvider _userIdProvider;
         private readonly IConnectionRequestIdProvider _connectionRequestIdProvider;
         private readonly Func<HttpContext, IEnumerable<Claim>> _claimsProvider;
@@ -55,7 +33,7 @@ namespace Microsoft.Azure.SignalR
         private readonly int _customHandshakeTimeout;
         private readonly string _hubName;
         private readonly ILogger<NegotiateHandler<THub>> _logger;
-        private readonly Func<HttpContext, HttpTransportType> _transportTypeProvider;
+        private readonly Func<HttpContext, HttpTransportType> _transportTypeDetector;
 
         public NegotiateHandler(
             IOptions<HubOptions> globalHubOptions,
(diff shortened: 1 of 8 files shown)


davidfowl:
PR description? What does this do and what is it for?

src/Microsoft.Azure.SignalR/HubHost/ServiceHubDispatcherOptions.cs
```diff
@@ -0,0 +1,16 @@
...
+    //todo: public for [HttpTransportType] feature
+    /// <summary>
+    /// Options used to configure the Azure SignalR Service hub dispatcher.
+    /// </summary>
+    internal class ServiceHubDispatcherOptions
+    {
+        /// <summary>
+        /// Gets or sets a bitmask combining one or more <see cref="HttpTransportType"/> values that specify what transports the server should use to receive HTTP requests.
+        /// </summary>
+        public HttpTransportType Transports { get; set; } = HttpTransports.All
```
vicancy, line 14: is putting this into current `ServiceOptions` enough?

Y-Sindo: In SignalR, this option is per hub, so I think the user might need to use different options for different hub.

Y-Sindo:
> PR description? What does this do and what is it for?

PR description updated.

Y-Sindo:
Hello @davidfowl ,
The PR description is updated.
Do you think to make the `Action<...DispatcherOptions>` a parameter of the `MapHub` method is a good design? Or if it better to put it into `ServiceOptions`? Your opinion will be helpful for us.

davidfowl:
I though this middleware was deprecated  in ASP.NET Core 3.0 applications since we moved to a routing model. Why does the UseAzureSignalR API have its own routing logic now?

Y-Sindo:
> I though this middleware was deprecated in ASP.NET Core 3.0 applications since we moved to a routing model. Why does the UseAzureSignalR API have its own routing logic now?

Yes, we don't need this middleware now in ASP.NET Core 3.0, but the problem is we cannot get the `HttpConnectionDispatcherOptions` set by the origin [`MapHub` method](https://github.com/dotnet/aspnetcore/blob/2be49d930a5fb53e781abd175c3b2a8f8b7827d4/src/SignalR/server/SignalR/src/HubEndpointRouteBuilderExtensions.cs#L40), so we can't reuse the origin MapHub method.

src/Microsoft.Azure.SignalR/HubHost/NegotiateHandler.cs
```diff
@@ -77,13 +101,14 @@ public async Task<NegotiationResponse> Process(HttpContext context)
...
             var queryString = GetQueryString(request.QueryString.HasValue ? request.QueryString.Value.Substring(1) : null, cultureName);
+            var availableTransports = GetAvailableTransports(_transportTypeProvider.Invoke(context));
 
             return new NegotiationResponse
             {
                 Url = provider.GetClientEndpoint(_hubName, originalPath, queryString),
                 AccessToken = await provider.GenerateClientAccessTokenAsync(_hubName, claims),
                 // Need to set this even though it's technically protocol violation https://github.com/aspnet/SignalR/issues/2133
-                AvailableTransports = new List<AvailableTransport>()
+                AvailableTransports = availableTransports
```
vicancy, line 111: client does not respect this `AvailableTransports` when it is a redirect response, we need to pass the info through claim to the service, and let service return `AvailableTransports`

davidfowl:
I'm not a fan of exposing behavior on the middleware like this. We deprecated the middleware in SignalR itself in favor of the routing method. Let me dig to see if there's a better way to go here.

Y-Sindo:
> I'm not a fan of exposing behavior on the middleware like this. We deprecated the middleware in SignalR itself in favor of the routing method. Let me dig to see if there's a better way to go here.

@davidfowl Thank you for your reply. Now we have made it a function of `HttpContext` and moved it into `ServiceOptions`, similar to the way we allow user to configure which claims are passed to SignalR runtime.
```cs
public class ServiceOptions : IServiceEndpointOptions {
...
       public Func<HttpContext, IEnumerable<Claim>> ClaimsProvider { get; set; } = null;
       public Func<HttpContext, HttpTransportType> TransportTypeDetector { get; set; } = null;
...
}
```
This new usage sample is updated in the PR description. It might not be the ideal solution, and we look forward to hearing from you.

vicancy (approved):
LGTM.

PR #1371, opened 2021-07-30
Add rest health checker

Y-Sindo:
In Mgm SDK, under persistent mode each negotiation will check the endpoint health via the status of WebSocket connection. However transient mode doesn't have this mechanism. So we need a health checker to check the endpoint health via REST api periodically.


src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/RestHealthCheckService.cs
```diff
@@ -1,22 +1,84 @@
...
+            _logger = logger;
+            _hubName = hubName;
+            _cancellationTokenSource = new();
+            _cancellationToken = _cancellationTokenSource.Token;
+        }
+
         public Task StartAsync(CancellationToken cancellationToken)
         {
-            return Task.CompletedTask;
+            _timer = new Timer(CheckEndpointHealthAsync, null, TimeSpan.Zero, CheckInterval);
```
vicancy, line 41: try `TimeAwaitable`

src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/RestHealthCheckService.cs
```diff
@@ -1,22 +1,84 @@
...
+        public Task StopAsync(CancellationToken _)
         {
+            _cancellationTokenSource.Cancel();
+            _firstCheckCompletionSource.TrySetCanceled();
             return Task.CompletedTask;
         }
+
+        private async void CheckEndpointHealthAsync(object _)
+        {
+            var endpoints = _serviceEndpointManager.GetEndpoints(_hubName);
```
vicancy, line 54: not used?

src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/RestHealthCheckService.cs
```diff
@@ -1,22 +1,84 @@
...
+                    var isHealthy = await client.IsServiceHealthy(_cancellationToken);
+                    endpoint.Online = isHealthy;
+                }
+                catch (OperationCanceledException ex) when (ex.CancellationToken == _cancellationToken)
+                {
+                    // the health checker is stopping.
+                }
+                catch (Exception ex)
+                {
+                    _logger.LogError(ex, "Check health status failed for endpoint {endpoint}", endpoint.Endpoint);
```
vicancy, line 69: use the standard log fomat

vicancy: warning is enough?

src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/RestHealthCheckService.cs
```diff
@@ -1,22 +1,84 @@
...
+                    endpoint.Online = false;
```
vicancy, line 70: is this too strict that when health check throws return false?

src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/RestHealthCheckService.cs
```diff
@@ -1,22 +1,84 @@
...
+                }
+                finally
+                {
+                    _firstCheckCompletionSource.TrySetResult(true);
```
vicancy, line 74: when one endpoint finishes healthy check, return true? what if error throws?

src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/RestHealthCheckService.cs
```diff
@@ -1,22 +1,84 @@
...
 
-        public Task StopAsync(CancellationToken cancellationToken)
+        public Task StopAsync(CancellationToken _)
         {
+            _cancellationTokenSource.Cancel();
+            _firstCheckCompletionSource.TrySetCanceled();
             return Task.CompletedTask;
         }
+
+        private async void CheckEndpointHealthAsync(object _)
```
vicancy, line 52: intentionally use `void`?

src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/RestHealthCheckService.cs
Y-Sindo: What is the usage of `TimeAwaitable`?

src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/RestHealthCheckService.cs
vicancy: what if client http timeout?

src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/RestHealthCheckService.cs
```diff
@@ -1,22 +1,110 @@
...
+                var isHealthy = false;
+                bool needRetry;
+                do
+                {
+                    try
+                    {
+                        using var client = _clientFactory.Create(endpoint);
+                        isHealthy = await client.IsServiceHealthy(default);
+                    }
+                    catch (Exception ex)
```
vicancy, line 63: when it throws `isHealthy` is not set to `false`, which means it is possible that isHealthy keeps last `true` value, is it by design?

src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/RestHealthCheckService.cs
```diff
@@ -1,22 +1,110 @@
...
+                while (await _timer)
+                {
+                    await CheckEndpointHealthAsync();
+                }
+            }
+        }
+
+        private static class Log
+        {
+            private static readonly Action<ILogger, string, Exception> _restHealthCheckFailed = LoggerMessage.Define<string>(LogLevel.Information, new EventId(1, nameof(RestHealthCheckFailed)), "Failed to check health state for endpoint {endpoint}");
```
vicancy, line 93: make the error message more like `Information` log

src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/RestHealthCheckService.cs
Y-Sindo: The local var `isHealthy` is initialized to `false`. If exception is thrown, `isHealthy` is not set new value and won't keep last true value.

vicancy (approved):

PR #1378, opened 2021-08-10
Raise user-friendly ex when aad auth failed.

terencefan:
According to [this](https://docs.microsoft.com/en-us/azure/role-based-access-control/troubleshooting#role-assignment-changes-are-not-being-detected), a role assignments change requires up to 30 minutes to take effect if the token had not been forced refreshed. 

However, it took seldomly longer than 15 minutes in our daily integration tests.

I'd like to try reducing the time to take effect, but now, let's give our customer some hints.

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/Exceptions/AzureSignalRUnauthorizedException.cs
+++ b/src/Microsoft.Azure.SignalR.Common/Exceptions/AzureSignalRUnauthorizedException.cs
@@ -9,7 +9,7 @@ namespace Microsoft.Azure.SignalR.Common
     [Serializable]
     public class AzureSignalRUnauthorizedException : AzureSignalRException
     {
-        private const string ErrorMessage = "Authorization failed. Make sure you provide the correct connection string and have access to the resource.";
+        private const string ErrorMessage = "Authorization failed. If you were using AccessKey, please check connection string and see if the AccessKey is correct. If you were using Azure Active Directory, please note that the role assignments will take up to 30 minutes to take effect if it was added recently.";
 
         public AzureSignalRUnauthorizedException(string requestUri, Exception innerException) : base(string.IsNullOrEmpty(requestUri) ? ErrorMessage : $"{ErrorMessage} Request Uri: {requestUri}", innerException)
         {
--- a/src/Microsoft.Azure.SignalR.Common/Exceptions/ExceptionExtensions.cs
+++ b/src/Microsoft.Azure.SignalR.Common/Exceptions/ExceptionExtensions.cs
@@ -4,9 +4,11 @@
 using System;
 using System.Net;
 using System.Net.Http;
+using System.Net.WebSockets;
+
 using Microsoft.Rest;
 
-namespace Microsoft.Azure.SignalR.Common.RestClients
+namespace Microsoft.Azure.SignalR.Common
 {
     internal static class ExceptionExtensions
     {
@@ -33,6 +35,20 @@ namespace Microsoft.Azure.SignalR.Common.RestClients
                     return new AzureSignalRInaccessibleEndpointException(baseUri.ToString(), requestException);
                 default:
                     return e;
+            }
+        }
+
+        internal static Exception WrapAsAzureSignalRException(this Exception e)
+        {
+            switch (e)
+            {
+                case WebSocketException webSocketException:
+                    if (e.Message.StartsWith("The server returned status code \"401\""))
+                    {
+                        return new AzureSignalRUnauthorizedException(webSocketException);
+                    }
+                    return e;
+                default: return e;
             }
         }
     }
--- a/src/Microsoft.Azure.SignalR.Management/ServiceManagerImpl.cs
+++ b/src/Microsoft.Azure.SignalR.Management/ServiceManagerImpl.cs
@@ -7,10 +7,8 @@ using System.Linq;
 using System.Security.Claims;
 using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.Azure.SignalR.Common.RestClients;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
-using Microsoft.Rest;
 
 namespace Microsoft.Azure.SignalR.Management
 {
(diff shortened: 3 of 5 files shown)


src/Microsoft.Azure.SignalR.Common/Exceptions/AzureSignalRUnauthorizedException.cs
```diff
@@ -9,12 +9,16 @@ namespace Microsoft.Azure.SignalR.Common
     [Serializable]
     public class AzureSignalRUnauthorizedException : AzureSignalRException
     {
-        private const string ErrorMessage = "Authorization failed. Make sure you provide the correct connection string and have access to the resource.";
+        private const string ErrorMessage = "Authorization failed. Please check the connection string and role assignments. The role assignments will take up to 30 minutes to take effect if it was added recently.";
```
vicancy, line 12: use different error messages for accessKey auth and aad auth?

src/Microsoft.Azure.SignalR.Management/ServiceManagerImpl.cs
```diff
@@ -7,10 +7,8 @@
 using System.Security.Claims;
 using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.Azure.SignalR.Common.RestClients;
```
vicancy, line 10: ad a general practice, if the file only changes format, revert the change to keep the code change PR minimum. (Format change can be in a separate PR)

src/Microsoft.Azure.SignalR.Common/Exceptions/ExceptionExtensions.cs
```diff
@@ -35,5 +37,19 @@ internal static Exception WrapAsAzureSignalRException(this Exception e, Uri base
                     return e;
             }
         }
+
+        internal static Exception WrapAsAzureSignalRException(this Exception e, WebSocketConnectionOptions options)
+        {
+            switch (e)
+            {
+                case WebSocketException webSocketException:
+                    if (options.AuthType == AuthType.AzureAd && e.Message.Contains("401"))
```
vicancy, line 46: looks cool, can it be more specific using "StartWith"? Could it be also applied to other auth types?

src/Microsoft.Azure.SignalR.Common/Exceptions/AzureSignalRUnauthorizedException.cs
terencefan: This message is only for AuthType = AzureAd

src/Microsoft.Azure.SignalR.Management/ServiceManagerImpl.cs
terencefan: I relocate this file and the namespace is no longer exists, so I think this change relates to the PR.

src/Microsoft.Azure.SignalR.Common/Exceptions/ExceptionExtensions.cs
terencefan: 401 makes the most sense in that error message. If we use the `StartWith`, it could be `.StartWith("The server returned status code '401'"), which is no better than `Contains`.

src/Microsoft.Azure.SignalR.Common/Exceptions/AzureSignalRUnauthorizedException.cs
vicancy: The exception class name looks much more general though.

src/Microsoft.Azure.SignalR.Common/Exceptions/ExceptionExtensions.cs
vicancy: Is it possible that other place contains some random 401 number that is not related to status code?

terencefan: The chance is low. They doesn't contain random strings as I know.

src/Microsoft.Azure.SignalR.Common/Exceptions/AzureSignalRUnauthorizedException.cs
terencefan: It's a public class, it doesn't make sense if we raised an exception with its type name contains "AzureAd".

src/Microsoft.Azure.SignalR.Common/Exceptions/AzureSignalRUnauthorizedException.cs
```diff
@@ -9,12 +9,16 @@ namespace Microsoft.Azure.SignalR.Common
...
 
         public AzureSignalRUnauthorizedException(string requestUri, Exception innerException) : base(string.IsNullOrEmpty(requestUri) ? ErrorMessage : $"{ErrorMessage} Request Uri: {requestUri}", innerException)
         {
         }
 
+        internal AzureSignalRUnauthorizedException(Exception innerException) : base(ErrorMessage, innerException)
```
vicancy, line 18: Let's either use a more specific Aad exception or pass in the AAd exception message, instead of having aad specific error message inside this general exception class

src/Microsoft.Azure.SignalR.Common/Exceptions/ExceptionExtensions.cs
```diff
@@ -35,5 +37,19 @@ internal static Exception WrapAsAzureSignalRException(this Exception e, Uri base
                     return e;
             }
         }
+
+        internal static Exception WrapAsAzureSignalRException(this Exception e, WebSocketConnectionOptions options)
+        {
+            switch (e)
+            {
+                case WebSocketException webSocketException:
```
vicancy, line 45: let's use `StartWith` instead of `Contains` to make the check more straight-forward

src/Microsoft.Azure.SignalR.Common/Exceptions/AzureSignalRUnauthorizedException.cs
terencefan: It's impossible.

First, we can't change an existing public exception name.
Second, it's impossible to determine if the a request was authorized by `AAD Token` or `AccessToken` when calling rest APIs.

vicancy (approved):

Issue #1361, opened 2021-07-21
AAD auth flow doesn't work when Network access control is configured.


terencefan:
It is common to allow only `Server connection` and `Client connection` if you don't want others to access your `REST API` from the public network.

However, our aad auth workflow depends on a REST API to get `AccessKey` to signing `AccessToken` when doing the negotiation.

In this scenario, you will get a 401 unauthorized error while starting the server, and the negotiation will also fail due to this issue.

**The workaround is to also allow devices from the public network to access REST API.**

![image](https://user-images.githubusercontent.com/4176855/126465938-0540c13b-7fdb-4a9c-927c-5c2d77be2a39.png)

We'd like to get rid of calling REST API in the aad auth flow, and it will be implemented in a few steps.

PR #1380, opened 2021-08-12
Send accesskey request message

terencefan:
Send `AccessKeyRequest` message with AAD Token while using AAD auth.

Changes:
1. Instead of handling message by message handler, each server connection will handle its own `AccessKeyResponse` message now and update the `AadAccessKey`. Since the references are the same, this is only a refactor.
2. Introduce a task to send an `AccessKeyRequest` with AAD token to refresh the accesskey for each server connection.

#1361


src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -145,12 +147,18 @@ public async Task StartAsync(string target = null)
                 _serviceConnectionStartTcs.TrySetResult(true);
                 try
                 {
+                    var syncTimer = new TimerAwaitable(TimeSpan.Zero, DefaultSyncAzureIdentityInterval);
```
vicancy, line 150: you don't dispose the timer if it is not aadKey.

Can we prevent creating the timer if it is not aadKey?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -455,6 +468,22 @@ private async Task<bool> ReceiveHandshakeResponseAsync(PipeReader input, Cancell
...
+        private async Task UpdateAzureIdentityAsync(AccessKey key, TimerAwaitable timer)
+        {
+            if (key is AadAccessKey aadKey)
+            {
+                using (timer)
+                {
+                    timer.Start();
+                    while (await timer)
+                    {
+                        var message = new AccessKeyRequestMessage(await aadKey.GenerateAadTokenAsync());
```
vicancy, line 480: can `await aadKey.GenerateAadTokenAsync()` throw?

terencefan: The server cannot start if it throwed.

PR #1387, opened 2021-08-18
Extend Management rest health check interval

Y-Sindo:
Once every 10 seconds is a little annoying

### DIFF
--- a/src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/RestHealthCheckService.cs
+++ b/src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/RestHealthCheckService.cs
@@ -54,13 +54,6 @@ namespace Microsoft.Azure.SignalR.Management
             return Task.CompletedTask;
         }
 
-        //used to reduce test time
-        internal class HealthCheckOption
-        {
-            public TimeSpan CheckInterval { get; set; }
-            public TimeSpan RetryInterval { get; set; }
-        }
-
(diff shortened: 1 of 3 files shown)


test/Microsoft.Azure.SignalR.Management.Tests/RestHealthCheckServiceFacts.cs
```diff
@@ -71,12 +71,13 @@ public async Task TestRestHealthCheckServiceWithEndpointFromHealthyToUnhealthy()
                 .WithLoggerFactory(loggerFactory)
                 .BuildServiceManager()
                 .CreateHubContextAsync(HubName, default);
-            
+
             //The first negotiation is OK
-            Assert.NotNull(serviceHubContext.NegotiateAsync().Result);
+            Assert.NotNull(serviceHubContext.NegotiateAsync().AsTask().Result);
```
vwxyzh, line 76: use `await`

src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/RestHealthCheckService.cs
```diff
@@ -46,6 +54,13 @@ public Task StopAsync(CancellationToken _)
             return Task.CompletedTask;
         }
 
+        //used to reduce test time
+        internal class HealthCheckOption
```
vicancy, line 58: move out from the class

vicancy (approved):

PR #1389, opened 2021-08-18
add constructor for aad keys

terencefan:
 - Detail 1
 - Detail 2


src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpoint.cs
```diff
@@ -66,16 +69,45 @@ public ServiceEndpoint(string connectionString, EndpointType type = EndpointType
...
         }
 
-        public ServiceEndpoint(ServiceEndpoint endpoint)
+        /// <summary>
+        /// Aad constructor with key
+        /// </summary>
+        /// <param name="key"></param>
+        /// <param name="endpoint"></param>
+        /// <param name="credential"></param>
+        public ServiceEndpoint(string key, Uri endpoint, TokenCredential credential) : this(endpoint, credential)
```
Y-Sindo, line 78: I think one ctor `ServiceEndpoint(Uri endpoint, TokenCredential credential, EndpointType endpointType, string name)` is enough.

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpoint.cs
```diff
@@ -66,16 +69,45 @@ public ServiceEndpoint(string connectionString, EndpointType type = EndpointType
...
+        {
+            (Name, EndpointType) = ParseKey(key);
+        }
+
+        /// <summary>
+        /// Aad constructor
+        /// </summary>
+        /// <param name="endpoint"></param>
+        /// <param name="credential"></param>
+        public ServiceEndpoint(Uri endpoint, TokenCredential credential = null)
```
Y-Sindo, line 88: I think we should ensure credential is not null. The fallback login of `TokenCredential` should be handled by the callee.

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpoint.cs
Y-Sindo: Currently I prefer not to set the endpoint type in the key.

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpoint.cs
```diff
@@ -47,11 +40,31 @@ public class ServiceEndpoint
...
+        /// The customized endpoint that the client will be redirected to
+        /// </summary>
+        internal string ClientEndpoint { get; }
+
+        internal string Version { get; }
+
+        internal AccessKey AccessKey { get; private set; }
+
+        /// <summary>
+        /// Connection string constructor with dict key
```
vicancy, line 53: nit: also update the comment

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpoint.cs
```diff
@@ -3,27 +3,20 @@
...
+using Azure.Core;
+
 namespace Microsoft.Azure.SignalR
 {
     public class ServiceEndpoint
     {
         public string ConnectionString { get; }
 
-        public EndpointType EndpointType { get; }
+        public EndpointType EndpointType { get; } = EndpointType.Primary;
```
vwxyzh, line 14: revert, `EndpointType.Primary` is the default value.

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpoint.cs
```diff
@@ -3,27 +3,20 @@
...
 
-        public virtual string Name { get; internal set; }
+        public virtual string Name { get; internal set; } = "";
```
vwxyzh, line 16: revert, set in constructor.

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpoint.cs
```diff
@@ -66,16 +79,48 @@ public ServiceEndpoint(string connectionString, EndpointType type = EndpointType
...
+        /// <summary>
+        /// Azure active directory constructor
+        /// </summary>
+        /// <param name="endpoint"></param>
+        /// <param name="credential"></param>
+        /// <param name="endpointType"></param>
+        /// <param name="name"></param>
+        public ServiceEndpoint(Uri endpoint, TokenCredential credential, EndpointType endpointType = EndpointType.Primary, string name = "")
+        {
+            if (endpoint.Scheme != Uri.UriSchemeHttp && endpoint.Scheme != Uri.UriSchemeHttps)
```
vicancy, line 102: null check

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpoint.cs
```diff
@@ -47,11 +40,31 @@ public class ServiceEndpoint
...
+            (Name, EndpointType) = Parse(nameWithEndpointType);
         }
 
+        /// <summary>
+        /// Connection string constructor
+        /// </summary>
+        /// <param name="connectionString"></param>
+        /// <param name="type"></param>
+        /// <param name="name"></param>
         public ServiceEndpoint(string connectionString, EndpointType type = EndpointType.Primary, string name = "")
```
vicancy, line 68: There should be one .ctor to be called by other .ctors

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpoint.cs
```diff
@@ -66,16 +79,48 @@ public ServiceEndpoint(string connectionString, EndpointType type = EndpointType
...
+        }
+
+        /// <summary>
+        /// Copy constructor
+        /// </summary>
+        /// <param name="other"></param>
+        public ServiceEndpoint(ServiceEndpoint other)
         {
-            if (endpoint != null)
+            if (other != null)
```
vwxyzh, line 116: should we throw exception when `other` is `null`?

vicancy (approved):

PR #1390, opened 2021-08-19
update constructor for access key

terencefan:
1. Update `AccessKey` and `AadAccessKey` constructor to use `Uri endpoint` insteadof `string endpoint` & `int? port`
2. Replace `StringBuilder` with `UriBuilder` to generate Urls and Tokens.

## This PR includes some breaking changes
1. `ServiceEndpoint.Port` will return 80 or 443 instead of `null` when Port is not given in the `ConnectionString`, depends on the scheme type of the given endpoint.
2. `ServiceEndpoint.Endpoint` will convert its hostname into **lowercase** even when the given hostname is in uppercase. It will also influence the generated **accessToken** and **url**. [RFC4343](https://datatracker.ietf.org/doc/html/rfc4343)

### DIFF
--- a/src/Microsoft.Azure.SignalR.AspNet/EndpointProvider/ServiceEndpointProvider.cs
+++ b/src/Microsoft.Azure.SignalR.AspNet/EndpointProvider/ServiceEndpointProvider.cs
@@ -21,10 +21,10 @@ namespace Microsoft.Azure.SignalR.AspNet
         private const string ServerPath = "aspnetserver";
 
         private readonly string _endpoint;
+        private readonly string _serverEndpoint;
         private readonly string _clientEndpoint;
         private readonly AccessKey _accessKey;
         private readonly string _appName;
-        private readonly int? _port;
         private readonly TimeSpan _accessTokenLifetime;
         private readonly AccessTokenAlgorithm _algorithm;
 
@@ -38,10 +38,10 @@ namespace Microsoft.Azure.SignalR.AspNet
 
             // Version is ignored for aspnet signalr case
             _endpoint = endpoint.Endpoint;
-            _clientEndpoint = endpoint.ClientEndpoint ?? endpoint.Endpoint;
+            _serverEndpoint = endpoint.ServerEndpoint;
+            _clientEndpoint = endpoint.ClientEndpoint;
             _accessKey = endpoint.AccessKey;
             _appName = options.ApplicationName;
-            _port = endpoint.Port;
             _algorithm = options.AccessTokenAlgorithm;
 
             Proxy = options.Proxy;
@@ -55,7 +55,6 @@ namespace Microsoft.Azure.SignalR.AspNet
         public Task<string> GenerateClientAccessTokenAsync(string hubName = null, IEnumerable<Claim> claims = null, TimeSpan? lifetime = null)
         {
             var audience = $"{_endpoint}/{ClientPath}";
-
             return _accessKey.GenerateAccessTokenAsync(audience, claims, lifetime ?? _accessTokenLifetime, _algorithm);
         }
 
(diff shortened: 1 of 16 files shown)


src/Microsoft.Azure.SignalR/EndpointProvider/DefaultServiceEndpointGenerator.cs
```diff
@@ -33,7 +32,14 @@ public DefaultServiceEndpointGenerator(ServiceEndpoint endpoint)
 
         public string GetClientEndpoint(string hubName, string applicationName, string originalPath, string queryString)
         {
-            var queryBuilder = new StringBuilder();
+            var uriBuilder = new UriBuilder(ClientEndpoint)
+            {
+                Path = $"{ClientPath}/"
```
vicancy, line 37: directly set `ClientPath` to `client/`?

src/Microsoft.Azure.SignalR/EndpointProvider/DefaultServiceEndpointGenerator.cs
```diff
@@ -44,28 +50,29 @@ public string GetClientEndpoint(string hubName, string applicationName, string o
...
-        public string GetServerEndpoint(string hubName, string applicationName) =>
-            InternalGetEndpoint(ServerPath, hubName, applicationName, Endpoint);
-
-        private string GetPrefixedHubName(string applicationName, string hubName)
+        public string GetServerEndpoint(string hubName, string applicationName)
         {
-            return string.IsNullOrEmpty(applicationName) ? hubName.ToLower() : $"{applicationName.ToLower()}_{hubName.ToLower()}";
+            var uriBuilder = new UriBuilder(ServerEndpoint)
+            {
+                Path = $"{ServerPath}/",
```
vicancy, line 67: directly set ServerPath to `server/`?

test/Microsoft.Azure.SignalR.Common.Tests/ServiceEndpointFacts.cs
```diff
@@ -78,13 +78,15 @@ public class ServiceEndpointFacts
...
+        [InlineData("http://localhost:80", "http://localhost", 80)]
         [InlineData("https://localhost", "https://localhost", 443)]
+        [InlineData("https://localhost:443", "https://localhost", 443)]
         [InlineData("http://localhost:5050", "http://localhost", 5050)]
         [InlineData("https://localhost:5050", "https://localhost", 5050)]
         [InlineData("http://localhost/", "http://localhost", 80)]
         [InlineData("http://localhost/foo", "http://localhost", 80)]
         [InlineData("https://localhost/foo/", "https://localhost", 443)]
-        public void TestAadConstructor(string url, string endpoint, int port)
+        public void TestAadConstructor(string url, string endpoint, int? port)
```
vicancy, line 89: `int port` is enough?

test/Microsoft.Azure.SignalR.Common.Tests/ConnectionStringParserFacts.cs
```diff
@@ -66,14 +64,13 @@ public void ValidApplicationConnectionString(string expectedEndpoint, string con
...
             Assert.Null(version);
-            Assert.Null(accessKey.Port);
             Assert.Null(clientEndpoint);
         }
 
         [Theory]
-        [InlineData("https://aaa", "1.0", null, "endpoint=https://aaa;AccessKey=bbb;version=1.0")]
-        [InlineData("https://aaa", "1.0-preview", null, "ENDPOINT=https://aaa/;ACCESSKEY=bbb;VERSION=1.0-preview")]
-        [InlineData("http://aaa", "1.1", null, "endpoint=http://aaa;AccessKey=bbb;Version=1.1")]
+        [InlineData("https://aaa", "1.0", 443, "endpoint=https://aaa;AccessKey=bbb;version=1.0")]
```
vicancy, line 71: add some tests for when Endpoint contains port, e.g. `Endpoint=https://aaa:5040`?

src/Microsoft.Azure.SignalR.AspNet/EndpointProvider/ServiceEndpointProvider.cs
```diff
@@ -76,46 +75,41 @@ public Task<string> GenerateServerAccessTokenAsync(string hubName, string userId
...
-                else
+                if (queryBuilder.Length != 0)
                 {
                     queryBuilder.Append("&");
                 }
-
                 queryBuilder
                     .Append(Constants.QueryParameter.OriginalPath)
                     .Append("=")
                     .Append(WebUtility.UrlEncode(originalPath));
```
vicancy, line 99: shall we use `Uri.EscapeDataString` according to https://stackoverflow.com/a/37110034/687394?
cc @vwxyzh

vicancy: How about reusing `QueryBuilder` https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.extensions.querybuilder?view=aspnetcore-5.0

src/Microsoft.Azure.SignalR.AspNet/EndpointProvider/ServiceEndpointProvider.cs
```diff
@@ -76,46 +75,41 @@ public Task<string> GenerateServerAccessTokenAsync(string hubName, string userId
...
         public string GetClientEndpoint(string hubName = null, string originalPath = null, string queryString = null)
         {
-            var queryBuilder = new StringBuilder();
-
-            if (!string.IsNullOrEmpty(queryString))
+            var uriBuilder = new UriBuilder(_clientEndpoint)
             {
-                queryBuilder.Append(queryString);
-            }
+                Path = ClientPath
```
vicancy, line 85: so this is the only path with no `/`, is it intentional?

src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpoint.cs
```diff
@@ -17,6 +17,10 @@ public class ServiceEndpoint
 
         public string Endpoint => AccessKey?.Endpoint;
 
+        public string ServerEndpoint => Port.HasValue ? $"{Endpoint}:{Port}" : Endpoint;
```
vicancy, line 20: can port or Endpoint change during runtime? If not, set it when ctor instead of calculating it on every get. Same principle applies to other properties.

src/Microsoft.Azure.SignalR.Common/Endpoints/AccessKey.cs
```diff
@@ -14,20 +14,20 @@ internal class AccessKey
...
         {
             Key = new Tuple<string, string>(key.GetHashCode().ToString(), key);
         }
 
-        protected AccessKey(string endpoint, int? port)
+        protected AccessKey(Uri endpoint)
         {
-            Endpoint = endpoint;
-            Port = port;
+            Endpoint = $"{endpoint.Scheme}://{endpoint.Host}";
```
vicancy, line 29: what if endpoint contains subpath? If it is not supported, shall we do a pre-check and throw if it contains subpath? otherwise we silently drop some infomation

PR #1392, opened 2021-08-20
Add client/server endpoint property to `ServiceEndpoint`

terencefan:
 - Detail 1
 - Detail 2

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpoint.cs
+++ b/src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpoint.cs
@@ -17,8 +17,6 @@ namespace Microsoft.Azure.SignalR
 
         public string Endpoint => AccessKey?.Endpoint;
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpoint.cs
```diff
@@ -160,6 +167,11 @@ public override bool Equals(object obj)
             return Endpoint == that.Endpoint && EndpointType == that.EndpointType && Name == that.Name;
         }
 
+        private static string BuildServerEndpoint(AccessKey key)
+        {
+            return new Uri(key.Port.HasValue ? $"{key.Endpoint}:{key.Port}" : key.Endpoint).AbsoluteUri.ToString().TrimEnd('/');
```
vicancy, line 172: >ToString() [](http://example.com/codeflow?start=104&length=10)

no need

vicancy (approved):

PR #1395, opened 2021-08-20
refactor AccessKey constructor

terencefan:
1. Update `AccessKey` and `AadAccessKey` constructor to use `Uri endpoint` instead of using `string endpoint` & `int? port`

## This PR includes some breaking changes
1. `ServiceEndpoint.Endpoint` will convert its hostname into **lowercase** even when the given hostname is in uppercase. It will also influence the generated **accessToken** and **URL**. [RFC4343](https://datatracker.ietf.org/doc/html/rfc4343)
2. `ServiceEndpoint.Endpoint` will automatically add a port postfix if `port=xxx` was in the connection string and was not equal to 80 when HTTP or 443 when HTTPS.
3. According to 2, endpoints with the same hostname but different in ports will **NOT** be considered as equal.


src/Microsoft.Azure.SignalR.Common/Endpoints/AadAccessKey.cs
```diff
@@ -113,16 +118,17 @@ internal async Task UpdateAccessKeyAsync()
...
             var aadToken = await GenerateAadTokenAsync(ctoken);
-            await AuthorizeWithTokenAsync(Endpoint, Port, aadToken, ctoken);
+            await AuthorizeWithTokenAsync(aadToken, ctoken);
         }
 
-        private async Task AuthorizeWithTokenAsync(string endpoint, int? port, string accessToken, CancellationToken token = default)
+        private async Task AuthorizeWithTokenAsync(string accessToken, CancellationToken token = default)
         {
-            if (port != null && port != 443)
+            var builder = new UriBuilder(Endpoint)
```
vicancy, line 126: `builder.Uri.AbsoluteUri` this can move to a private field?

vicancy (approved):

src/Microsoft.Azure.SignalR.Common/Endpoints/AadAccessKey.cs
```diff
@@ -34,11 +34,17 @@ internal class AadAccessKey : AccessKey
 
         public TokenCredential TokenCredential { get; }
 
+        private string AuthorizeUrl { get; }
```
vicancy, line 37: private please use fields: `_authorizedUrl`

PR #1401, opened 2021-08-23
Fix negotiation test

terencefan:
Fix `TestRunAzureSignalRWithDefaultRouterNegotiateWithFallback`

Sometimes the test runs too quick, which will cause the `ConnectionInitializedTask` return immediately without service connection even tries to establish.


src/Microsoft.Azure.SignalR.AspNet/ServerConnections/ServiceConnectionManager.cs
```diff
@@ -70,6 +73,7 @@ public void Initialize(IServiceConnectionContainerFactory connectionFactory)
 
                 _hubConnections = connections;
             }
+            _managerInitializedTcs.TrySetResult(null);
```
vicancy, line 76: wrap in finally?

vicancy (approved):

PR #1406, opened 2021-08-24
Deprecate `IServiceManagerBuilder` 

Y-Sindo:


src/Microsoft.Azure.SignalR.Management/IServiceManagerBuilder.cs
```diff
@@ -1,17 +1,21 @@
...
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
+
 namespace Microsoft.Azure.SignalR.Management
 {
     /// <summary>
     /// A builder abstraction for configuring <see cref="IServiceManager"/> instances.
     /// </summary>
+    [Obsolete]
```
vicancy, line 11: add some comment here?

vicancy (approved):

PR #1418, opened 2021-08-27
Send serverId via http header.

terencefan:
`ServerId` is required while using our SDK in `server-sticky` mode. 

Well, with aad auth enabled, the server connection will use `AadToken` to connect to our service instead of `AccessToken` signed by our `AccessKey`, which does NOT contain the correct `UserIdentifier`.

To solve this, I'd like to move the `serverId` to the HTTP request header.

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ConnectionFactory.cs
+++ b/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ConnectionFactory.cs
@@ -14,12 +14,12 @@ namespace Microsoft.Azure.SignalR
     internal class ConnectionFactory : IConnectionFactory
     {
         private readonly ILoggerFactory _loggerFactory;
-        private readonly string _userId;
+        private readonly string _serverId;
 
         public ConnectionFactory(IServerNameProvider nameProvider, ILoggerFactory loggerFactory)
         {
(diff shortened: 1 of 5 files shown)


vicancy (approved):

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ConnectionFactory.cs
```diff
@@ -31,8 +31,15 @@ public ConnectionFactory(IServerNameProvider nameProvider, ILoggerFactory logger
...
             var provider = hubServiceEndpoint.Provider;
             var hubName = hubServiceEndpoint.Hub;
-            Task<string> accessTokenGenerater() => provider.GenerateServerAccessTokenAsync(hubName, _userId);
+            Task<string> accessTokenGenerater() => provider.GenerateServerAccessTokenAsync(hubName, _serverId);
             var url = GetServiceUrl(provider, hubName, connectionId, target);
+
+            headers ??= new Dictionary<string, string>();
+            if (!headers.ContainsKey(Constants.Headers.AsrsServerId))
+            {
+                headers.Add(Constants.Headers.AsrsServerId, _serverId);
```
vicancy, line 40: server null check

vicancy (approved):

PR #1424, opened 2021-09-03
Fix persistent mode still check endpoint health with single endpoint 

Y-Sindo:
* Add unit tests
Previously `IEndpointRouter` implementation `AutoHealthCheckRouter` is overridden by `DefaultEndpointRouter` introduced by ServerSDK

### DIFF
--- a/src/Microsoft.Azure.SignalR.Management/DependencyInjectionExtensions.cs
+++ b/src/Microsoft.Azure.SignalR.Management/DependencyInjectionExtensions.cs
@@ -77,10 +77,12 @@ namespace Microsoft.Azure.SignalR.Management
         {
             services.AddSingleton<IServiceManager, ServiceManagerImpl>();
             services.PostConfigure<ServiceManagerOptions>(o => o.ValidateOptions());
-            var tempServices = new ServiceCollection().AddSignalR()
-                .AddAzureSignalR<CascadeServiceOptionsSetup>().Services;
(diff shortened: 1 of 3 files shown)


src/Microsoft.Azure.SignalR.Management/DependencyInjectionExtensions.cs
```diff
@@ -77,10 +77,10 @@ private static IServiceCollection AddSignalRServiceCore(this IServiceCollection
         {
             services.AddSingleton<IServiceManager, ServiceManagerImpl>();
             services.PostConfigure<ServiceManagerOptions>(o => o.ValidateOptions());
-            services.TryAddSingleton<IEndpointRouter, AutoHealthCheckRouter>();
             var tempServices = new ServiceCollection().AddSignalR()
                 .AddAzureSignalR<CascadeServiceOptionsSetup>().Services;
             services.Add(tempServices.Where(service => service.ServiceType != typeof(IHostedService)));
+            services.AddSingleton<IEndpointRouter, AutoHealthCheckRouter>();
```
vicancy, line 83: this actually add two IEndpointRouter to the service? 

`AddAzureSignalR` is actually calling `TryAddSingleton`, so you can also try 
`new services => services.Add the one you want => services.AddAzureSignalR..... =>...`

vicancy (approved):

PR #1427, opened 2021-09-03
Make `ServiceHubContext` implements `IDisposable`

Y-Sindo:
* Make `ServiceHubContext` implements `IDisposable` 
  So that we can use `using var serviceHubContext = ...`
* Avoid disposing twice

* Make the class `sealed` to avoid dispose pattern

* Make `Dispose` virtual to allow properly implements dispose pattern in the non-sealed subclass.
  Our current sub class is sealed, but in case user want to implement non-sealed subclass.

### DIFF
--- a/src/Microsoft.Azure.SignalR.Management/ServiceHubContextImpl.cs
+++ b/src/Microsoft.Azure.SignalR.Management/ServiceHubContextImpl.cs
@@ -23,7 +23,7 @@ namespace Microsoft.Azure.SignalR.Management
         private readonly NegotiateProcessor _negotiateProcessor;
         private readonly IServiceEndpointManager _endpointManager;
         
-        private bool _disposed;
+        private bool _disposing;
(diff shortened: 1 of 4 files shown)


src/Microsoft.Azure.SignalR.Management/ServiceHubContextImpl.cs
```diff
@@ -54,8 +55,14 @@ public override ValueTask<NegotiationResponse> NegotiateAsync(NegotiationOptions
 
         public override async Task DisposeAsync()
         {
-            using var host = ServiceProvider.GetRequiredService<IHost>();
-            await host.StopAsync();
+            // Check _disposed to avoid disposing twice.
+            // When host is diposed, it will dispose all the disposable services including this class.
+            if (!_disposed)
```
vicancy, line 60: this is not thread-safe, do you want thread-safety for this call?

vicancy (approved):

PR #1429, opened 2021-09-06
Add more property to GroupBroadcastMessage

vicancy:
Provide the ability for group broadcast to exclude users, and also add "senderId" property to group broadcast for client pub-sub

### DIFF
--- a/src/Microsoft.Azure.SignalR.Protocols/MulticastDataMessage.cs
+++ b/src/Microsoft.Azure.SignalR.Protocols/MulticastDataMessage.cs
@@ -147,9 +147,9 @@ namespace Microsoft.Azure.SignalR.Protocol
         public IReadOnlyList<string> ExcludedUserList { get; set; }
 
         /// <summary>
-        /// Gets or sets the user ID of the sender
+        /// Gets or sets the user ID of the message caller
         /// </summary>
-        public string SenderId { get; set; }
+        public string CallerUserId { get; set; }
 
         /// <summary>
         /// Initializes a new instance of the <see cref="GroupBroadcastDataMessage"/> class.
--- a/src/Microsoft.Azure.SignalR.Protocols/MulticastDataMessage.cs
+++ b/src/Microsoft.Azure.SignalR.Protocols/MulticastDataMessage.cs
@@ -142,6 +142,16 @@ namespace Microsoft.Azure.SignalR.Protocol
         public IReadOnlyList<string> ExcludedList { get; set; }
 
         /// <summary>
+        /// Gets or sets the list of excluded user Ids.
+        /// </summary>
+        public IReadOnlyList<string> ExcludedUserList { get; set; }
+
+        /// <summary>
+        /// Gets or sets the user ID of the sender
+        /// </summary>
+        public string SenderId { get; set; }
+
+        /// <summary>
         /// Initializes a new instance of the <see cref="GroupBroadcastDataMessage"/> class.
         /// </summary>
         /// <param name="groupName">The group name.</param>
--- a/src/Microsoft.Azure.SignalR.Protocols/ServiceProtocol.cs
+++ b/src/Microsoft.Azure.SignalR.Protocols/ServiceProtocol.cs
@@ -450,12 +450,14 @@ namespace Microsoft.Azure.SignalR.Protocol
 
         private static void WriteGroupBroadcastDataMessage(ref MessagePackWriter writer, GroupBroadcastDataMessage message)
         {
-            writer.WriteArrayHeader(5);
+            writer.WriteArrayHeader(7);
             writer.Write(ServiceProtocolConstants.GroupBroadcastDataMessageType);
             writer.Write(message.GroupName);
             WriteStringArray(ref writer, message.ExcludedList);
             WritePayloads(ref writer, message.Payloads);
             message.WriteExtensionMembers(ref writer);
+            WriteStringArray(ref writer, message.ExcludedUserList);
+            writer.Write(message.SenderId);
         }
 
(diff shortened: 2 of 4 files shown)


src/Microsoft.Azure.SignalR.Protocols/ServiceProtocol.cs
```diff
@@ -869,6 +871,8 @@ private static GroupBroadcastDataMessage CreateGroupBroadcastDataMessage(ref Mes
             {
                 result.ReadExtensionMembers(ref reader);
             }
+            result.ExcludedUserList = ReadStringArray(ref reader, "excludedUserList");
```
vwxyzh, line 874: check `if (arrayLength >= 7)`

test/Microsoft.Azure.SignalR.Protocols.Tests/ServiceProtocolFacts.cs
```diff
@@ -497,15 +547,6 @@ public static IEnumerable<object[]> TestParseOldData
...
                 binary: "lBGmY29ubjExpmdyb3VwMoEBzQTS"),
-            new ProtocolTestData(
-                name: "GroupBroadcastWithTracingId",
-                message: new GroupBroadcastDataMessage("group3",
-                    new Dictionary<string, ReadOnlyMemory<byte>>
-                    {
-                        ["json"] = new byte[] {6, 7, 1, 2, 3, 4, 5},
-                        ["messagepack"] = new byte[] {7, 1, 2, 3, 4, 5, 6}
-                    }, tracingId: 1234L),
-                binary: "lQ2mZ3JvdXAzkIKkanNvbsQHBgcBAgMEBattZXNzYWdlcGFja8QHBwECAwQFBoEBzQTS"),
```
vwxyzh, line 508: why remove?

test/Microsoft.Azure.SignalR.Protocols.Tests/ServiceProtocolFacts.cs
```diff
@@ -402,7 +430,7 @@ public static IEnumerable<object[]> TestParseOldData
                         ["json"] = new byte[] {6, 7, 1, 2, 3, 4, 5},
                         ["messagepack"] = new byte[] {7, 1, 2, 3, 4, 5, 6}
                     }),
-                binary: "lQ2mZ3JvdXAzkIKkanNvbsQHBgcBAgMEBattZXNzYWdlcGFja8QHBwECAwQFBoA="),
+                binary: "lw2mZ3JvdXAzkIKkanNvbsQHBgcBAgMEBattZXNzYWdlcGFja8QHBwECAwQFBoCQwA=="),
```
vwxyzh, line 433: add the old one to `TestParseOldData`

vicancy: Already added to L253

src/Microsoft.Azure.SignalR.Protocols/MulticastDataMessage.cs
```diff
@@ -141,6 +141,16 @@ public class GroupBroadcastDataMessage : MulticastDataMessage
         /// </summary>
         public IReadOnlyList<string> ExcludedList { get; set; }
 
+        /// <summary>
+        /// Gets or sets the list of excluded user Ids.
+        /// </summary>
+        public IReadOnlyList<string> ExcludedUserList { get; set; }
```
vwxyzh, line 147: should we support exclude connection id and user id at the same time?
if not, should we add an ExcludeTarget (connection id/user id) instead?

test/Microsoft.Azure.SignalR.Protocols.Tests/ServiceProtocolFacts.cs
vicancy: reorder to merge with other `GroupBroadcastDataMessage` messages

src/Microsoft.Azure.SignalR.Protocols/ServiceProtocol.cs
vicancy: nice catch

src/Microsoft.Azure.SignalR.Protocols/MulticastDataMessage.cs
vicancy: I think we can support both at the same time

src/Microsoft.Azure.SignalR.Protocols/MulticastDataMessage.cs
```diff
@@ -141,6 +141,16 @@ public class GroupBroadcastDataMessage : MulticastDataMessage
...
+
+        /// <summary>
+        /// Gets or sets the user ID of the sender
+        /// </summary>
+        public string SenderId { get; set; }
```
JialinXin, line 152: CallerId?

vwxyzh (approved):

src/Microsoft.Azure.SignalR.Protocols/MulticastDataMessage.cs
vicancy: Since it is used in `SendToGroup`, does `sender` make more sense?

vicancy: caller sounds more general and can be applied to other messages

vicancy: since this is a `message`, I would prefer using `sender`.

JialinXin (approved):

PR #1449, opened 2021-09-13
add constructor for AzureKeyCredential

terencefan:
 - Detail 1
 - Detail 2

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpoint.cs
+++ b/src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpoint.cs
@@ -2,7 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
-
+using Azure;
 using Azure.Core;
 
 namespace Microsoft.Azure.SignalR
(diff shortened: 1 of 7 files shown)


src/Microsoft.Azure.SignalR.Common/Endpoints/ServiceEndpoint.cs
```diff
@@ -97,21 +97,20 @@ public ServiceEndpoint(string nameWithEndpointType, Uri endpoint, TokenCredentia
...
-                throw new ArgumentException("Endpoint scheme must be 'http://' or 'https://'");
-            }
             AccessKey = new AadAccessKey(endpoint, credential);
-            AudienceBaseUrl = BuildAudienceBaseUrl(endpoint);
-            Endpoint = BuildServerEndpoint(endpoint);
-            ClientEndpoint ??= Endpoint;
+            (Name, EndpointType) = (name, endpointType);
+        }
 
+        public ServiceEndpoint(string nameWithEndpointType, Uri endpoint, AzureKeyCredential credential) : this(endpoint, credential)
```
vicancy, line 106: In which case user pass in an AzureKeyCredential?

vicancy: supporting ConnectionString should be enough?

terencefan: I would like to follow the `WebPubSub` C# SDK. I though it may be a convention since it was developed by `Azure SDK Team`

vicancy: maybe not.. we are not following the azure sdk convention, and SignalR SDK is a little bit special that it is not only a REST client

PR #1451, opened 2021-09-13
reduce retry interval when auth failed

terencefan:
 - Detail 1
 - Detail 2


src/Microsoft.Azure.SignalR.Common/Endpoints/AadAccessKey.cs
```diff
@@ -88,45 +97,52 @@ public virtual async Task<string> GenerateAadTokenAsync(CancellationToken ctoken
...
-            _isAuthorized = true;
-            _initializedTcs.TrySetResult(null);
+            Authorized = true;
         }
 
-        internal async Task UpdateAccessKeyAsync()
+        internal async Task UpdateAccessKeyAsync(CancellationToken token = default)
         {
-            if (DateTime.UtcNow - _lastUpdatedTime < AuthorizeInterval)
+            var delta = DateTime.UtcNow - _lastUpdatedTime;
```
vicancy, line 105: is it possible to be called by multiple threads?

terencefan: No, there is only one `AccessKeySynchronizer` who will handle all the updates.

terencefan: And this method is `idempotent`.
It will work even if the method had been called twice or more times at once.

vicancy (approved):

PR #1453, opened 2021-09-14
add code cleanup docs

terencefan:
 - Detail 1
 - Detail 2


vicancy:
Thanks for the effort. How about adding a contribution guidance in the repo, as similar to this one https://github.com/Azure/azure-sdk-for-net/blob/main/CONTRIBUTING.md , content here can be in the section of "How to setup develop environment"?

terencefan:
@vicancy I think we can have this simple doc merged first and let everyone have a try on these tools.
Then decide if we could crystalize the convention into laws.

src/Microsoft.Azure.SignalR.Common/Endpoints/HubServiceEndpoint.cs
```diff
@@ -40,11 +25,25 @@ internal class HubServiceEndpoint : ServiceEndpoint
         /// </summary>
         public Task ScaleTask => _scaleTcs?.Task ?? Task.CompletedTask;
 
+        public long UniqueIndex => _uniqueIndex;
+
+        public HubServiceEndpoint(
+                                                            string hub,
```
vicancy, line 31: a strange indent

vicancy (approved):

terencefan:
Close since no activity

PR #1537, opened 2022-01-14
Enable source link and symbol package

Y-Sindo:
Improve cross-project debug experience. Currently if we want to debug SDK in other projects, we have to reference the local SDK project. With source link or symbol package we don't have to do that.

As the new symbol package format `*.snupkg` is not supported by KoreBuild, we cannot enable symbol package yet.
[Symbol packages](https://docs.microsoft.com/dotnet/standard/library-guidance/nuget#symbol-packages)
[Source link](https://docs.microsoft.com/dotnet/standard/library-guidance/sourcelink)


vicancy (approved):
Pretty cool!

PR #1542, opened 2022-01-21
Support implicit conversion from `ServiceHubContextImpl<T>` to `ServiceHubContextImpl`.

Y-Sindo:
In SignalR Functions extensions, `ServerlessHub<T>` is suggested being derived from `ServerlessHub`, which requires that we can get a `ServiceHubContext` out of a `ServiceHubContext<T>`. 

Why don't we make `ServiceHubContext<T>` derived from `ServiceHubContext`? 
I am not sure whether we should expose the inheritance to users. Currently it's only required for Functions extensions.

* Refactor: `ServiceHubContext` and `ServiceHubContext<T>` are no more injected, so that they won't get disposed twice when `IServiceProvider` is disposed.
Consider this usage:
```
using ServiceHubContext<T> c = serviceManager.CreateHubContextAsync<T>("hubName",cancellationToken);
using ServiceHubContext c1 = c as ServiceHubContextImpl<T>;
```


Y-Sindo:
Close as we don't need it now.

PR #1557, opened 2022-02-15
Indicating persistent transport type in the emulator's document

Arash-Sabet:
 - SignalR Emulator document updated to reflect the lack of persistent transport type

### DIFF
--- a/docs/emulator.md
+++ b/docs/emulator.md
@@ -70,3 +70,6 @@ The emulator also provides advanced options when start, for example, you can use
 7. In the browser, navigate to `http://localhost:7071/api/index` to play with the demo.
 
 ![image](https://user-images.githubusercontent.com/668244/93066361-78d11e80-f6ac-11ea-8b2e-bff28196850c.png)
+
+## Supporting persistent transport types
+This emualtor is unable to support persistent transport types i.e. `"AzureSignalRServiceTransportType": "Persistent"`. Therefore, tools like Azure Relay or ngrok are required to divert the webhook traffic from Azure SignalR Services to the local development machines.


Y-Sindo:
Thanks for contributing. I happened to add another PR #1556 to mention that emulator not working for `persistent` mode, but I failed to link it to the issue.

docs/emulator.md
```diff
@@ -70,3 +70,6 @@ The emulator also provides advanced options when start, for example, you can use
 7. In the browser, navigate to `http://localhost:7071/api/index` to play with the demo.
 
 ![image](https://user-images.githubusercontent.com/668244/93066361-78d11e80-f6ac-11ea-8b2e-bff28196850c.png)
+
+## Supporting persistent transport types
```
Y-Sindo, line 74: The suggestion on persistent mode local debug is missing in the latest doc. How about combining this part with the first paragraph in the latest doc?
> When developing serverless applications, we provide an Azure SignalR Local Emulator to make the local development and integration easier. Please note that the emulator works only for serverless scenarios, for Default mode that the Azure SignalR Service acts as a proxy, you can directly use self-host SignalR to do local development. Please also note that emulator only works for Transient transport type (the default one) and doesn't work for Persistent transport type.

Arash-Sabet: Seems good to me.

Arash-Sabet: Sounds good to me.

docs/emulator.md
```diff
@@ -70,3 +70,6 @@ The emulator also provides advanced options when start, for example, you can use
...
+This emualtor is unable to support persistent transport types i.e. `"AzureSignalRServiceTransportType": "Persistent"`. Therefore, tools like Azure Relay or ngrok are required to divert the webhook traffic from Azure SignalR Services to the local development machines.
```
vicancy, line 75: Since the development experience for using Transient mode and Persistent mode should be the same, maybe another option is to switch `AzureSignalRServiceTransportType` to transient for local development?

vicancy:
Content added.

PR #1567, opened 2022-02-23
add server connection and connection quota for endpoint metrics

bjqian:
When using multible endpoint, this metrics will enable customer to route request according to the instance weight.


vwxyzh (approved):

src/Microsoft.Azure.SignalR.Common/ServiceMessages/RuntimeServicePingMessages.cs
```diff
@@ -97,6 +99,28 @@ public static bool TryGetClientCount(this ServicePingMessage ping, out int clien
             }
             clientCount = count;
             return true;
+        }   
+        
+        public static bool TryGetServerCount(this ServicePingMessage ping, out int serverCount)
```
JialinXin, line 104: Naming to `TryGetServerConnectionCount()` would be more accurate, since the another `servers` ping is trying to get server count...

PR #1568, opened 2022-02-24
Biqian/weighted router

bjqian:
change the default router from random to weighted random

### DIFF
--- a/src/Microsoft.Azure.SignalR/EndpointRouters/DefaultEndpointRouter.cs
+++ b/src/Microsoft.Azure.SignalR/EndpointRouters/DefaultEndpointRouter.cs
@@ -47,6 +47,11 @@ namespace Microsoft.Azure.SignalR
             return secondary;
         }
 
+        /// <summary>
+        ///  Choose endpoint randomly by weight. 
+        ///  The weight is defined as the remaining connection quota.
+        ///  The least weight is set to 1. So instance with no connection quota still has chance.
+        /// </summary>
         private ServiceEndpoint GetEndpointAccordingToWeight(ServiceEndpoint[] availableEndpoints)
         {
             //first check if weight is available or necessary
@@ -70,12 +75,18 @@ namespace Microsoft.Azure.SignalR
             }
 
             var index = StaticRandom.Next(totalCapacity);
-            var key = Array.BinarySearch(we, index);
-            key = key < 0 ? ~key : key + 1;
-            return availableEndpoints[key];
+            for (var i = 0; i < we.Length; i++)
+            {
+                if (we[i] > index)
+                {
+                    return availableEndpoints[i];
+                }
+            }
+
+            return GetEndpointRandomly(availableEndpoints);
         }
 
-        private ServiceEndpoint GetEndpointRandomly(ServiceEndpoint[] availableEndpoints)
+        private static ServiceEndpoint GetEndpointRandomly(ServiceEndpoint[] availableEndpoints)
         {
             return availableEndpoints[StaticRandom.Next(availableEndpoints.Length)];
         }
--- a/src/Microsoft.Azure.SignalR/EndpointRouters/DefaultEndpointRouter.cs
+++ b/src/Microsoft.Azure.SignalR/EndpointRouters/DefaultEndpointRouter.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using Microsoft.AspNetCore.Http;
@@ -19,8 +20,8 @@ namespace Microsoft.Azure.SignalR
         public ServiceEndpoint GetNegotiateEndpoint(HttpContext context, IEnumerable<ServiceEndpoint> endpoints)
         {
             // get primary endpoints snapshot
-            var availbaleEndpoints = GetNegotiateEndpoints(endpoints);
-            return availbaleEndpoints[StaticRandom.Next(availbaleEndpoints.Length)];
+            var availableEndpoints = GetNegotiateEndpoints(endpoints);
+            return GetEndpointAccordingToWeight(availableEndpoints);
         }
 
         /// <summary>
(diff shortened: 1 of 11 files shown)


src/Microsoft.Azure.SignalR/EndpointRouters/DefaultEndpointRouter.cs
```diff
@@ -45,5 +46,38 @@ private ServiceEndpoint[] GetNegotiateEndpoints(IEnumerable<ServiceEndpoint> end
...
+                var remain = endpointMetrics.ConnectionCapacity -
+                             (endpointMetrics.ClientConnectionCount +
+                              endpointMetrics.ServerConnectionCount);
+                var weight = remain > 0 ? remain : 1;
+                totalCapacity += weight;
+                we[i] = totalCapacity;
+            }
+
+            var index = StaticRandom.Next(totalCapacity);
+            var key = Array.BinarySearch(we, index);
```
vwxyzh, line 73: there is no need to use `BinarySearch`, in normal case the `availableEndpoints` should be than 10.

src/Microsoft.Azure.SignalR/EndpointRouters/DefaultEndpointRouter.cs
```diff
@@ -45,5 +46,38 @@ private ServiceEndpoint[] GetNegotiateEndpoints(IEnumerable<ServiceEndpoint> end
...
+
+            var we = new int[availableEndpoints.Length];
+            var totalCapacity = 0;
+            for (var i = 0; i < availableEndpoints.Length; i++)
+            {
+                var endpointMetrics = availableEndpoints[i].EndpointMetrics;
+                var remain = endpointMetrics.ConnectionCapacity -
+                             (endpointMetrics.ClientConnectionCount +
+                              endpointMetrics.ServerConnectionCount);
+                var weight = remain > 0 ? remain : 1;
```
vwxyzh, line 67: if there is a instance with no remain, we should not route clients to that instance.

src/Microsoft.Azure.SignalR/EndpointRouters/DefaultEndpointRouter.cs
bjqian: Change to a for loop

src/Microsoft.Azure.SignalR/EndpointRouters/DefaultEndpointRouter.cs
bjqian: Keep it 1 for simplicity

src/Microsoft.Azure.SignalR/EndpointRouters/DefaultEndpointRouter.cs
```diff
@@ -45,5 +46,49 @@ private ServiceEndpoint[] GetNegotiateEndpoints(IEnumerable<ServiceEndpoint> end
...
+
+        /// <summary>
+        ///  Choose endpoint randomly by weight. 
+        ///  The weight is defined as the remaining connection quota.
+        ///  The least weight is set to 1. So instance with no connection quota still has chance.
+        /// </summary>
+        private ServiceEndpoint GetEndpointAccordingToWeight(ServiceEndpoint[] availableEndpoints)
+        {
+            //first check if weight is available or necessary
+            if (availableEndpoints.Any(endpoint => endpoint.EndpointMetrics.ConnectionCapacity == 0) ||
```
JialinXin, line 58: Does `Capacity==0` mean status is not initiated? Should we random pick only when `All` instead of `Any` are not initiated.

src/Microsoft.Azure.SignalR/EndpointRouters/DefaultEndpointRouter.cs
```diff
@@ -19,8 +20,8 @@ internal class DefaultEndpointRouter : DefaultMessageRouter, IEndpointRouter
         public ServiceEndpoint GetNegotiateEndpoint(HttpContext context, IEnumerable<ServiceEndpoint> endpoints)
```
JialinXin, line 20: Shall we add UT to cover the method as now complicated logic introduced here?

src/Microsoft.Azure.SignalR/EndpointRouters/DefaultEndpointRouter.cs
```diff
@@ -45,5 +46,49 @@ private ServiceEndpoint[] GetNegotiateEndpoints(IEnumerable<ServiceEndpoint> end
...
+            }
+
+            var index = StaticRandom.Next(totalCapacity);
+            for (var i = 0; i < we.Length; i++)
+            {
+                if (we[i] > index)
+                {
+                    return availableEndpoints[i];
+                }
+            }
```
JialinXin, line 84: `return available[Array.FindLastIndex(we, x => x < index) + 1]`?

src/Microsoft.Azure.SignalR/EndpointRouters/DefaultEndpointRouter.cs
JialinXin: Exclude `remain<=0` ones from `GetNegotiateEndpoints(endpoints)` method? So they won't be in the list of available endpoints as they're overload.

bjqian: if remain=1 and others have capacity like 10k, then this instance will have very low chance of being connected.
Completely removing that will be a breaking change as in fact we could use 10% more quota. 
Also if all instances have used up the quota, they will all have weight 1. The traffic will be equally distributed rather than throwing an exception. I try to keep the behavious same as before.

src/Microsoft.Azure.SignalR/EndpointRouters/DefaultEndpointRouter.cs
bjqian: Once runtime is up, it's inited.  Capacity==0 happens when it's free or it's old version runtime.

JialinXin: Why free doesn't have capacity? Even 20 is a valid number.
And should a free one break the weight algorithm always?

bjqian: The Capacity might be 0 when no ping has been received. That happens when server starts up. Since no ping for some time will lead to a connection drop. That will be a short time. Using `any` just to explicitly show some backward compatity though removing that won't have difference.

src/Microsoft.Azure.SignalR/EndpointRouters/DefaultEndpointRouter.cs
bjqian: simplified

src/Microsoft.Azure.SignalR/EndpointRouters/DefaultEndpointRouter.cs
bjqian: Add ut EndpointRouterTest

JialinXin (approved):

PR #1582, opened 2022-04-01
Try fix the logic for single endpoint and ackable message in multiple…

vicancy:
1. When all the endpoints fail, we throw, otherwise we log? --- this makes single/multi endpoint behavior consistent
2. We always wait until all endpoints return or throw
3. Whenever an endpoint returns true, we believe the result is true?

-- endpoint1: true
-- endpoint2: false
-- endpoint3: throw Timeout / ServerConnectionNotActive

Connection related actions: return true.
User/Group related actions: throw?

-- endpoint1: false
-- endpoint2: throw Timeout/ ServerConnectionNotActive
User/Group/Connection related actions: throw?


src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointMessageWriter.cs
```diff
@@ -52,11 +53,55 @@ public Task WriteAsync(ServiceMessage serviceMessage)
...
+
+            return dict.All(i => i);
+        }
+
+        /// <summary>
+        /// For connection related operations, since connectionId is globally unique, only one endpoint can have the connection
+        /// Strategy:
+        /// Don't need to wait until all endpoints return or throw
+        /// 1. Whenever any endpoint returns true: return true
+        /// 2. When all endpoints return false, return false
```
zackliu, line 97: Update the summary to match the implementation. The second step is to throw ex if any endpoint throw. And the third part is return false

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointMessageWriter.cs
```diff
@@ -52,11 +53,55 @@ public Task WriteAsync(ServiceMessage serviceMessage)
...
+            {
+                return WriteMultiResultAckableMessage(serviceMessage, cancellationToken);
+            }
+        }
+
+        /// <summary>
+        /// For user or group related operations, different endpoints might return different results
+        /// Strategy:
+        /// Always wait until all endpoints return or throw
+        /// 1. When any endpoint returns true, return true
```
zackliu, line 72: The step is 1. throw if any throw 2. false if any false 3. true if all ture ?

zackliu (approved):

PR #1612, opened 2022-05-19
Revert "Rever ServiceEndpoint.Name setting to unblock func users."

JialinXin:
Reverts Azure/azure-signalr#1611

revert squash


JialinXin:
Looks some newly added security check blocks. Use another way to fix.

PR #1616, opened 2022-05-20
Add type=azure for connection string.

terencefan:
For historical reasons, we will use `ManagedIdentityCredential` by default when using `type=aad` with no other designated params. It is NOT a best practice with the latest `msal` library anymore.

We added `type=azure` in this PR to change the default behavior, the old `type=aad` will be deprecated and removed in future releases.

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/Utilities/ConnectionStringParser.cs
+++ b/src/Microsoft.Azure.SignalR.Common/Utilities/ConnectionStringParser.cs
@@ -16,7 +16,7 @@ namespace Microsoft.Azure.SignalR
 
         private const string ClientCertProperty = "clientCert";
 
-        private const string ClientEndpointProperty = "ClientEndpoint";
+        private const string ClientEndpointProperty = "clientEndpoint";
 
         private const string ClientIdProperty = "clientId";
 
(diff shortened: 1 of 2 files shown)


docs/connection-string-guide.md
```diff
@@ -0,0 +1,58 @@
+# Connection string guide for SignalR server SDK
```
vicancy, line 1: since we are in a progress of moving docs into docs.microsoft.com repo, could we add this article there directly?

vicancy: this doc talks about connection string https://docs.microsoft.com/en-us/azure/azure-signalr/concept-connection-string

src/Microsoft.Azure.SignalR.Common/Utilities/ConnectionStringParser.cs
```diff
@@ -144,6 +177,8 @@ internal static bool TryGetEndpointUri(string endpoint, out Uri uriResult)
 
         private static AccessKey BuildAadAccessKey(Uri uri, Dictionary<string, string> dict)
         {
+            Console.WriteLine("Connection string property type=aad is deprecated, please use type=AzureAD instead. See https://github.com/Azure/azure-signalr/blob/dev/docs/connection-string-guide.md for more info.");
```
vicancy, line 180: no `Console`, use `logger` instead

src/Microsoft.Azure.SignalR.Common/Utilities/ConnectionStringParser.cs
```diff
@@ -180,5 +215,35 @@ private static AccessKey BuildAccessKey(Uri uri, Dictionary<string, string> dict
             }
             throw new ArgumentException(MissingAccessKeyProperty, AccessKeyProperty);
         }
+
+        private static AccessKey BuildAzureAdAccessKey(Uri uri, Dictionary<string, string> dict)
+        {
+            if (dict.TryGetValue(ClientIdProperty, out var clientId))
```
vicancy, line 221: extract common code with BuildAadAccessKey?

docs/connection-string-guide.md
terencefan: I'd like to but there it takes time to publish a doc there. Maybe we could have this merged in advance.

src/Microsoft.Azure.SignalR.Common/Utilities/ConnectionStringParser.cs
terencefan: Not a good idea since they're legacy codes.

src/Microsoft.Azure.SignalR.Common/Utilities/ConnectionStringParser.cs
```diff
@@ -11,22 +11,43 @@ namespace Microsoft.Azure.SignalR
...
         private const string PortProperty = "port";
+
+        private const string ServerEndpoint = "ServerEndpoint";
+
         // For SDK 1.x, only support Azure SignalR Service 1.x
         private const string SupportedVersion = "1";
 
         private const string TenantIdProperty = "tenantId";
+
+        private const string TypeAzureAd = "azuread";
```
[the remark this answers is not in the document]
terencefan: Good idea

src/Microsoft.Azure.SignalR.Common/Utilities/ConnectionStringParser.cs
```diff
@@ -164,11 +167,11 @@ private static AccessKey BuildAadAccessKey(Uri uri, Dictionary<string, string> d
                 {
                     if (dict.TryGetValue(ClientSecretProperty, out var clientSecret))
                     {
-                        return new AadAccessKey(uri, new ClientSecretCredential(tenantId, clientId, clientSecret));
+                        return new AadAccessKey(uri, new ClientSecretCredential(tenantId, clientId, clientSecret), true);
```
vicancy, line 170: don't see AadAccessKey ctor change to support `true`, what is it?

terencefan: failed to revert the patch when rebasing, fixed it

vicancy (approved):

PR #1637, opened 2022-06-29
Remove `InternalsVisibleTo` Functions extensions

Y-Sindo:
`InternalsVisibleTo` Functions extensions turns out to be problematic because it is easy to remove an internal interface depended by functions extensions and causes breaking change. As we thing the `WithEndpoints` interface is stable enough and we can make it public.

This change itself is breaking, but I think it is worth breaking once and then we can avoid the problem forever.


JialinXin (approved):

src/Microsoft.Azure.SignalR.Management/DependencyInjectionExtensions.cs
```diff
@@ -108,13 +108,25 @@ private static IServiceCollection AddSignalRServiceCore(this IServiceCollection
...
+        /// Allows functions extensions to add additional product info.
+        /// </summary>
+        public static IServiceCollection AddUserAgent(this IServiceCollection services, string userAgent)
+        {
+            if (userAgent is null)
+            {
+                throw new ArgumentNullException(nameof(userAgent));
+            }
+
+            return services.PostConfigure<ServiceManagerOptions>(o => o.ProductInfo += userAgent);
```
vicancy, line 127: what if user call WithUserAgent(..).WithAssembly()?

Y-Sindo: The post configure actions are always executed after configure actions.

vicancy: throw if ProductInfo is null?

Y-Sindo: The `AddSignalRServiceCore` makes sure `ProductInfo` is not null. Will update to "string. Empty + userAgent" if product is null.

Y-Sindo: I think throwing if null is better now as it can serve as a validation as well.

PR #1639, opened 2022-07-01
Update net50 samples to use 50 client version

vicancy:
Clients to use 5.0 version client libraries too


zackliu (approved):

samples/ChatSample/ChatSample.Net50/wwwroot/index.html
```diff
@@ -177,13 +174,31 @@ <h2 class="text-center" style="margin-top: 0; padding-top: 30px; padding-bottom:
...
-                })
-                .catch(function (error) {
-                    console.warn(error);
+            function delay(time) {
+                return new Promise((resolve) => {
+                    setTimeout(() => resolve(), time);
                 });
+            }
+            
+            function retry(attempts) {
```
chenkennt, line 183: Why not just use a loop since you're already using async.

chenkennt (approved):

PR #1659, opened 2022-08-18
Hide 401 exceptions when the old key is still available

terencefan:
 - Detail 1
 - Detail 2

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/Endpoints/AadAccessKey.cs
+++ b/src/Microsoft.Azure.SignalR.Common/Endpoints/AadAccessKey.cs
@@ -17,14 +17,19 @@ namespace Microsoft.Azure.SignalR
     internal class AadAccessKey : AccessKey
     {
         internal const int AuthorizeIntervalInMinute = 55;
+
         internal const int AuthorizeMaxRetryTimes = 3;
+
         internal const int AuthorizeRetryIntervalInSec = 3;
 
         private const string DefaultScope = "https://signalr.azure.com/.default";
 
         private static readonly TokenRequestContext _defaultRequestContext = new TokenRequestContext(new string[] { DefaultScope });
+
         private static readonly TimeSpan AuthorizeInterval = TimeSpan.FromMinutes(AuthorizeIntervalInMinute);
+
         private static readonly TimeSpan AuthorizeRetryInterval = TimeSpan.FromSeconds(AuthorizeRetryIntervalInSec);
+
         private static readonly TimeSpan AuthorizeTimeout = TimeSpan.FromSeconds(10);
 
         private readonly TaskCompletionSource<object> _initializedTcs = new TaskCompletionSource<object>(TaskCreationOptions.RunContinuationsAsynchronously);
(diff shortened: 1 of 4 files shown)


src/Microsoft.Azure.SignalR.Common/Endpoints/AadAccessKey.cs
```diff
@@ -93,6 +98,17 @@ internal void UpdateAccessKey(string kid, string accessKey)
             _initializedTcs.TrySetResult(null);
         }
 
+        /// <summary>
+        /// For unit test only
+        /// </summary>
+        /// <param name="datetime"></param>
+        internal void UpdateLastUpdatedTime(DateTime datetime)
```
vicancy, line 105: don't design a method only for test

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -314,18 +324,18 @@ private Task OnEventMessageAsync(ServiceEventMessage message)
...
                 }
-            }
-            else
-            {
-                var exception = new AzureSignalRUnauthorizedException(new AzureSignalRException(keyMessage.ErrorMessage));
-                Log.AuthorizeFailed(Logger, exception);
-                return Task.CompletedTask;
+                else if (key.HasExpired)
+                {
+                    var exception = new AzureSignalRUnauthorizedException(new AzureSignalRException(keyMessage.ErrorMessage));
```
vicancy, line 335: we don't need to create an exception only for logging purposes. How about improving Log to log the exception message?

src/Microsoft.Azure.SignalR.Common/Endpoints/AadAccessKey.cs
```diff
@@ -139,14 +146,7 @@ private async Task AuthorizeWithTokenAsync(string accessToken, CancellationToken
...
 
-        private async Task<bool> HandleHttpResponseAsync(HttpResponseMessage response)
-        {
-            if (await HandleHttpResponseAsyncCore(response))
-            {
-                return true;
-            }
-            return false;
-        }
+        private async Task<bool> HandleHttpResponseAsync(HttpResponseMessage response) => await HandleHttpResponseAsyncCore(response);
```
vicancy, line 149: async and await can be removed

test/Microsoft.Azure.SignalR.Tests/ServiceMessageTests.cs
```diff
@@ -218,17 +218,32 @@ public async Task TestAccessKeyResponseMessage(Type keyType)
...
+        public async Task TestAccessKeyResponseMessageWithError(int minutesElapsed, int expectedLogCount)
         {
             using (StartVerifiableLog(out var loggerFactory, LogLevel.Error, expectedErrors: c => true,
                 logChecker: logs =>
                 {
-                    Assert.Equal(1, logs.Count);
+                    Assert.Equal(expectedLogCount, logs.Count);
+                    if (logs.Count > 0)
+                    {
+                        logs[0].ToString().StartsWith("Service returned 401 unauthorized.");
```
vicancy, line 233: looks like `.` is updated to `:`?

vicancy (approved):

PR #1663, opened 2022-08-29
Update Actions to use dotnet build

vicancy:


Y-Sindo:
I just checked source code and found that `Newtonsoft.Json` is not an override, but exactly the same import. So it's safe to delete it.....

Y-Sindo (approved):

Issue #1660, opened 2022-08-19
Microsoft.Azure.SignalR.Common.AzureSignalRException: Ack-able message Microsoft.Azure.SignalR.Protocol.JoinGroupWithAckMessage gets error ack status 4


SergeyOsipenko:
### Describe the bug
When adding a user to several groups I get this error:

System.AggregateException: One or more errors occurred. (Ack-able message Microsoft.Azure.SignalR.Protocol.JoinGroupWithAckMessage gets error ack status 4.)
 ---> Microsoft.Azure.SignalR.Common.AzureSignalRException: Ack-able message Microsoft.Azure.SignalR.Protocol.JoinGroupWithAckMessage **gets error ack status 4**.
   at Microsoft.Azure.SignalR.ServiceConnectionContainerBase.WriteAckableMessageAsync(ServiceMessage serviceMessage, CancellationToken cancellationToken)
   at Microsoft.Azure.SignalR.MultiEndpointMessageWriter.<>c__DisplayClass10_0.<b__0>d.MoveNext()
--- End of stack trace from previous location ---
   at Microsoft.Azure.SignalR.MultiEndpointMessageWriter.WriteSingleEndpointMessageAsync(HubServiceEndpoint endpoint, ServiceMessage serviceMessage, Func`2 inner)
   at Microsoft.Azure.SignalR.MultiEndpointMessageWriter.WriteMultiEndpointMessageAsync(ServiceMessage serviceMessage, Func`2 inner)
   --- End of inner exception stack trace ---
   at Microsoft.Azure.SignalR.MultiEndpointMessageWriter.WriteMultiEndpointMessageAsync(ServiceMessage serviceMessage, Func`2 inner)
   at Microsoft.Azure.SignalR.MultiEndpointMessageWriter.WriteSingleResultAckableMessage(ServiceMessage serviceMessage, CancellationToken cancellationToken)
   at Microsoft.Azure.SignalR.ServiceLifetimeManagerBase`1.WriteAckableCoreAsync[T](T message, Func`2 task)

### To Reproduce

call AddToGroupAsync several times for the user

```
foreach(string roomId in roomIDs)
{
     await Groups.AddToGroupAsync(Context.ConnectionId, roomId ); 
}
```
It happens time to time and with any server load

### Further technical details
- Microsoft.Azure.SignalR 1.18.1
- .Net 6
- No SDK version for client. Pure WebSocket protocol is used to get access to the server and send messages\events

[...]

PR #1669, opened 2022-09-05
Add InternalServerError to retryable error

vicancy:
To fix https://github.com/Azure/azure-signalr/issues/1660 from SDK side

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
+++ b/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
@@ -255,6 +255,7 @@ namespace Microsoft.Azure.SignalR
                 case AckStatus.NotFound:
                     return false;
                 case AckStatus.Timeout:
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -255,6 +255,7 @@ public async Task<bool> WriteAckableMessageAsync(ServiceMessage serviceMessage,
                 case AckStatus.NotFound:
                     return false;
                 case AckStatus.Timeout:
+                case AckStatus.InternalServerError:
```
zackliu, line 258: Why it's throwing TimeoutException?

vicancy: From customer's perspective it is the same as Timeout

terencefan (approved):

zackliu (approved):

PR #1677, opened 2022-09-06
Update CI to enable net7.0 preview builds

JialinXin:
 - Detail 1
 - Detail 2

### DIFF
--- a/test/Microsoft.Azure.SignalR.Tests/Microsoft.Azure.SignalR.Tests.csproj
+++ b/test/Microsoft.Azure.SignalR.Tests/Microsoft.Azure.SignalR.Tests.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net5.0</TargetFramework>
+    <TargetFramework>net7.0</TargetFramework>
     <IsPackable>false</IsPackable>
     <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
     <DefineConstants>MULTIFRAMEWORK</DefineConstants>
--- a/test/Microsoft.Azure.SignalR.Tests/MultiEndpointServiceConnectionContainerTests.cs
+++ b/test/Microsoft.Azure.SignalR.Tests/MultiEndpointServiceConnectionContainerTests.cs
@@ -1050,8 +1050,8 @@ namespace Microsoft.Azure.SignalR.Tests
             var serversTag = "Server1;Server2;Server3";
             await Task.WhenAll(containers.Select(c => c.MockReceivedServersPing(serversTag)));
 
-            // wait one interval+ for Ready state and check negotiation is added.
-            await Task.Delay(6000);
+            await hubEndpoints[1].ScaleTask;
+            await Task.Delay(100);
 
             ngoEps = sem.GetEndpoints("hub").OrderBy(x => x.Name).ToArray();
             Assert.Equal(2, ngoEps.Length);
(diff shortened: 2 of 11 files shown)


test/Microsoft.Azure.SignalR.Tests/MultiEndpointServiceConnectionContainerTests.cs
```diff
@@ -1347,7 +1350,8 @@ public async Task TestEndpointManagerWithMultiHubsWithServerEndpointUpdate()
             await Task.WhenAll(containers.Select(x => x.MockReceivedServersPing("aaa;bbb")));
             containers1 = container1.GetTestOnlineContainers();
             await Task.WhenAll(containers1.Select(x => x.MockReceivedServersPing("aaa;bbb")));
-            await Task.Delay(6000);
+            await hubEndpoints.Single(x => x.Name == "1" && x.ServerEndpoint.AbsoluteUri == testSe).ScaleTask;
+            await Task.Delay(100);
```
vicancy, line 1354: Is there still need to delay?

test/Microsoft.Azure.SignalR.Tests/Microsoft.Azure.SignalR.Tests.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net5.0</TargetFramework>
+    <TargetFramework>net7.0</TargetFramework>
```
vicancy, line 4: hm.. I think we might need to separate net7 support to another  PR or maybe in the same PR supporting client invocation, we need to make sure net7 features like client invocation are supported before we claim that the package now supports net7

test/Microsoft.Azure.SignalR.Tests/MultiEndpointServiceConnectionContainerTests.cs
JialinXin: Yes. Little delay required as there's a few operations after the `ScaleTask` completes which is to sync the endpoints changes in the other side. So minor delay is preferred to ensure the validation success.

test/Microsoft.Azure.SignalR.Tests/Microsoft.Azure.SignalR.Tests.csproj
JialinXin: OK.

test/Microsoft.Azure.SignalR.Tests/MultiEndpointServiceConnectionContainerTests.cs
```diff
@@ -1050,8 +1050,8 @@ public async Task TestEndpointManagerWithAddEndpointsWithServersTag()
             var serversTag = "Server1;Server2;Server3";
             await Task.WhenAll(containers.Select(c => c.MockReceivedServersPing(serversTag)));
 
-            // wait one interval+ for Ready state and check negotiation is added.
-            await Task.Delay(6000);
+            await hubEndpoints[1].ScaleTask;
```
vicancy, line 1053: await hubEndpoints[1].ScaleTask.OrTimeout(10000) ?

JialinXin: `ScaleTask` is completed, but other operations need a few moment delay before negotiation endpoints get updated. I think `OrTimeout()` means the longest time that should be await when target task doesn't complete as expected.

In normal case it takes about 5.2s of an interval. Do you mean add `OrTimeout()` to make it fail fast in error case?

vicancy (approved):

PR #1686, opened 2022-09-23
Add filter property......

vicancy:
Potentially we can use `filter` to filter out subscribers in signalr too...

### DIFF
--- a/src/Microsoft.Azure.SignalR.Protocols/MulticastDataMessage.cs
+++ b/src/Microsoft.Azure.SignalR.Protocols/MulticastDataMessage.cs
@@ -6,12 +6,9 @@ using System.Collections.Generic;
 
 namespace Microsoft.Azure.SignalR.Protocol
 {
-    /// <summary>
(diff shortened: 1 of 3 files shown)


vwxyzh (approved):

src/Microsoft.Azure.SignalR.Protocols/MulticastDataMessage.cs
```diff
@@ -6,12 +6,9 @@
 
 namespace Microsoft.Azure.SignalR.Protocol
 {
-    /// <summary>
-    /// Base class for multicast data messages between Azure SignalR Service and SDK.
-    /// </summary>
-    public abstract class MulticastDataMessage : ExtensibleServiceMessage, IMessageWithTracingId
+    public abstract class MultiPayloadDataMessage : ExtensibleServiceMessage, IMessageWithTracingId
```
JialinXin, line 9: Add summary?

JialinXin (approved):

PR #1691, opened 2022-09-28
Fix `HubConnectionContext.UserIdentifier` is null when negotiation with Management SDK

### LINKED ISSUE #1679: `HubConnectionContext.UserIdentifier` is null if negotiation with Management SDK and set user ID.

Y-Sindo:
When clients negotiatie with Management SDK and connect to SignalR server, IUserIdProvider might not work as the user ID is set directly in the Management SDK.

To make HubConnectionContext.UserIdentifier have the valid value in this case, we should set it before the server accesses it. HubLifetimeManager{THub}.OnConnectedAsync(HubConnectionContext) is the only chance we can set the value. However, we cannot access the Constants.ClaimType.UserId as ASRS system claims are trimmed there. HubConnectionContext.Features is the place where we can store the user Id.

The following code is the injection point.
https://github.com/dotnet/aspnetcore/blob/v6.0.9/src/SignalR/server/Core/src/HubConnectionHandler.cs#L132-L141

Fixes #1679


src/Microsoft.Azure.SignalR/Internals/ServiceUserIdFeature.cs
```diff
@@ -0,0 +1,15 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Azure.SignalR
+{
+    internal class ServiceUserIdFeature : IServiceUserIdFeature
```
vicancy, line 6: no need to have an interface? The class is enough?

Y-Sindo: Make senses. Fixed.

vicancy (approved):

PR #1716, opened 2022-11-10
Fix test cases

terencefan:
Add a `RetryWhenExceptionThrows` method for some fail-prone cases.

### DIFF
--- a/test/Microsoft.Azure.SignalR.Common.Tests/BackOffPolicyFacts.cs
+++ b/test/Microsoft.Azure.SignalR.Common.Tests/BackOffPolicyFacts.cs
@@ -4,51 +4,39 @@
 using System;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Azure.SignalR.Tests.Common;
 using Xunit;
+using Xunit.Abstractions;
 
 namespace Microsoft.Azure.SignalR.Common.Tests
 {
-    public class BackOffPolicyFacts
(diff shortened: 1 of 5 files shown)


src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
```diff
@@ -66,12 +66,10 @@ public override async Task RemoveFromGroupAsync(string connectionId, string grou
                 throw new ArgumentException(NullOrEmptyStringErrorMessage, nameof(connectionId));
             }
 
-            if (string.IsNullOrEmpty(groupName))
-            {
```
Y-Sindo, line 70: Why is this for? We've already have removing user from all groups API. If we want to remove a client connection from all groups, I think it'd be better to have a new API. This class implements interface [HubLifetimeManager<THub>](https://source.dot.net/#Microsoft.AspNetCore.SignalR.Core/HubLifetimeManager.cs) and in original SignalR implementation doesn't allow group name to be null (The `groupName` is not nullable).

terencefan: @xingsy97 failed to update the E2E test so this test will constantly fail before we fixed. I think he is working on fixing the test cases.

I don't know why the previous PR could pass the CI.

vicancy (approved):

PR #1719, opened 2022-11-11
Add `IAccessTokenProvider` to help differentiate token types.

terencefan:


src/Microsoft.Azure.SignalR.AspNet/EndpointProvider/ServiceEndpointProvider.cs
```diff
@@ -110,6 +93,24 @@ public string GetServerEndpoint(string hubName)
             return $"{_serverEndpoint}{ServerPath}/?hub={GetPrefixedHubName(_appName, hubName)}";
         }
 
+        public IAccessTokenProvider GetServerAccessTokenProvider(string hubName, string serverId)
+        {
+            if (_accessKey is AadAccessKey aadAccessKey)
```
vicancy, line 98: this can be determined in .ctor

src/Microsoft.Azure.SignalR.Common/Auth/LocalTokenProvider.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        private readonly TimeSpan _tokenLifetime;
+
+        private readonly IEnumerable<Claim> _claims;
+
+        public LocalTokenProvider(
+            AccessKey accessKey,
+            string audience,
+            IEnumerable<Claim> claims,
+            AccessTokenAlgorithm? algorithm = null,
```
vicancy, line 27: AccessTokenAlgorithm algorithm = AccessTokenAlgorithm.HS256,

src/Microsoft.Azure.SignalR.Common/Interfaces/IAccessTokenProvider.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.SignalR
+{
+    public interface IAccessTokenProvider
```
vicancy, line 8: internal

src/Microsoft.Azure.SignalR.Common/Interfaces/IServiceEndpointProvider.cs
```diff
@@ -15,7 +15,7 @@ internal interface IServiceEndpointProvider
 
         string GetClientEndpoint(string hubName, string originalPath, string queryString);
 
-        Task<string> GenerateServerAccessTokenAsync(string hubName, string userId, TimeSpan? lifetime = null);
+        IAccessTokenProvider GetServerAccessTokenProvider(string hubName, string serverId);
```
vicancy, line 18: why serverId is a parameter, it sounds like a static field per app

src/Microsoft.Azure.SignalR/EndpointProvider/ServiceEndpointProvider.cs
```diff
@@ -52,30 +52,31 @@ public Task<string> GenerateClientAccessTokenAsync(string hubName, IEnumerable<C
...
+        {
+            if (_accessKey is AadAccessKey aadAccessKey)
+            {
+                return new AadTokenProvider(aadAccessKey);
+            }
+            else if (_accessKey is not null)
+            {
+                var audience = _generator.GetServerAudience(hubName, _appName);
+                var claims = serverId != null ? new[] { new Claim(ClaimTypes.NameIdentifier, serverId) } : null;
+                return new LocalTokenProvider(_accessKey, audience, claims, _algorithm, _accessTokenLifetime);
```
vicancy, line 72: serverId can be a field in LocalTokenProvider?

src/Microsoft.Azure.SignalR.AspNet/EndpointProvider/ServiceEndpointProvider.cs
terencefan: Why should we do this? The provider has to be generated every time when this method is being called.

src/Microsoft.Azure.SignalR.Common/Interfaces/IServiceEndpointProvider.cs
terencefan: It's by our current design. 

`ConnectionFactory` generated the `serverId` and make sure this value in the HTTP header and in the identifier claim are equal.

vicancy (approved):

Issue #1707, opened 2022-11-02
Reduce unnecessary exception logs when service reload


vicancy:
### Describe the bug

Misconfigured error level when server connection handshake fails.

### To Reproduce

When doing maintenance work the server might experience such failures when sending messages.
A typical customer logic of sending messages with retry enabled:
![image](https://user-images.githubusercontent.com/668244/201815535-71758d09-418a-4504-87ce-c1a4f491d0eb.png)

### Further technical details
- Your Azure SignalR SDK version: Microsoft.Azure.SignalR/1.19.0-rtm
- Your Server ASPNETCORE version or Assembly version of `Microsoft.AspNetCore.SignalR` 
- Your SignalR Client SDK version

vicancy:
NotActive exception is already mitigated in https://github.com/Azure/azure-signalr/pull/1631

PR #1722, opened 2022-11-15
Change connection drop related log to info since the impact is little

vicancy:
Fix https://github.com/Azure/azure-signalr/issues/1707

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
+++ b/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
@@ -64,6 +64,8 @@ namespace Microsoft.Azure.SignalR
 
         private int _started;
 
+        private readonly string _endpointName;
+
         protected HubServiceEndpoint HubEndpoint { get; }
 
         protected string ServerId { get; }
@@ -120,7 +122,7 @@ namespace Microsoft.Azure.SignalR
 
             _connectionType = connectionType;
             HubEndpoint = endpoint;
-
+            _endpointName = HubEndpoint.ToString();
             if (serviceProtocol != null)
             {
                 _cachedPingBytes = serviceProtocol.GetMessageBytes(PingMessage.Instance);
@@ -181,7 +183,7 @@ namespace Microsoft.Azure.SignalR
                 }
                 catch (Exception ex)
                 {
-                    Log.ConnectionDropped(Logger, HubEndpoint.ToString(), ConnectionId, ex);
+                    Log.ConnectionDropped(Logger, _endpointName, ConnectionId, ex);
                 }
                 finally
                 {
@@ -332,7 +334,7 @@ namespace Microsoft.Azure.SignalR
                 }
                 else if (key.HasExpired)
                 {
-                    Log.AuthorizeFailed(Logger, keyMessage.ErrorMessage, null);
+                    Log.AuthorizeFailed(Logger, _endpointName, keyMessage.ErrorMessage, null);
                     return Task.CompletedTask;
                 }
             }
@@ -354,7 +356,7 @@ namespace Microsoft.Azure.SignalR
                 }
                 catch (Exception ex)
                 {
-                    Log.HandshakeError(Logger, ex.Message, ConnectionId);
+                    Log.HandshakeError(Logger, _endpointName, ex.Message, ConnectionId);
                     await DisposeConnection(connectionContext);
                     return null;
                 }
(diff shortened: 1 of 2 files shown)


vwxyzh (approved):

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -675,24 +666,18 @@ private static class Log
...
                 LoggerMessage.Define(LogLevel.Debug, new EventId(22, "HandshakeComplete"), "Handshake with service completes.");
 
-            private static readonly Action<ILogger, string, Exception> _errorReceivingHandshakeResponse =
-                LoggerMessage.Define<string>(LogLevel.Error, new EventId(23, "ErrorReceivingHandshakeResponse"), "Error receiving handshake response. Id: {ServiceConnectionId}");
-
-            private static readonly Action<ILogger, string, string, Exception> _handshakeError =
-                LoggerMessage.Define<string, string>(LogLevel.Critical, new EventId(24, "HandshakeError"), "Service returned handshake error: {Error}. Id: {ServiceConnectionId}");
+            private static readonly Action<ILogger, string, string, string, Exception> _handshakeError =
+                LoggerMessage.Define<string, string, string>(LogLevel.Information, new EventId(24, "HandshakeError"), "Connection to service '{Endpoint}' returned handshake error, probably caused by network instability or service restart. Will retry after the back off period. Error detail: {Error}. Id: {ServiceConnectionId}");
```
wanlwanl, line 670: we define "Error" in service, should we keep the same?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -637,16 +631,16 @@ private static class Log
         {
             // Category: ServiceConnection
             private static readonly Action<ILogger, ulong?, string, string, Exception> _failedToWrite =
-                LoggerMessage.Define<ulong?, string, string>(LogLevel.Error, new EventId(1, "FailedToWrite"), "Failed to send message {tracingId} to the service: {message}. Id: {ServiceConnectionId}");
+                LoggerMessage.Define<ulong?, string, string>(LogLevel.Error, new EventId(1, "FailedToWrite"), "Failed to send message {TracingId} to the service. Error detail: {Error}. Id: {ServiceConnectionId}");
 
             private static readonly Action<ILogger, string, string, string, Exception> _failedToConnect =
-                LoggerMessage.Define<string, string, string>(LogLevel.Error, new EventId(2, "FailedToConnect"), "Failed to connect to '{endpoint}', will retry after the back off period. Error detail: {message}. Id: {ServiceConnectionId}");
+                LoggerMessage.Define<string, string, string>(LogLevel.Information, new EventId(2, "FailedToConnect"), "Failed to connect to '{Endpoint}', will retry after the back off period. Error detail: {Error}. Id: {ServiceConnectionId}");
```
wanlwanl, line 637: we define "Error" in service, should we keep the same?

wanlwanl, line 637: we define "Warning" (maybe caused by service reload, timeout, ..., so we use "Warning") in service, should we keep the same?

vicancy: I feel there is nothing to warn the customer since it does not have real impact

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
vicancy: Error means the customers receive alerts and they need to do action to solve it. Since customers can do little when the connection drops and actually it has little impact on the customer, Info sounds reasonable.

vicancy: We have warning logs when all the connections to the endpoints are dropped. -- and it is when there is real impact to the customer.

wanlwanl: Error/warning is a way to notify customer there is/likely to be a problem no matter customer can do action? and Info should be expected behavior like xxx success/start/end?

vicancy: Yeah so the issue here is there is no real problem here so we don't want to notify customers. Maybe we should change it to "Verbose"?

vicancy: `Debug` sounds make more sense. Changed

vicancy: Offline discussed. We will keep connection related logs information, customer to aware logs warning, and application exceptions error.

PR #1726, opened 2022-11-17
some code cleanup for rest api

terencefan:
 - Detail 1
 - Detail 2

### DIFF
--- a/src/Microsoft.Azure.SignalR.Management/RestApiProvider.cs
+++ b/src/Microsoft.Azure.SignalR.Management/RestApiProvider.cs
@@ -16,25 +16,22 @@ namespace Microsoft.Azure.SignalR.Management
 
         private readonly RestApiAccessTokenGenerator _restApiAccessTokenGenerator;
 
-        private readonly string _audienceBaseUrl;
-        private readonly string _serverEndpoint;
+        private readonly ServiceEndpoint _serviceEndpoint;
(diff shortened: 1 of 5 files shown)


src/Microsoft.Azure.SignalR.Management/RestApiProvider.cs
```diff
@@ -16,25 +16,22 @@ internal class RestApiProvider
 
         private readonly RestApiAccessTokenGenerator _restApiAccessTokenGenerator;
 
-        private readonly string _audienceBaseUrl;
-        private readonly string _serverEndpoint;
+        private readonly ServiceEndpoint _serviceEndpoint;
+
+        private string AudienceBaseUrl => _serviceEndpoint.AudienceBaseUrl;
```
Y-Sindo, line 21: Why do we use private property?

terencefan: To avoiding allocating 2 string references (and an enum value in the next PR)

terencefan: I think it's better than using multiple private fields that belong to the same object?

There will be an additional `AuthType` field in the next PR on `serviceEndpoint`.

Y-Sindo (approved):

PR #1744, opened 2023-01-05
Support MessagePack for persistent mode

Y-Sindo:
Add method `ServiceManagerBuilder.WithHubProtocols`, `ServiceManagerBuilder.AddHubProtocol`

### DIFF
--- a/src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/ServiceHubLifetimeManagerFactory.cs
+++ b/src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/ServiceHubLifetimeManagerFactory.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using System.Net.Http;
 using Azure.Core.Serialization;
 using Microsoft.AspNetCore.SignalR;
+using Microsoft.AspNetCore.SignalR.Protocol;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Options;
 
@@ -36,6 +37,7 @@ namespace Microsoft.Azure.SignalR.Management
                     }
                 case ServiceTransportType.Transient:
                     {
+                        CheckHubProtocols();
 #pragma warning disable CS0618 // Type or member is obsolete
                         var payloadSerializerSettings = _options.JsonSerializerSettings;
 #pragma warning restore CS0618 // Type or member is obsolete
@@ -52,6 +54,15 @@ namespace Microsoft.Azure.SignalR.Management
                         return new RestHubLifetimeManager<THub>(hubName, serviceEndpoint, _options.ProductInfo, _options.ApplicationName, restClient);
                     }
                 default: throw new InvalidEnumArgumentException(nameof(ServiceManagerOptions.ServiceTransportType), (int)_options.ServiceTransportType, typeof(ServiceTransportType));
+            }
+        }
+
+        private void CheckHubProtocols()
+        {
+            var protocols = _serviceProvider.GetServices<IHubProtocol>().ToArray();
+            if (protocols.Length > 1 || protocols.Where(p => p.Name.Equals(Constants.Protocol.MessagePack)).Any())
+            {
+                throw new InvalidOperationException("ServiceManagerBuilder.WithHubProtocols method is not supported for transient(default) mode yet.");
             }
         }
     }
--- a/src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/ServiceHubLifetimeManagerFactory.cs
+++ b/src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/ServiceHubLifetimeManagerFactory.cs
@@ -60,7 +60,7 @@ namespace Microsoft.Azure.SignalR.Management
         private void CheckHubProtocols()
         {
             var protocols = _serviceProvider.GetServices<IHubProtocol>().ToArray();
-            if (protocols.Length > 1 || protocols.Where(p => p.Name.Equals(Constants.Protocol.MessagePack)).Any())
+            if (protocols.Length > 1 || protocols.Any(p => p.Name.Equals(Constants.Protocol.MessagePack, StringComparison.OrdinalIgnoreCase)))
             {
                 throw new InvalidOperationException("ServiceManagerBuilder.WithHubProtocols method is not supported for transient(default) mode yet.");
             }
--- a/src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/ServiceHubLifetimeManagerFactory.cs
+++ b/src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/ServiceHubLifetimeManagerFactory.cs
@@ -60,7 +60,7 @@ namespace Microsoft.Azure.SignalR.Management
         private void CheckHubProtocols()
         {
             var protocols = _serviceProvider.GetServices<IHubProtocol>().ToArray();
-            if (protocols.Length > 1 || protocols.Any(p => p.Name.Equals(Constants.Protocol.MessagePack, StringComparison.OrdinalIgnoreCase)))
+            if (protocols.Any(p => p.Name.Equals(Constants.Protocol.MessagePack, StringComparison.OrdinalIgnoreCase)))
             {
                 throw new InvalidOperationException("ServiceManagerBuilder.WithHubProtocols method is not supported for transient(default) mode yet.");
             }
--- a/src/Microsoft.Azure.SignalR.Management/Serialization/JsonObjectSerializerHubProtocol.cs
+++ b/src/Microsoft.Azure.SignalR.Management/Serialization/JsonObjectSerializerHubProtocol.cs
@@ -20,14 +20,14 @@ using Microsoft.AspNetCore.SignalR.Protocol;
 namespace Microsoft.Azure.SignalR.Management
 {
     /// <summary>
-    /// Implements the SignalR Hub Protocol using <see cref="ObjectSerializer"/>.
+    /// Implements the SignalR Hub Protocol using <see cref="global::Azure.Core.Serialization.ObjectSerializer"/>.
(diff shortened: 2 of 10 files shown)


src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
```diff
@@ -79,14 +80,40 @@ public ServiceManagerBuilder WithNewtonsoftJson(Action<NewtonsoftServiceHubProto
...
+        /// <param name="hubProtocols">Only the protocols named "json" or "messagepack" are allowed.</param>
+        /// <returns>The <see cref="ServiceHubContextBuilder"/> instance itself.</returns>
+        public ServiceManagerBuilder WithHubProtocol(IEnumerable<IHubProtocol> hubProtocols)
+        {
+            if (hubProtocols == null)
+            {
+                throw new ArgumentNullException(nameof(hubProtocols));
+            }
+            // Allows the user to use MessagePack only.
+            _services.RemoveAll<IHubProtocol>();
```
vicancy, line 105: shall we move this check to ServiceManagerBuilder.Build()

src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/ServiceHubLifetimeManagerFactory.cs
```diff
@@ -54,5 +56,14 @@ public ServiceHubLifetimeManagerFactory(IServiceProvider sp, IOptions<ServiceMan
                 default: throw new InvalidEnumArgumentException(nameof(ServiceManagerOptions.ServiceTransportType), (int)_options.ServiceTransportType, typeof(ServiceTransportType));
             }
         }
+
+        private void CheckHubProtocols()
+        {
+            var protocols = _serviceProvider.GetServices<IHubProtocol>().ToArray();
+            if (protocols.Length > 1 || protocols.Where(p => p.Name.Equals(Constants.Protocol.MessagePack)).Any())
+            {
+                throw new InvalidOperationException("ServiceManagerBuilder.WithHubProtocols method is not supported for transient(default) mode yet.");
```
vicancy, line 65:
```suggestion
                throw new InvalidOperationException("ServiceManagerBuilder.WithHubProtocol method is not supported for transient(default) mode yet.");
```

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
```diff
@@ -79,14 +80,40 @@ public ServiceManagerBuilder WithNewtonsoftJson(Action<NewtonsoftServiceHubProto
...
         }
 
+        /// <summary>
+        /// Sets the SignalR hub protocols to serialize messages sent to SignalR clients.
+        /// </summary>
+        /// <remarks><para>Currently it only works for <b>persistent</b> mode. The support for <b>transiet(default)</b> mode is to be done. </para> 
+        /// <para>Calling this method clears the default hub protocol.</para></remarks>
+        /// <param name="hubProtocols">Only the protocols named "json" or "messagepack" are allowed.</param>
+        /// <returns>The <see cref="ServiceHubContextBuilder"/> instance itself.</returns>
+        public ServiceManagerBuilder WithHubProtocol(IEnumerable<IHubProtocol> hubProtocols)
```
vicancy, line 98: `WithHubProtocol(IHubProtocol hubProtocol)` and support multiple such calls?

Y-Sindo: The problem of `WithHubProtocol(IHubProtocol hubProtocol)` is that it only adds protocols, but cannot removes. If the user only wants MessagePack, they cannot achieve it.

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
Y-Sindo: Why do we want to move the check to `ServiceManagerBuilder.Build()`?

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
vicancy: Then use `WithHubProtocols`

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
vicancy: Build() is the time when everything is set. What if IHubProtocol is added after `WithHubProtocol`?

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
vicancy: how about `WithHubProtocols(params IHubProtocol[] ...)`

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
```diff
@@ -79,14 +80,40 @@ public ServiceManagerBuilder WithNewtonsoftJson(Action<NewtonsoftServiceHubProto
...
+        {
+            if (hubProtocols == null)
+            {
+                throw new ArgumentNullException(nameof(hubProtocols));
+            }
+            // Allows the user to use MessagePack only.
+            _services.RemoveAll<IHubProtocol>();
+            foreach (var hubProtocol in hubProtocols)
+            {
+                if (!hubProtocol.Name.Equals(Constants.Protocol.Json, StringComparison.OrdinalIgnoreCase) && !hubProtocol.Name.Equals(Constants.Protocol.MessagePack, StringComparison.OrdinalIgnoreCase))
```
vicancy, line 108: is the protocol name case-insensitive?

src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/ServiceHubLifetimeManagerFactory.cs
```diff
@@ -54,5 +56,14 @@ public ServiceHubLifetimeManagerFactory(IServiceProvider sp, IOptions<ServiceMan
                 default: throw new InvalidEnumArgumentException(nameof(ServiceManagerOptions.ServiceTransportType), (int)_options.ServiceTransportType, typeof(ServiceTransportType));
             }
         }
+
+        private void CheckHubProtocols()
+        {
+            var protocols = _serviceProvider.GetServices<IHubProtocol>().ToArray();
+            if (protocols.Length > 1 || protocols.Where(p => p.Name.Equals(Constants.Protocol.MessagePack)).Any())
```
vicancy, line 63:
```suggestion
            if (protocols.Length > 1 || protocols.Any(p => p.Name.Equals(Constants.Protocol.MessagePack)))
```
case sensitive or insensitive here?

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
Y-Sindo: updated

src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs
Y-Sindo: Yes, consistent with SignalR implementation https://source.dot.net/#Microsoft.AspNetCore.SignalR.Core/Internal/DefaultHubProtocolResolver.cs,22

galvesribeiro (approved):
LGTM

src/Microsoft.Azure.SignalR.Management/Serialization/JsonObjectSerializerHubProtocol.cs
```diff
@@ -57,15 +57,15 @@ internal sealed class JsonObjectSerializerHubProtocol : IHubProtocol
         private const string ProtocolName = "json";
         private const int ProtocolVersion = 1;
 
-        private readonly ObjectSerializer _objectSerializer;
```
Y-Sindo, line 60: It has to be public later when implementing full support for MessagePack in transient mode. Make it public first for tests.

vicancy (approved):
Looks great to me!

Y-Sindo:
@galvesribeiro The package is released.

galvesribeiro:
@Y-Sindo thanks. But is it fixed for the Transient mode as well? Or it is just for persistent mode?

The case I've opened with the support is relative to Transient mode explicitly.

Thanks

Y-Sindo:
> @Y-Sindo thanks. But is it fixed for the Transient mode as well? Or it is just for persistent mode?
> 
> The case I've opened with the support is relative to Transient mode explicitly.
> 
> Thanks

@galvesribeiro Not yet. The transient mode currently just sends JSON payload in REST API and lets the service translate JSON into MessagePack, which is why it doesn't work as you expected. To improve the behavior, we need to send MessagePack payload directly in the REST API, and the change will be bigger than the persistent mode and not that easy. To track this issue, please see #1745 .

galvesribeiro:
Ok, then my ticket on the support is not fixed then...

Y-Sindo:
@galvesribeiro Is it OK for you to switch to persistent mode as a workaround?

Y-Sindo:
@galvesribeiro The [doc](https://github.com/Azure/azure-signalr/blob/dev/docs/management-sdk-guide.md) is also updated to remind users of the limitation of the MessagePack support for transient mode. We are also working actively to improve that.

galvesribeiro:
Is it ok to have a single management connection as a singleton for the whole service assuming it has a very hight message volume to multiple target clients? In other words, is it thread-safe?

If that is ok, does it reconnect in case the connection is lost with SRS?

Is it queueing messages to be dispatched from multiple callers?

Y-Sindo:
> Is it ok to have a single management connection as a singleton for the whole service assuming it has a very hight message volume to multiple target clients? In other words, is it thread-safe?

It is thread-safe. Each thread tries to wait for a lock to write message to the connection. 

> If that is ok, does it reconnect in case the connection is lost with SRS?

The connection will try to reconnect if it is lost. 

> Is it queueing messages to be dispatched from multiple callers?

No. As in the first answer, each caller will try to get a lock, and when the method returns, the message is already received by the service. During the interval when the connection is disconnected, you have to queue the message yourself.

PR #1748, opened 2023-01-09
Support MessagePack in transient mode in SDK-side

### LINKED ISSUE #1745: Full support for MessagePack in transient mode

Y-Sindo:
Close #1745 
1. Make the `Protocol` project internal visible to `Common`, so that `MessagePackWriter` class can be used by `Common` project. As the `SequenceOfT` was under the `Common` project source directory and referred as source file by `Protocol` before, which was conflicted after internal visible, I move the file.
2. Avoid breaking change. The default hub protocol for transient mode is `JsonObjectSerializerHubProtocol`. If user doesn't add their own `IHubProtocol` ( the detecting logic is in the `PayloadBuilderResolver`) , then JSON payload is used, and the service side still does JSON->MessagePack conversion. Otherwise binary payload is used.


Y-Sindo:
@galvesribeiro This PR supports MessagePack in transient mode in SDK-side, please take a look.

src/Microsoft.Azure.SignalR.Management/Serialization/PayloadBuilderResolver.cs
```diff
@@ -0,0 +1,35 @@
...
+        private readonly IHubProtocolResolver _hubProtocolResolver;
+
+        public PayloadBuilderResolver(IHubProtocolResolver hubProtocolResolver)
+        {
+            _hubProtocolResolver = hubProtocolResolver;
+        }
+
+        public IPayloadContentBuilder GetPayloadContentBuilder()
+        {
+            if (_hubProtocolResolver.AllProtocols.Count == 1 && _hubProtocolResolver.AllProtocols.Single() is JsonObjectSerializerHubProtocol jsonObjectSerializerHubProtocol)
```
vicancy, line 23:
```suggestion
            if (_hubProtocolResolver.AllProtocols.Count == 1 && _hubProtocolResolver.AllProtocols[0] is JsonObjectSerializerHubProtocol jsonObjectSerializerHubProtocol)
```

src/Microsoft.Azure.SignalR.Common/Utilities/Rest/BinaryPayloadMessageContent.cs
```diff
@@ -0,0 +1,70 @@
...
+        };
+        private static readonly MediaTypeHeaderValue ContentType = new("application/octet-stream");
+
+        private readonly PayloadMessage _payloadMessage;
+        private readonly IReadOnlyList<IHubProtocol> _hubProtocols;
+
+        public BinaryPayloadMessageContent(PayloadMessage payloadMessage, IReadOnlyList<IHubProtocol> hubProtocols)
+        {
+            _payloadMessage = payloadMessage ?? throw new ArgumentNullException(nameof(payloadMessage));
+            _hubProtocols = hubProtocols ?? throw new ArgumentNullException(nameof(hubProtocols));
```
vicancy, line 33: if hubprotocols.Length = 0, no need to send out the request?

src/Microsoft.Azure.SignalR.Management/DependencyInjectionExtensions.cs
```diff
@@ -88,8 +89,19 @@ private static IServiceCollection AddSignalRServiceCore(this IServiceCollection
...
-                return objectSerializer != null ? new JsonObjectSerializerHubProtocol(objectSerializer) : new JsonHubProtocol();
+                var serviceManagerOptions = sp.GetRequiredService<IOptions<ServiceManagerOptions>>().Value;
+                var objectSerializer = serviceManagerOptions.ObjectSerializer;
+                if (objectSerializer != null)
+                {
+                    return new JsonObjectSerializerHubProtocol(objectSerializer);
+                }
+#pragma warning disable CS0618 // Type or member is obsolete
+                // The default protocol is different for historical reason.
+                return serviceManagerOptions.ServiceTransportType == default ?
```
vicancy, line 100: Use the explicit value for readability 
```suggestion
                return serviceManagerOptions.ServiceTransportType == ServiceTransportType.Transient ?
```

src/Microsoft.Azure.SignalR.Protocols/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,4 @@
+
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("Microsoft.Azure.SignalR.Common, PublicKey=0024000004800000940000000602000000240000525341310004000001000100f33a29044fa9d740c9b3213a93e57c84b472c84e0b8a0e1ae48e67a9f8f6de9d5f7f3d52ac23e48ac51801f1dc950abe901da34d2a9e3baadb141a17c77ef3c565dd5ee5054b91cf63bb3c6ab83f72ab3aafe93d0fc3c2348b764fafb0b1c0733de51459aeab46580384bf9d74c4e28164b7cde247f891ba07891c9d872ad2bb")]
```
vicancy, line 4: for which method?

Y-Sindo: `MessagePackWriter`

src/Microsoft.Azure.SignalR.Common/Utilities/Rest/BinaryPayloadMessageContent.cs
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        private void WriteMessageCore(IBufferWriter<byte> bufferWriter)
+        {
+            var invocationMessage = new InvocationMessage(_payloadMessage.Target, _payloadMessage.Arguments);
+            var messagePackWriter = new MessagePackWriter(bufferWriter);
+            messagePackWriter.WriteMapHeader(_hubProtocols.Count);
+            foreach (var hubProtocol in _hubProtocols)
+            {
+                messagePackWriter.WriteString(ProtocolMap[hubProtocol.Name]);
```
vwxyzh, line 64: what about new protocols added by user?

Y-Sindo: Currently we don't allow users to add other custom protocols, to keep consistent with function upstream which only supports JSON and MessagePack.

src/Microsoft.Azure.SignalR.Common/Utilities/Rest/BinaryPayloadMessageContent.cs
```diff
@@ -0,0 +1,70 @@
...
+            Headers.ContentType = ContentType;
+        }
+
+        protected override async Task SerializeToStreamAsync(Stream stream, TransportContext context)
+        {
+            var memoryBufferWriter = MemoryBufferWriter.Get();
+            try
+            {
+                WriteMessageCore(memoryBufferWriter);
+                await memoryBufferWriter.CopyToAsync(stream);
```
vicancy, line 43: MemoryBufferWriter._cacheInstance is thread local, so using await inside makes `Return` to another thread

src/Microsoft.Azure.SignalR.Common/Utilities/Rest/BinaryPayloadMessageContent.cs
Y-Sindo: Do we want to allow zero hub protocol? I don't think it's normal. How about throwing exception if we detect no hub protocol?

src/Microsoft.Azure.SignalR.Common/Utilities/Rest/BinaryPayloadMessageContent.cs
Y-Sindo: As memory buffer writer actually doesn't hold any byte array after `Reset()`, I think it's safe to directly new a buffer writer. Will update.

galvesribeiro (approved):
LGTM

src/Microsoft.Azure.SignalR.Common/Utilities/Rest/BinaryPayloadMessageContent.cs
Y-Sindo: This class can't control whether to send out the request. I find that SignalR doesn't detect whether length=0, therefore keep it unchanged.

src/Microsoft.Azure.SignalR.Common/Utilities/Rest/BinaryPayloadMessageContent.cs
Y-Sindo: updated

vwxyzh (approved):

Y-Sindo:
@galvesribeiro  We have a [preview version](https://www.myget.org/feed/azure-signalr-dev/package/nuget/Microsoft.Azure.SignalR.Management) now. Could you have a test and tell us if it solves your problem? Once you confirm that, we will release a new version.

galvesribeiro:
@Y-Sindo I can confirm that it works. Thank you.

Y-Sindo:
@galvesribeiro  The new version is released. Please upgrade to 1.21.0

PR #1779, opened 2023-04-14
Use sticky service connections for ackable messages

KKhurin:
#### This ensures correct order of ackable and non ackable messages.
For example, the following scenarios may occasionally fail (due to race condition) without this fix:
 - Removing after send should not prevent sending message to group:
~~~
{
 await Groups.AddToGroupAsync(connId, group);
 await Clients.Group(group).SendAsync(method, msg); // msg may never get delivered to connId
 await Groups.RemoveFromGroupAsync(connId, group);
}
~~~
 
 - Messages sent to a group should not get delivered to connections added to group afterwards:
~~~
{
 await Groups.RemoveFromGroupAsync(connId, group);
 await Clients.Group(group).SendAsync(method, msg); // msg may get sent to connId even though it is only added after SendAsync.
 await Groups.AddToGroupAsync(connId, group);
}
~~~
#### Note:
The race conditions affecting the scenarios above are rare and they should work _most of the time_. However, if your App logic depends on such scenarios, then please update to the future SDK version that will contain this fix.

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
+++ b/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
@@ -244,8 +244,7 @@ namespace Microsoft.Azure.SignalR
             var task = _ackHandler.CreateAck(out var id, cancellationToken);
             ackableMessage.AckId = id;
 
-            // There is no need to write ackable message to sticky connections
-            await WriteWithRetry(serviceMessage, null, ServiceConnections);
+            await WriteToScopedOrRandomAvailableConnection(serviceMessage);
 
             var status = await task;
             switch (status)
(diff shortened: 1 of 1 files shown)


src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -244,8 +244,7 @@ public async Task<bool> WriteAckableMessageAsync(ServiceMessage serviceMessage,
             var task = _ackHandler.CreateAck(out var id, cancellationToken);
             ackableMessage.AckId = id;
 
-            // There is no need to write ackable message to sticky connections
-            await WriteWithRetry(serviceMessage, null, ServiceConnections);
+            await WriteToScopedOrRandomAvailableConnection(serviceMessage);
```
vicancy, line 247: also add some comments here?

vicancy (approved):

vicancy (approved):

PR #1801, opened 2023-06-26
Fix message order issue for management SDK

Y-Sindo:


src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -515,22 +516,21 @@ private async Task WriteToScopedOrRandomAvailableConnection(ServiceMessage servi
             }
             else
             {
-                await WriteWithRetry(serviceMessage, null, currentConnections);
+                var connectionUsed = await WriteWithRetry(serviceMessage, _inUseConnection, currentConnections, IterateConnectionsInFixedOrder);
+                // Similarly, here is currently no synchronization when persisting selected connection in _inUseConnection.
+
+                if (connectionUsed != _inUseConnection)
```
vicancy, line 522: check+set is not atomic, maybe simply set

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -515,22 +516,21 @@ private async Task WriteToScopedOrRandomAvailableConnection(ServiceMessage servi
             }
             else
             {
-                await WriteWithRetry(serviceMessage, null, currentConnections);
+                var connectionUsed = await WriteWithRetry(serviceMessage, _inUseConnection, currentConnections, IterateConnectionsInFixedOrder);
```
vicancy, line 519: so it changes the current behavior that the traffic distributes to random connections for proxy mode. is using random connections enough?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -542,23 +542,33 @@ private async Task<IServiceConnection> WriteWithRetry(ServiceMessage serviceMess
...
                     catch (ServiceConnectionNotActiveException)
                     {
-                        if (retry == maxRetry - 1)
-                        {
-                            throw;
-                        }
                     }
                 }
+                iterator ??= iterateConnections(currentConnections).GetEnumerator();
+                connection = iterator.MoveNext() ? iterator.Current : throw new ServiceConnectionNotActiveException();
```
vicancy, line 548: using `do.. while(iterator.MoveNext())` pattern?

PR #1828, opened 2023-09-07
Partial support client results in sharding mode.

JialinXin:
Make client invocation supported in sharding mode. Currently CI manager rely on single response from service side which make it not work in multiple endpoints scenarios that server side has to manage all the endpoints results before completion. 

- [X] Support CI ServiceMessages in routing. Limitation: server should have knowledge of client connection and endpoint mapping so custom router has to return __single__ endpoint always and won't get noise by irrelevant endpoints. Otherwise, throw exception before fully support.
- [ ] Make CI manager supports multiple endpoints results management.

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
+++ b/src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
@@ -427,7 +427,7 @@ namespace Microsoft.Azure.SignalR
 
         private IEnumerable<ServiceEndpoint> SingleOrNotSupported(IEnumerable<ServiceEndpoint> endpoints, ServiceMessage message)
         {
-            if (endpoints.Count() == 1)
+            if (endpoints.ToList().Count == 1)
             {
                 return endpoints;
             }
--- a/src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
+++ b/src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
@@ -230,6 +230,15 @@ namespace Microsoft.Azure.SignalR
                 case CloseConnectionMessage closeConnectionMessage:
                     return _router.GetEndpointsForConnection(closeConnectionMessage.ConnectionId, endpoints);
 
+                case ClientInvocationMessage clientInvocationMessage:
+                    return SingleOrNotSupported(_router.GetEndpointsForConnection(clientInvocationMessage.ConnectionId, endpoints), clientInvocationMessage);
+
+                case ServiceMappingMessage serviceMappingMessage:
+                    return SingleOrNotSupported(_router.GetEndpointsForConnection(serviceMappingMessage.ConnectionId, endpoints), serviceMappingMessage);
+
+                case ServiceCompletionMessage serviceCompletionMessage:
+                    return SingleOrNotSupported(_router.GetEndpointsForConnection(serviceCompletionMessage.ConnectionId, endpoints), serviceCompletionMessage);
+
                 default:
                     throw new NotSupportedException(message.GetType().Name);
             }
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
```diff
@@ -416,6 +425,15 @@ private async Task WaitForClientsDisconnect(HubServiceEndpoint endpoint)
             Log.TimeoutWaitingClientsDisconnect(_logger, endpoint.ToString(), (int)_scaleTimeout.TotalSeconds);
         }
 
+        private IEnumerable<ServiceEndpoint> SingleOrNotSupported(IEnumerable<ServiceEndpoint> endpoints, ServiceMessage message)
+        {
+            if (endpoints.Count() <= 1)
```
vicancy, line 430: why <1 works?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
```diff
@@ -416,6 +425,15 @@ private async Task WaitForClientsDisconnect(HubServiceEndpoint endpoint)
...
+            {
+                return endpoints;
+            }
+            throw new NotSupportedException(message.GetType().Name);
```
vicancy, line 434: give it a clearer message that multiple endpoint does not yet support such type

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
```diff
@@ -416,6 +425,15 @@ private async Task WaitForClientsDisconnect(HubServiceEndpoint endpoint)
             Log.TimeoutWaitingClientsDisconnect(_logger, endpoint.ToString(), (int)_scaleTimeout.TotalSeconds);
         }
 
+        private IEnumerable<ServiceEndpoint> SingleOrNotSupported(IEnumerable<ServiceEndpoint> endpoints, ServiceMessage message)
+        {
+            if (endpoints.Count() == 1)
```
vwxyzh, line 430: what about 0?

JialinXin: 0 simply logs and in client invocation scenario our CI manager will wait for the result until timeout. So to avoid wasting time, suppose custom router should not return 0 endpoints. https://github.com/Azure/azure-signalr/blob/dev/src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointMessageWriter.cs#L136-L140

Any suggestion about check count under IEnumerable<T>? ToList() first?

JialinXin: Separate 0/many in exceptions for better information.

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
```diff
@@ -416,6 +425,20 @@ private async Task WaitForClientsDisconnect(HubServiceEndpoint endpoint)
...
+        private IEnumerable<ServiceEndpoint> SingleOrNotSupported(IEnumerable<ServiceEndpoint> endpoints, ServiceMessage message)
+        {
+            var endpointCnt = endpoints.ToList().Count;
+            if (endpointCnt == 1)
+            {
+                return endpoints;
+            }
+            if (endpointCnt == 0)
+            {
+                throw new ArgumentException("Client invocation is not sent because no endpoint is returned from the endpoint router.");
```
vicancy, line 437: throw AzureSignalRNoEndpointAvailableException?

vicancy (approved):

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointServiceConnectionContainer.cs
JialinXin: The exception sounds like endpoints are down but actually results are returned from (custom) router. So suggest current information is clear enough.

PR #1834, opened 2023-09-14
Update dev version to 1.21.8

JialinXin:


vicancy (approved):
don't forget to remove not needed descrip

PR #1842, opened 2023-10-12
Support different EndpointRoutingMode

bjqian:
### Summary of the changes 
 - Add different endpoint routing mode: `Weighted`, `LeastConnection`, `Random`
 - For weighted mode: change the weight definition from _remaining connection quota_ to ( _remaining connection quota / connection capacity)_

### DIFF
--- a/src/Microsoft.Azure.SignalR/EndpointRouters/DefaultEndpointRouter.cs
+++ b/src/Microsoft.Azure.SignalR/EndpointRouters/DefaultEndpointRouter.cs
@@ -6,29 +6,41 @@ using System.Collections.Generic;
 using System.Linq;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Azure.SignalR.Common;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.Azure.SignalR
 {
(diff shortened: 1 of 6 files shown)


src/Microsoft.Azure.SignalR/EndpointRouters/DefaultEndpointRouter.cs
```diff
@@ -69,7 +80,7 @@ private ServiceEndpoint GetEndpointAccordingToWeight(ServiceEndpoint[] available
                 var remain = endpointMetrics.ConnectionCapacity -
                              (endpointMetrics.ClientConnectionCount +
                               endpointMetrics.ServerConnectionCount);
-                var weight = remain > 0 ? remain : 1;
+                var weight = Math.Max((int)((double)remain / endpointMetrics.ConnectionCapacity * 1000), 1);
```
vwxyzh, line 83: minor: we can keep it in integer operations:
`remain * 1000 / endpointMetrics.ConnectionCapacity`
unit 1000 * 1000 connections/unit * 1000 = 1G, it is less than max value of integer (2G).

vwxyzh (approved):

PR #1844, opened 2023-10-16
Add retry capability to REST client to handle transient errors

Y-Sindo:
Add retry capability to REST client to handle transient errors. Tracking issue: https://github.com/Azure/azure-functions-signalrservice-extension/issues/293

## Add two retry modes:
1. Fixed delay
3. Exponential backoff delay （Do we need to add jitter to the delay?)

## For message APIs:
Only the situations when HTTP status code > 500  are safe to retry.
For 500 error, runtime returns 500 when timeout, so it's not safe to retry.

## For non-message APIs:
The following situations are considered to be transient:
1. HTTP response status code 5xx or 408(Request timeout).
2. Request timeout with the timeout length configured as `ServiceManagerOptions.HttpClientTimeout`

### DIFF
--- a/src/Microsoft.Azure.SignalR.Management/Configuration/RetryOptions.cs
+++ b/src/Microsoft.Azure.SignalR.Management/Configuration/RetryOptions.cs
@@ -1,0 +1,34 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Microsoft.Azure.SignalR.Management;
+
+#nullable enable
+
+public class RetryOptions
+{
+    /// <summary>
+    /// The maximum number of retry attempts before giving up.
+    /// </summary>
+    public int MaxRetries { get; set; } = 3;
+
+    /// <summary>
+    /// The delay between retry attempts for a fixed approach or the delay
+    /// on which to base calculations for a backoff-based approach.
+    /// </summary>
+    public TimeSpan Delay { get; set; } = TimeSpan.FromSeconds(0.8);
+
+    /// <summary>
+    /// The maximum permissible delay between retry attempts.
+    /// </summary>
+    public TimeSpan MaxDelay { get; set; } = TimeSpan.FromMinutes(1);
+
+    /// <summary>
+    /// The approach to use for calculating retry delays.
+    /// </summary>
+    public RetryMode Mode { get; set; } = RetryMode.Fixed;
+}
+
+
--- a/src/Microsoft.Azure.SignalR.Management/DependencyInjectionExtensions.cs
+++ b/src/Microsoft.Azure.SignalR.Management/DependencyInjectionExtensions.cs
@@ -156,15 +156,52 @@ namespace Microsoft.Azure.SignalR.Management
             return services.Configure<ServiceManagerOptions>(o => o.ProductInfo ??= productInfo);
         }
 
-        private static IServiceCollection AddRestClientFactory(this IServiceCollection services) => services
-            .AddHttpClient(Options.DefaultName, (sp, client) => client.Timeout = sp.GetRequiredService<IOptions<ServiceManagerOptions>>().Value.HttpClientTimeout)
-            .ConfigurePrimaryHttpMessageHandler(sp => new HttpClientHandler() { Proxy = sp.GetRequiredService<IOptions<ServiceManagerOptions>>().Value.Proxy }).Services
-            .AddSingleton(sp =>
+        private static IServiceCollection AddRestClientFactory(this IServiceCollection services)
+        {
+            // For AAD, health check.
+            services
+                .AddHttpClient(Options.DefaultName, (sp, client) => client.Timeout = sp.GetRequiredService<IOptions<ServiceManagerOptions>>().Value.HttpClientTimeout)
+                .ConfigurePrimaryHttpMessageHandler(ConfigureProxy);
+
+            // For other data plane APIs.
+            services.AddSingleton<IBackOffPolicy>(sp =>
+            {
+                var options = sp.GetRequiredService<IOptions<ServiceManagerOptions>>().Value;
+                var retryOptions = options.RetryOptions;
+                return retryOptions == null
+                    ? new DummyBackOffPolicy()
+                    : retryOptions.Mode switch
+                    {
+                        RetryMode.Fixed => ActivatorUtilities.CreateInstance<FixedBackOffPolicy>(sp),
+                        RetryMode.Exponential => ActivatorUtilities.CreateInstance<ExponentialBackOffPolicy>(sp),
+                        _ => throw new NotSupportedException($"Retry mode {retryOptions.Mode} is not supported.")
+                    };
+            });
+            services
+                .AddHttpClient(Constants.HttpClientNames.Resilient, (sp, client) =>
+                {
+                    var options = sp.GetRequiredService<IOptions<ServiceManagerOptions>>().Value;
+                    if (options.RetryOptions == null)
+                    {
+                        client.Timeout = options.HttpClientTimeout;
+                    }
+                    // Else the timeout is enforced by TimeoutHttpMessageHandler.
+                })
+                .ConfigurePrimaryHttpMessageHandler(ConfigureProxy)
+                .AddHttpMessageHandler(sp => ActivatorUtilities.CreateInstance<RetryHttpMessageHandler>(sp))
+                .AddHttpMessageHandler(sp => ActivatorUtilities.CreateInstance<TimeoutHttpMessageHandler>(sp));
+
+            services.AddSingleton(sp =>
             {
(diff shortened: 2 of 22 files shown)


src/Microsoft.Azure.SignalR.Management/Resilient/RetryHttpMessageHandler.cs
```diff
@@ -0,0 +1,79 @@
...
+                exceptions.Add(ex);
+            }
+            await Task.Delay(delays.Current, cancellationToken);
+        } while (true);
+    }
+
+    private static bool IsTransientError(HttpStatusCode code)
+    {
+        return code >= HttpStatusCode.InternalServerError ||
+               code == HttpStatusCode.RequestTimeout;
```
vwxyzh, line 77: clients will receive two messages when the first request timeout.

src/Microsoft.Azure.SignalR.Management/Configuration/RetryOptions.cs
```diff
@@ -0,0 +1,34 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Microsoft.Azure.SignalR.Management;
+
+#nullable enable
+
+public class RetryOptions
```
vicancy, line 10: Give the class and `RetryMode` a more specific naming, e.g. `HttpClientRetryOptions`?

src/Microsoft.Azure.SignalR.Management/DependencyInjectionExtensions.cs
```diff
@@ -156,15 +157,65 @@ private static IServiceCollection TrySetProductInfo(this IServiceCollection serv
...
+            });
+            services
+                .AddHttpClient(Constants.HttpClientNames.Resilient, (sp, client) =>
+                {
+                    var options = sp.GetRequiredService<IOptions<ServiceManagerOptions>>().Value;
+                    if (options.RetryOptions == null)
+                    {
+                        client.Timeout = options.HttpClientTimeout;
+                    }
+                    // Else the timeout is enforced by TimeoutHttpMessageHandler.
```
vicancy, line 189: question: why not always use TimeoutHttpMessageHandler?

src/Microsoft.Azure.SignalR.Management/Resilient/ExponentialBackoffPolicy.cs
```diff
@@ -0,0 +1,44 @@
...
+    private readonly int _maxRetries;
+    private readonly TimeSpan _minDelay;
+    private readonly TimeSpan _maxDelay;
+
+    public ExponentialBackOffPolicy(IOptions<ServiceManagerOptions> options)
+    {
+        var retryOptions = options.Value.RetryOptions ?? throw new InvalidOperationException();
+        if (retryOptions.Mode != RetryMode.Exponential)
+        {
+            throw new InvalidOperationException();
```
vicancy, line 21: nit: defensive code returns ArgumentException?

src/Microsoft.Azure.SignalR.Management/Resilient/ExponentialBackoffPolicy.cs
```diff
@@ -0,0 +1,44 @@
...
+        _maxRetries = retryOptions.MaxRetries;
+        _minDelay = retryOptions.Delay;
+        _maxDelay = retryOptions.MaxDelay;
+    }
+    public IEnumerable<TimeSpan> GetDelays()
+    {
+        var lastDelay = TimeSpan.MinValue;
+        for (var i = 0; i < _maxRetries; i++)
+        {
+            if (lastDelay > _maxDelay)
```
vicancy, line 32: `>=`?

src/Microsoft.Azure.SignalR.Management/Resilient/FixedBackoffPolicy.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace Microsoft.Azure.SignalR.Management;
+
+internal class FixedBackOffPolicy : IBackOffPolicy
+{
+    private readonly int _maxRetries;
+    private readonly TimeSpan _delay;
+    public FixedBackOffPolicy(IOptions<ServiceManagerOptions> options)
+    {
+        var retryOptions = options.Value.RetryOptions ?? throw new InvalidOperationException();
```
vicancy, line 16: same here, argumentException?

src/Microsoft.Azure.SignalR.Management/Resilient/RetryHttpMessageHandler.cs
```diff
@@ -0,0 +1,76 @@
...
+        IEnumerator<TimeSpan>? delays = null;
+        do
+        {
+            Exception? ex;
+            try
+            {
+                var response = await base.SendAsync(request, cancellationToken);
+                if (_isTransientError(response.StatusCode))
+                {
+
```
vicancy, line 39: nit: remove empty line

src/Microsoft.Azure.SignalR.Management/Resilient/RetryHttpMessageHandler.cs
```diff
@@ -0,0 +1,76 @@
...
+using Microsoft.Azure.SignalR.Common;
+
+namespace Microsoft.Azure.SignalR.Management;
+
+#nullable enable
+
+internal class RetryHttpMessageHandler : DelegatingHandler
+{
+    private readonly IBackOffPolicy _retryDelayProvider;
+    private readonly Func<HttpStatusCode, bool> _isTransientError;
```
vicancy, line 19: naming: `_canRetry`?

src/Microsoft.Azure.SignalR.Management/Resilient/RetryHttpMessageHandler.cs
```diff
@@ -0,0 +1,76 @@
...
+            {
+                // Thrown by our timeout handler
+                ex = operationCanceledException;
+            }
+            delays ??= _retryDelayProvider.GetDelays().GetEnumerator();
+            if (!delays.MoveNext())
+            {
+                if (exceptions == null)
+                {
+                    throw ex;
```
vicancy, line 60: is it better to always throw `AzureSignalRRuntimeException` with innerException as `HttpRequestException` or `AggregatedException`

src/Microsoft.Azure.SignalR.Management/DependencyInjectionExtensions.cs
Y-Sindo: To keep backward-compatibility. According to [HttpClient API doc](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.sendasync?view=net-7.0), in case of timeout, different exceptions are thrown on different .NET implementations. I don't want to imitate all the implementations, so just let the original `HttpClient` to handle the timeout if no RetryOptions is set.

src/Microsoft.Azure.SignalR.Management/Resilient/TimeoutHttpMessageHandler.cs
```diff
@@ -0,0 +1,47 @@
...
+internal class TimeoutHttpMessageHandler : DelegatingHandler
+{
+    private readonly TimeSpan _timeout;
+    public TimeoutHttpMessageHandler(IOptions<ServiceManagerOptions> serviceManagerOptions)
+    {
+        var options = serviceManagerOptions.Value;
+        if (options.RetryOptions == null)
+        {
+            // Timeout handled by HttpClient for backward compatibility
+            _timeout = Timeout.InfiniteTimeSpan;
```
vicancy, line 23: use another boolean to configure enable the handler or not?

src/Microsoft.Azure.SignalR.Management/Resilient/RetryHttpMessageHandler.cs
```diff
@@ -0,0 +1,75 @@
...
+                $"Response status code does not indicate success: {(int)response.StatusCode} ({response.ReasonPhrase})");
+                    ex = new AzureSignalRRuntimeException(request.RequestUri?.ToString(), innerException);
+                    response.Dispose();
+                }
+                else
+                {
+                    return response;
+                }
+            }
+            catch (TaskCanceledException operationCanceledException) when (!cancellationToken.IsCancellationRequested && operationCanceledException.InnerException is TimeoutException)
```
vicancy, line 49: I am not sure if it is too specific. Is it possible that InnerException change later to not TimeoutException? Will that changer be aware of that the change will break retry?

src/Microsoft.Azure.SignalR.Management/Resilient/TimeoutHttpMessageHandler.cs
Y-Sindo: fixed

vicancy (approved):

PR #1845, opened 2023-10-17
Fix DI issue in EndpointRouterDecorator

bjqian:
The `EndpointRouterDecorator` is injected before the `DefaultEndpointRouter` in our example. To avoid the DI issue when customer partially override the `EndpointRouterDecorator`, DI `options` too.


bjqian:
Change to use a new ctor to avoid binary breaking change

src/Microsoft.Azure.SignalR/EndpointRouters/EndpointRouterDecorator.cs
```diff
@@ -15,6 +16,11 @@ public EndpointRouterDecorator(IEndpointRouter router = null)
             _inner = router ?? new DefaultEndpointRouter(null);
         }
 
+        public EndpointRouterDecorator(IEndpointRouter router = null, IOptions<ServiceOptions> options = null)
```
vicancy, line 19: change to ctors without default value?
```
public EndpointRouterDecorator()
public EndpointRouterDecorator(IEndpointRouter router)
public EndpointRouterDecorator(IOptions<ServiceOptions> options)
```
```

bjqian:
Remove `endpointRouteMode` from serviceOptions

PR #1850, opened 2023-10-18
Add mutli ack to AckHandler

xingsy97:
1. Add multiple ack for `AckHandler`.
2. Refactor it

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/Utilities/AckHandler.cs
+++ b/src/Microsoft.Azure.SignalR.Common/Utilities/AckHandler.cs
@@ -5,44 +5,42 @@ using System.Threading.Tasks;
 using Microsoft.Azure.SignalR.Common;
 using Microsoft.Azure.SignalR.Protocol;
 
+#nullable enable
+
 namespace Microsoft.Azure.SignalR
 {
     internal sealed class AckHandler : IDisposable
     {
-        private readonly ConcurrentDictionary<int, AckInfo> _acks = new ConcurrentDictionary<int, AckInfo>();
-
+        private readonly ConcurrentDictionary<int, IAckInfo> _acks = new();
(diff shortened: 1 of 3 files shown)


vicancy:
> Refactor it to make it consistent with [RT version]

Is this a breaking change from SDK's perspective of view?

xingsy97:
> > Refactor it to make it consistent with [RT version]
> 
> Is this a breaking change from SDK's perspective of view?

`AckHandler` is an internal class. I think it's not.

vicancy:
>While SDK AckHandler will set its result with AckStatus.Timeout rather than cancelling it

So now ackHandler callers needs to catch the CancelldException?

xingsy97:
> > While SDK AckHandler will set its result with AckStatus.Timeout rather than cancelling it
> 
> So now ackHandler callers needs to catch the CancelldException?

No, In SDK, AckHandler won't cancel the task when timeout. So catching `CancelldException` is not needed. It will set a result whose value is `AckStatus.Timeout`

src/Microsoft.Azure.SignalR.Common/Utilities/AckHandler.cs
```diff
@@ -56,62 +54,228 @@ public static bool HandleAckStatus(IAckableMessage message, AckStatus status)
...
+            // `foreach (var (id, ack) in _acks)` cannot pass under netstandard2.0
+            foreach (var item in _acks)
+            {
+                var id = item.Key;
+                var ack = item.Value;
+                if (utcNow > ack.TimeoutAt)
+                {
+                    if (_acks.TryRemove(id, out _))
+                    {
+                        // This part is slightly different from RT.
```
vicancy, line 128: remove RT related comments

vicancy (approved):

PR #1855, opened 2023-10-24
Set product info in DI instead of Rest Client

Y-Sindo:


src/Microsoft.Azure.SignalR.Management/DependencyInjectionExtensions.cs
```diff
@@ -221,6 +230,10 @@ private static IServiceCollection AddRestClientFactory(this IServiceCollection s
             static bool IsTransientErrorForNonMessageApi(HttpStatusCode code) =>
                 code >= HttpStatusCode.InternalServerError ||
                 code == HttpStatusCode.RequestTimeout;
+
+            static void ConfigureProduceInfo(IServiceProvider sp, HttpClient client) =>
+                client.DefaultRequestHeaders.Add(Constants.AsrsUserAgent, sp.GetRequiredService<IOptions<ServiceManagerOptions>>().Value.ProductInfo);
```
vicancy, line 235: could ProductInfo be null?

Y-Sindo: No

vicancy: Give it a default value if it is null?

Y-Sindo: Fixed.

vicancy (approved):

PR #1860, opened 2023-10-25
Remove net3.0 support

vicancy:


src/Microsoft.Azure.SignalR/Microsoft.Azure.SignalR.csproj
```diff
@@ -7,7 +7,7 @@
   </PropertyGroup>
   
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;netcoreapp3.0;net5.0;net6.0;net7.0</TargetFrameworks>
```
JialinXin, line 10: Shall we remove net 5.0 as well considering net 5.0 is also end of support already.

JialinXin: Maybe also upgrade netsandard2.0 to netstandard2.1?

vicancy: netsandard2.0 supports net framework while netstandard2.1 does not

vicancy: Remove net 5.0? Not for now, Azure.Core also still supports net5.0 https://www.nuget.org/packages/Azure.Core/

JialinXin (approved):

vwxyzh (approved):

PR #1861, opened 2023-10-25
Remove generated REST client which depends on deprecated package.

Y-Sindo:
* For `RestHealthCheckService` (not really used now), the HttpClient timeout is not impacted by user settings.
* For `ServiceManager.IsServiceHeathy`, the http client timeout is impacted by user settings.
* For AAD related REST API calls, because they use a static `HttpClientFactory` instance, they have never impacted by the user settings.


src/Microsoft.Azure.SignalR.Management/HubInstanceFactories/RestHealthCheckService.cs
```diff
@@ -63,32 +61,31 @@ public Task StopAsync(CancellationToken _)
...
+                        isHealthy = await IsServiceHealthy(endpoint);
+                        needRetry = !isHealthy && retry < MaxRetries;
+                        if (needRetry)
+                        {
+                            Log.WillRetryHealthCheck(_logger, endpoint.Endpoint, _retryInterval);
+                            await Task.Delay(_retryInterval);
+                        }
+                        retry++;
+                    } while (needRetry);
+                    endpoint.Online = isHealthy;
```
vicancy, line 82: what happens when the endpoint count hotreloads to 1 while the last check is Online=false? It does not check again so even if the endpoint is online it does not change back to online?

Y-Sindo: Now the checking will continue if it detects any of the endpoint is offline.

PR #1883, opened 2023-12-22
Add AddAzureSignalRWithConnectionName

vicancy:
Step 1 for https://github.com/dotnet/aspire/issues/799

Add a `AddAzureSignalRWithConnectionName` that reads from `ConnectionStrings_{name}`, so that the SDK could integrate with Aspire provisioned configuration. Configurations will also be read from `Azure:SignalR:<connectionName>`
* AddKeyedxxx is not a valid scenario when using Azure SignalR since the service is transparent to the customer when they are using SignalR hubs.


src/Microsoft.Azure.SignalR/ServiceOptionsSetup.cs
```diff
@@ -70,14 +76,14 @@ private TimeSpan GetConfiguredTimeSpanFromSeconds(int? seconds, TimeSpan default
             return defaultValue;
         }
 
-        private (string AppName, string ConnectionString, ServiceEndpoint[] Endpoints, ConfigurableServiceOptions configurableOptions) ParseConfiguration()
+        private (string AppName, string ConnectionString, ServiceEndpoint[] Endpoints, ConfigurableServiceOptions configurableOptions) ParseConfiguration(string connectionName)
         {
-            var sectionKey = Constants.Keys.AzureSignalRSectionKey;
+            var sectionKey = string.IsNullOrEmpty(connectionName) ? Constants.Keys.AzureSignalRSectionKey : $"{Constants.Keys.AzureSignalRSectionKey}:{connectionName}";
```
Y-Sindo, line 81: An empty `connectionName` should be treated as `Azure:SignalR:`? `Azure:SignalR:` is also valid section name.

src/Microsoft.Azure.SignalR/DependencyInjectionExtensions.cs
```diff
@@ -36,6 +36,26 @@ public static ISignalRServerBuilder AddAzureSignalR(this ISignalRServerBuilder b
             return builder.AddAzureSignalR<ServiceOptionsSetup>();
         }
 
+        /// <summary>
+        /// Add Azure SignalR with specified connection name, the connection string could be read from ConnectionStrings.{name}, the settings are loaded from Azure:SignalR:{name} section
```
Y-Sindo: `ConnectionStrings.{name}`  -> `ConnectionStrings_{name}`?

src/Microsoft.Azure.SignalR/ServiceOptionsSetup.cs
vicancy: empty connectionName is considered as the default case, added a null check for AddAzureSignalRWithConnectionName

src/Microsoft.Azure.SignalR/DependencyInjectionExtensions.cs
```diff
@@ -36,6 +36,28 @@ public static ISignalRServerBuilder AddAzureSignalR(this ISignalRServerBuilder b
...
+        /// </summary>
+        /// <param name="builder">The <see cref="ISignalRServerBuilder"/>.</param>
+        /// <param name="name">The name of the Azure SignalR service that settings and connection strings are read from</param>
+        /// <param name="configure">A callback to configure the <see cref="ServiceOptions"/>.</param>
+        /// <returns>The same instance of the <see cref="ISignalRServerBuilder"/> for chaining.</returns>
+        public static ISignalRServerBuilder AddAzureSignalRWithConnectionName(this ISignalRServerBuilder builder, string name, Action<ServiceOptions> configure = null)
+        {
+            if (string.IsNullOrEmpty(name))
+            {
+                throw new ArgumentNullException(nameof(name));
```
Y-Sindo: ArgumentException?

src/Microsoft.Azure.SignalR/DependencyInjectionExtensions.cs
```diff
@@ -36,6 +36,28 @@ public static ISignalRServerBuilder AddAzureSignalR(this ISignalRServerBuilder b
...
+            }
+
+            builder.Services.SetupOptions<ServiceOptions, ServiceOptionsSetup>(s => ActivatorUtilities.CreateInstance<ServiceOptionsSetup>(s, name));
+            if (configure != null)
+            {
+                builder.Services.Configure(configure).PostConfigure<ServiceOptions>(o => o.Validate());
```
Y-Sindo: We should always validate options no matter configure is null or not.

Y-Sindo: Actually I don't think we need to provide the paramter `Action<ServiceOptions>`, because the user can call `builder.Services.Configure(configure)` themselves.

Y-Sindo: https://github.com/dotnet/roslyn-analyzers/blob/main/src/PublicApiAnalyzers/Microsoft.CodeAnalysis.PublicApiAnalyzers.md#rs0026-do-not-add-multiple-public-overloads-with-optional-parameters

Adding multiple public overloads with optional parameter seems not a good practice

vicancy: we don't have method overloads with optional parameter?

Y-Sindo (approved):

PR #1886, opened 2024-01-08
Fix UI Culture parsing bug

### LINKED ISSUE #1872: Culture and UI culture are not applied correctly in Blazor application
When using the .NET cookie (.AspNetCore.Culture) to set culture and UI culture of a Blazor application (as described here https://learn.microsoft.com/en-us/aspnet/core/blazor/globalization-localization?view=aspnetcore-8.0#dynamically-set-the-server-side-culture-by-user-preference), the culture (i.e., numbers/dates formatting) and the UI culture (i.e., language) are not set correctly after the introduction of the Azure SignalR service. Examples:

Without Azure Signal R (expected behavior):
- When cookie value is "c=en|uic=it", english is used for numbers/dates formatting, while italian is used for texts
- When cookie value is "c=it|uic=en, italian is used for numbers/dates formatting,  while english is used for texts

After the introduction of Azure Signal R (with .AddAzureSignalR())
- When cookie value is "c=en|uic=it", english is used for both numbers/dates formatting and texts
- When cookie value is "c=it|uic=en, italian is used for both numbers/dates formatting and texts

-> it seems that with Azure SignalR, only the "c" part of the cookie is used to set both culture and UI culture of the application.

### Further technical details
- .NET version: 7.0.403
- Azure SignalR SDK version: 1.22.0

xingsy97:
Fix Bug #1872
Root cause: https://github.com/Azure/azure-signalr/issues/1872#issuecomment-1839163024

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/Constants.cs
+++ b/src/Microsoft.Azure.SignalR.Common/Constants.cs
@@ -77,7 +77,7 @@ namespace Microsoft.Azure.SignalR
             public const string OriginalPath = "asrs.op";
             public const string ConnectionRequestId = "asrs_request_id";
             public const string RequestCulture = "asrs_lang";
-            public const string RequestUiCulture = "asrs_ui_lang";
+            public const string RequestUICulture = "asrs_ui_lang";
         }
 
         public static class CustomizedPingTimer
--- a/src/Microsoft.Azure.SignalR.Common/Constants.cs
+++ b/src/Microsoft.Azure.SignalR.Common/Constants.cs
@@ -77,6 +77,7 @@ namespace Microsoft.Azure.SignalR
             public const string OriginalPath = "asrs.op";
             public const string ConnectionRequestId = "asrs_request_id";
             public const string RequestCulture = "asrs_lang";
+            public const string RequestUiCulture = "asrs_ui_lang";
         }
 
         public static class CustomizedPingTimer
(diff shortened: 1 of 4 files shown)


src/Microsoft.Azure.SignalR/HubHost/NegotiateHandler.cs
```diff
@@ -80,6 +80,7 @@ public async Task<NegotiationResponse> Process(HttpContext context)
             var claims = BuildClaims(context);
             var request = context.Request;
             var cultureName = context.Features.Get<IRequestCultureFeature>()?.RequestCulture.Culture.Name;
+            var uiCultureName = context.Features.Get<IRequestCultureFeature>()?.RequestCulture.UICulture.Name;
```
JialinXin, line 83: Also add UT?

src/Microsoft.Azure.SignalR.Common/Constants.cs
```diff
@@ -77,6 +77,7 @@ public static class QueryParameter
             public const string OriginalPath = "asrs.op";
             public const string ConnectionRequestId = "asrs_request_id";
             public const string RequestCulture = "asrs_lang";
+            public const string RequestUiCulture = "asrs_ui_lang";
```
Y-Sindo, line 80: Do we need changes in the runtime?

src/Microsoft.Azure.SignalR/HubHost/NegotiateHandler.cs
xingsy97: added

src/Microsoft.Azure.SignalR.Common/Constants.cs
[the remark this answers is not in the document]
xingsy97: updated

vicancy (approved):

xingsy97:
I cannot push more commit to this pr for it's not created on my own repo. Will open a new one.

PR #1943, opened 2024-05-17
Suppress 403 when try getting access key

### LINKED ISSUE #1913: Should not require temporary access key for management SDK in transient mode.
### Describe the bug
When using AAD, management SDK should not require temporary access key in transient mode.

### To Reproduce

* Use AAD auth.
* Use transient mode.
* Only give `REST API owner` for SignalR service.

Management SDK will get consistent 403s from Access Key APIs.

terencefan:
### Suppress 403 errors when try getting access key

The `ServiceManager` is designed to try getting the access key even in `Transient` mode, as it may need to handle negotiate requests.

However, considering some cases where the given `TokenCredential` lacks permission for the `GetAccessKey` API, the error messages from failed attempts can be annoying. Therefore, we have chosen to suppress these errors and only throw the latest exception when necessary.

Fixes #1913


src/Microsoft.Azure.SignalR.AspNet/Middleware/NegotiateMiddleware.cs
```diff
@@ -184,10 +184,10 @@ private Task ProcessNegotiationRequest(IOwinContext owinContext, HostContext con
             {
                 return GenerateClientAccessTokenAsync(provider, context, url, claims);
             }
-            catch (AzureSignalRAccessTokenNotAuthorizedException e)
+            catch (AzureSignalRCredentialUnauthorizedException e)
             {
                 Log.NegotiateFailed(_logger, e.Message);
-                context.Response.StatusCode = 500;
+                context.Response.StatusCode = 403;
```
vicancy: it sounds 500 since it is the app server having no permission not the client?

src/Microsoft.Azure.SignalR.Common/Endpoints/AccessKeyForMicrosoftEntra.cs
```diff
@@ -37,12 +35,14 @@ internal class AccessKeyForMicrosoftEntra : AccessKey
 
         private static readonly TimeSpan GetAccessKeyIntervalWhenUnauthorized = TimeSpan.FromMinutes(5);
 
-        private static readonly TimeSpan GetAccessKeyRetryInterval = TimeSpan.FromSeconds(GetAccessKeyRetryIntervalInSec);
+        private static readonly TimeSpan GetAccessKeyRetryInterval = TimeSpan.FromSeconds(3);
```
vicancy, line 38: why not using const?

src/Microsoft.Azure.SignalR.Common/Exceptions/AzureSignalRAccessTokenNotAuthorizedException.cs
```diff
@@ -1,11 +1,14 @@
...
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
+
 namespace Microsoft.Azure.SignalR.Common
 {
     /// <summary>
     /// The exception throws when AccessKey is not authorized.
     /// </summary>
+    [Obsolete("Use AzureSignalRCredentialUnauthorizedException instead.")]
```
vicancy: why? is using this exception acceptable? This is a breaking change?

src/Microsoft.Azure.SignalR.Common/Endpoints/AccessKeyForMicrosoftEntra.cs
terencefan: This value hasn't been used elsewhere.

And I think `private static readonly` is already a sort of `const`?

src/Microsoft.Azure.SignalR.AspNet/Middleware/NegotiateMiddleware.cs
terencefan: sounds correct.

src/Microsoft.Azure.SignalR.Common/Exceptions/AzureSignalRAccessTokenNotAuthorizedException.cs
terencefan: Use the same exception but add a new constructor.

src/Microsoft.Azure.SignalR.Common/Exceptions/AzureSignalRAccessTokenNotAuthorizedException.cs
```diff
@@ -1,17 +1,28 @@
...
+        [Obsolete("use (TokenCredential, Exception) instead.")]
+        public AzureSignalRAccessTokenNotAuthorizedException(string message) : base(message)
+        {
+        }
+
+        /// <summary>
         /// Initializes a new instance of the <see cref="AzureSignalRAccessTokenNotAuthorizedException"/> class.
         /// </summary>
-        public AzureSignalRAccessTokenNotAuthorizedException(string message) : base(message)
+        public AzureSignalRAccessTokenNotAuthorizedException(TokenCredential credential, Exception innerException) :
```
vicancy: use string name?

src/Microsoft.Azure.SignalR.Common/Exceptions/AzureSignalRAccessTokenNotAuthorizedException.cs
```diff
@@ -1,17 +1,28 @@
...
 {
     /// <summary>
     /// The exception throws when AccessKey is not authorized.
     /// </summary>
     public class AzureSignalRAccessTokenNotAuthorizedException : AzureSignalRException
     {
-         /// <summary>
+        private const string Postfix = " appears to lack the permission to generate access tokens, see innerException for more details.";
+
+        [Obsolete("use (TokenCredential, Exception) instead.")]
```
vicancy: update the description

src/Microsoft.Azure.SignalR.Common/Endpoints/AccessKeyForMicrosoftEntra.cs
```diff
@@ -37,20 +35,23 @@ internal class AccessKeyForMicrosoftEntra : AccessKey
...
         private DateTime _lastUpdatedTime = DateTime.MinValue;
 
-        public bool Authorized
+        public bool IsAuthorized
         {
             get => _isAuthorized;
             private set
             {
                 _lastUpdatedTime = DateTime.UtcNow;
+                _lastException = value ? null : _lastException;
```
vwxyzh: readable:
```cs
if (value)
{
  _lastException = null;
}
```

vicancy (approved):

PR #1964, opened 2024-07-11
Add connection FlowControl messages

terencefan:
Add messages for client connection migration.

`FlowControlMessage` with `Pause`, `PauseAck` and `Resume` state.

image

### DIFF
--- a/src/Microsoft.Azure.SignalR.Protocols/ServiceMessage.cs
+++ b/src/Microsoft.Azure.SignalR.Protocols/ServiceMessage.cs
@@ -7,7 +7,7 @@ using MessagePack;
 
 namespace Microsoft.Azure.SignalR.Protocol
 {
-    public enum OutgoingTaskControlState
+    public enum FlowControlState
     {
         /// <summary>
         /// Send from ASRS to the app server.
@@ -192,19 +192,14 @@ namespace Microsoft.Azure.SignalR.Protocol
     /// The ASRS send it to the source server connection to initiate a client connection migration.
     /// Indicates that incoming messages are blocked.
     /// </summary>
-    public class OutgoingTaskControlMessage : ConnectionMessage
+    /// <remarks>
+    ///
+    /// </remarks>
+    /// <param name="connectionId">The client connection ID</param>
+    /// <param name="state">The state of control message</param>
+    public class FlowControlMessage(string connectionId, FlowControlState state) : ConnectionMessage(connectionId)
     {
-        public OutgoingTaskControlState State { get; }
-
-        /// <summary>
-        ///
-        /// </summary>
-        /// <param name="connectionId">The client connection ID</param>
-        /// <param name="state">The state of control message</param>
-        public OutgoingTaskControlMessage(string connectionId, OutgoingTaskControlState state) : base(connectionId)
-        {
-            State = state;
-        }
+        public FlowControlState State { get; } = state;
     }
 
     /// <summary>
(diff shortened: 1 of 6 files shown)


src/Microsoft.Azure.SignalR.Protocols/ServiceMessage.cs
```diff
@@ -126,6 +126,50 @@ public interface IAckableMessage
         int AckId { get; set; }
     }
 
+    /// <summary>
+    /// The ASRS send it to the source server connection to initiate a client connection migration.
+    /// Indicates that incoming messages are blocked.
+    /// The app server should block outgoing messages and reply with a <see cref="MigrateConnectionAckMessage"/>.
+    /// </summary>
+    public class MigrateConnectionRequestMessage : ConnectionMessage
```
vicancy: So this message could also be shared by every migration scenario

vicancy: How about... a `MigrateConnectionMessage` with state: `Requested` `Accepted` `Started`

vwxyzh: what about `Pause`, `PauseAck`, `Resume`
migrating client is our purpose, the real actions are pausing and resuming.

vwxyzh: and we can also use pausing and resuming to control traffic for streaming

terencefan: updated to `OutgoingTaskControlMessage` with `Pause`, `PauseAck` and `Resume` state.

src/Microsoft.Azure.SignalR.Protocols/ServiceMessage.cs
```diff
@@ -157,11 +189,22 @@ internal void ReadExtensionMembers(ref MessagePackReader reader)
     }
 
     /// <summary>
-    /// Interface of ack-able message 
+    /// The ASRS send it to the source server connection to initiate a client connection migration.
+    /// Indicates that incoming messages are blocked.
     /// </summary>
-    public interface IAckableMessage
+    public class OutgoingTaskControlMessage : ConnectionMessage
```
vicancy: how about `FlowControlMessage`?

terencefan: updated

src/Microsoft.Azure.SignalR.Protocols/ServiceMessage.cs
```diff
@@ -7,6 +7,35 @@
 
 namespace Microsoft.Azure.SignalR.Protocol
 {
+    public enum FlowControlState
```
vwxyzh: i do not think they are states, they are operations.

src/Microsoft.Azure.SignalR.Protocols/ServiceMessage.cs
```diff
@@ -157,11 +189,17 @@ internal void ReadExtensionMembers(ref MessagePackReader reader)
...
+    /// The ASRS send it to the source server connection to initiate a client connection migration.
+    /// Indicates that incoming messages are blocked.
     /// </summary>
-    public interface IAckableMessage
+    /// <remarks>
+    ///
+    /// </remarks>
+    /// <param name="connectionId">The client connection ID</param>
+    /// <param name="state">The state of control message</param>
+    public class FlowControlMessage(string connectionId, FlowControlState state) : ConnectionMessage(connectionId)
```
vicancy: Do we need Connection prefix? `ConnectionFlowControlMessage` since connectionId is a required parameter?

vicancy (approved):

PR #1971, opened 2024-07-18
Add IClientConnection interface

terencefan:


src/Microsoft.Azure.SignalR.AspNet/ServerConnections/ServiceConnection.cs
```diff
@@ -252,7 +251,7 @@ private async Task OnConnectedAsyncCore(ClientConnectionContext clientContext, O
         }
     }
 
-    private void ProcessOutgoingMessages(ClientConnectionContext clientContext, ConnectionDataMessage connectionDataMessage)
+    private void ProcessConnectionDataMessage(ClientConnectionContext clientContext, ConnectionDataMessage connectionDataMessage)
```
vwxyzh: i think `ProcessOutgoingMessages` is better.

vwxyzh (approved):

PR #2019, opened 2024-08-07
Fix emulator response unexpected Header:x-ms-error-code

### LINKED ISSUE #1912: ClientManager.ConnectionExistsAsync throws exception on expected 404 response from emulator
### Describe the bug
I'm calling "Microsoft.Azure.SignalR.Management.ClientManager.ConnectionExistsAsync" for non-existing connection, while talking to emulator. Method call throws exception, saying that emulator responded "404", but it seems like expected behaviour, so I guess method call shouldn't throw exception. Debugging deeply, I see that response doesn't contain header "x-ms-error-code", which management sdk is expecting, so it might be an emulator issue though.

Tested with azure cloud SignalR, works fine, not throwing exception, so problem is probably in the emulator.

### To Reproduce

1. Run emulator
2. Create CLI app with "Microsoft.Azure.SignalR.Management" referenced
3. Create instance of ServiceHubContext
4. Call `hubContext.ClientManager.ConnectionExistsAsync("non_existing_connection");`

### Exceptions
```
Microsoft.Azure.SignalR.Common.AzureSignalRInaccessibleEndpointException: Unable to access SignalR service. Response status code does not indicate success: 404 (Not Found) Please make sure the endpoint or DNS setting is correct. Request Uri: http://localhost:8888/api/hubs/signalrhub/connections/xxxxxxxxxxxxxxxxxxxxxx?api-version=2022-06-01
 ---> System.Net.Http.HttpRequestException: Response status code does not indicate success: 404 (Not Found)
   --- End of inner exception stack trace ---
   at Microsoft.Azure.SignalR.RestClient.ThrowExceptionOnResponseFailureAsync(HttpResponseMessage response) in ...\azure-signalr-1.24.0\src\Microsoft.Azure.SignalR.Common\Utilities\RestClient.cs:line 106
   at Microsoft.Azure.SignalR.RestClient.SendAsyncCore(String httpClientName, RestApiEndpoint api, HttpMethod httpMethod, String methodName, Object[] args, Func`2 handleExpectedResponseAsync, CancellationToken cancellationToken) in ...\azure-signalr-1.24.0\src\Microsoft.Azure.SignalR.Common\Utilities\RestClient.cs:line 138
   at Microsoft.Azure.SignalR.Management.RestHubLifetimeManager`1.ConnectionExistsAsync(String connectionId, CancellationToken cancellationToken) in ...\azure-signalr-1.24.0\src\Microsoft.Azure.SignalR.Management\RestHubLifetimeManager.cs:line 273
   ...
```

### Further technical details
- SignalR SDK version: 1.24.0
- Microsoft.AspNetCore.SignalR: 1.0.0
- Microsoft.AspNet.SignalR.Client: 2.4.3

phanthaiduong22:
 - [Describe the bug](https://github.com/Azure/azure-signalr/issues/1912#:~:text=edited-,Describe%20the%20bug,-I%27m%20calling%20%22Microsoft)

 - How to fix:
Add the expected header `x-ms-error-code` for responses from the API that checks for existence in the emulator (CheckConnectionExistence, CheckGroupExistence, CheckUserExistence). The expected value of x-ms-error-code should be the same as what is defined in management to avoid throwing exceptions in this function.
https://github.com/Azure/azure-signalr/blob/fa7b063f4639d988108ea0d20586b1c06e4d5bdb/src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs#L276

Fixes #1912


zackliu:
There're some other methods that can also return `NotFound`. They need to add `ErrorCode` as well.

src/Microsoft.Azure.SignalR.Emulator/Common/Constants.cs
```diff
@@ -37,5 +37,18 @@ public static class Headers
             public const string AsrsClientCertThumbprint = AsrsHeaderPrefix + "Client-Cert-Thumbprint";
             public const string AsrsConnectionGroups = AsrsHeaderPrefix + "Connection-Group";
         }
+
+        public static class ErrorCodes
+        {
+            public const string WarningConnectionNotExisted = "Warning.Connection.NotExisted";
```
zackliu: What about add some `Enum` for `level`, `scope`, `errorKind` and use a method to build a ErrorCode like `level.scope.errorKind` instead of fully using a constant?

zackliu:
Try to add some unit tests under `test/Microsoft.Azure.SignalR.Emulator.Tests`

src/Microsoft.Azure.SignalR.Emulator/Controllers/SignalRServiceEmulatorWebApi.cs
```diff
@@ -220,11 +224,13 @@ public SignalRServiceEmulatorWebApi(DynamicHubContextStore store, ILogger<Signal
             }
 
             if (_store.TryGetLifetimeContext(GetInternalHubName(application, hub), out var c))
-            { 
+            {
                 c.UserGroupManager.RemoveConnectionFromAllGroups(connectionId);
                 return Ok();
             }
 
+            Response.SetMsErrorCodeHeader(Constants.ErrorCodes.Error.ConnectionNotExisted);
```
zackliu: It returns `OK`, so I don't think it need to set this header. But you missed `RemoveConnectionFromGroup`

src/Microsoft.Azure.SignalR.Emulator/HubEmulator/DynamicHubContext.cs
```diff
@@ -24,10 +24,14 @@ internal class DynamicHubContext
             ConnectionHandler = connectionHandler;
         }
 
+        public DynamicHubContext()
```
zackliu: Remove if not used

src/Microsoft.Azure.SignalR.Emulator/Controllers/SignalRServiceEmulatorWebApi.cs
```diff
@@ -26,10 +24,10 @@ internal class SignalRServiceEmulatorWebApi : SignalRServiceWebApiDefinition
     {
         private const string HubPattern = "^[A-Za-z][A-Za-z0-9_`,.[\\]]{0,127}$";
         private const string GroupPattern = "^\\S{1,1024}$";
-        private readonly DynamicHubContextStore _store;
+        private readonly IDynamicHubContextStore _store;
```
zackliu, line 27: As you changed to use interface, you should change every usage of this concreate class to interface. As well as `AddSignalREmulator`

src/Microsoft.Azure.SignalR.Emulator/Properties/AssemblyInfo.cs
```diff
@@ -1,3 +1,7 @@
...
+[assembly: InternalsVisibleTo("Microsoft.Azure.SignalR.Emulator.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100f33a29044fa9d740c9b3213a93e57c84b472c84e0b8a0e1ae48e67a9f8f6de9d5f7f3d52ac23e48ac51801f1dc950abe901da34d2a9e3baadb141a17c77ef3c565dd5ee5054b91cf63bb3c6ab83f72ab3aafe93d0fc3c2348b764fafb0b1c0733de51459aeab46580384bf9d74c4e28164b7cde247f891ba07891c9d872ad2bb")]
+[assembly: InternalsVisibleTo("DynamicProxyGenAssembly2, PublicKey=0024000004800000940000000602000000240000525341310004000001000100c547cac37abd99c8db225ef2f6c8a3602f3b3606cc9891605d02baa56104f4cfc0734aa39b93bf7852f7d9266654753cc297e7d2edfe0bac1cdcf9f717241550e0a7b191195b7667bb4f64bcb8e2121380fd1d9d46ad2d92d2d15605093924cceaf74c4861eff62abf69b9291ed0a340e113be11e6a7d3113e92484cf7045cc7")]
```
vicancy, line 7: what is this for?

zackliu: For mock

src/Microsoft.Azure.SignalR.Emulator/HubEmulator/DynamicHubContext.cs
phanthaiduong22: Oh I need this for unit tests, since I changed property `UserGroupManager` to virtual and its required parameterless constructor

zackliu: What about making it `protected`

src/Microsoft.Azure.SignalR.Emulator/Properties/AssemblyInfo.cs
vicancy: DynamicProxyGenAssembly2?

vicancy (approved):

phanthaiduong22:
hi @zackliu 

I have resolved the review. Can u help double check that I can use: `Error.Connection.NotExisted` for both functions: `RemoveConnectionFromGroup` and `AddConnectionToGroup`. I have self-tested the sample (Management) with the `emulator` after resolving about changing from `concrete class` to `interface`. It works now. If everything is ok, I think we can merge

test/Microsoft.Azure.SignalR.Emulator.Tests/Controllers/SignalRServiceEmulatorWebApiTests.cs
```diff
@@ -0,0 +1,423 @@
...
+using Microsoft.Azure.SignalR.Emulator.HubEmulator;
+using Microsoft.Extensions.Logging;
+using Moq;
+using Xunit;
+
+namespace Microsoft.Azure.SignalR.Emulator.Tests.Controllers
+{
+    public class SignalRServiceEmulatorWebApiTests
+    {
+        private readonly Mock<IDynamicHubContextStore> storeMock;
```
zackliu: Move normal fields below const field https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1203.md. 
And our naming convention is using underscore for private field like `_storeMock`.

test/Microsoft.Azure.SignalR.Emulator.Tests/Controllers/SignalRServiceEmulatorWebApiTests.cs
```diff
@@ -0,0 +1,423 @@
...
+        private readonly Mock<ILogger<SignalRServiceEmulatorWebApi>> loggerMock;
+        private readonly SignalRServiceEmulatorWebApi controller;
+
+        // Constants
+        private const string testHub = "testHub";
```
zackliu: Constant fields should follow CamelCase e.g. `TestHub`

test/Microsoft.Azure.SignalR.Emulator.Tests/Controllers/SignalRServiceEmulatorWebApiTests.cs
```diff
@@ -0,0 +1,423 @@
...
+
+        // CheckConnectionExistence Tests
+        [Fact]
+        public void CheckConnectionExistenceValidConnectionReturnsOk()
+        {
+            // Arrange
+            var connectionContext = new DefaultConnectionContext();
+            connectionContext.ConnectionId = testConnectionId;
+
+            var mockLoggerFactory = new Mock<ILoggerFactory>();
```
zackliu: Just a hint. If the logger/loggerFactory is not used in test, you can also use `NullLoggerFactory.Instance` to simplify.

zackliu (approved):

src/Microsoft.Azure.SignalR.Emulator/Properties/AssemblyInfo.cs
phanthaiduong22: its auto generated and for mocking purpose for unit test. https://stackoverflow.com/a/17179923/10575423

`
DynamicProxyGenAssembly2 is a temporary assembly built by mocking systems that use CastleProxy like Moq or NSubsitute. It is generated when the mock is needed and disposed of after the tests are finished.
`

PR #2021, opened 2024-08-13
Fix the ack message issue when MapHub for the same hub is called multiple times

### LINKED ISSUE #2003: AddToGroupAsync throws a TimeoutException exception when MapHub is called twice for the same hub
### Describe the bug
**Groups.AddToGroupAsync()** throws a **TimeoutException** approximately half the time when MapHub is called twice for the same hub. This issue is caused by changes introduced in #1779. It can be reproduced with Microsoft.Azure.SignalR version 1.21.3 or later works fine with version 1.21.2.

I'm unsure if calling **MapHub** twice for the same hub is allowed. If it isn't, this behavior should be documented.

### To Reproduce
Update the ChatRoom sample project as follows:

Program.cs
```
app.MapHub<ChatSampleHub>("chat"); 
app.MapHub<ChatSampleHub>("chat/v2");
```

ChatSampleHub.OnConnectedAsync
`await Groups.AddToGroupAsync(Context.ConnectionId, $"{Guid.NewGuid()}");`

### Exceptions (if any)
fail: Microsoft.AspNetCore.SignalR.HubConnectionHandler[1]
      Error when dispatching 'OnConnectedAsync' on hub.
      System.TimeoutException: Ack-able message Microsoft.Azure.SignalR.Protocol.JoinGroupWithAckMessage(ackId: 1) timed out.
         at Microsoft.Azure.SignalR.AckHandler.HandleAckStatus(IAckableMessage message, AckStatus status)
         at Microsoft.Azure.SignalR.ServiceConnectionContainerBase.WriteAckableMessageAsync(ServiceMessage serviceMessage, CancellationToken cancellationToken)
         at Microsoft.Azure.SignalR.MultiEndpointMessageWriter.<>c__DisplayClass10_0.<<WriteSingleResultAckableMessage>b__0>d.MoveNext()

### Further technical details
- Your Azure SignalR SDK version - 1.21.3
- .NET Version: 8

```[tasklist]
### Tasks
```

vicancy:
Fix https://github.com/Azure/azure-signalr/issues/2003

It is possible that the multiple MapHub calls the same hub, so that ack messages could be received by another instance of ServiceConnectionContainer

Use the ack handler singleton to allow ack message to be acked by another container instance


vwxyzh:
whats behavior of standalone signalr?
* Two separated hubs with same methods?
* One hub with two paths?

vwxyzh (approved):

vicancy:
Self-host signalr shares the hub with 2 endpoints. We keep the same behavior.

PR #2023, opened 2024-08-15
Sticky to one service connection if not in async scope

### LINKED ISSUE #1655: Azure SignalR doesn't process messages in order under high load
### Describe the bug
SignalR messages sent to the client are processed in order (when sent from the same source to the same clients). This behaviour is observed when hosting signalr yourself. However when Azure SignalR is used then some messages are received out of order when there is a high volume of messages. I couldn't find anything about this in the docs, but according to @davidfowl this should be the case, https://twitter.com/bacobart/status/1557296764100743168.

### Expected Behavior

When adding AzureSignalR messages received on the client are in order.

### Steps To Reproduce

Reproduction at https://github.com/bacobart/azure-signalr-out-of-order-repro

Without adding .AddAzureSignalR() in the server messages are processed in order as expected. When .AddAzureSignalR() is added some messages are processed out of order.

### Exceptions (if any)
none

### Further technical details
.net 6.0.400
server and client are 6.0.8

vicancy:
Fix https://github.com/Azure/azure-signalr/issues/1655

- [x] Add test coverage

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
+++ b/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
@@ -175,6 +175,9 @@ internal abstract partial class ServiceConnectionBase : IServiceConnection
                         _ = UpdateAzureIdentityAsync(key, syncTimer);
                     }
                     await ProcessIncomingAsync(connection);
+
+                    // mark the status as Disconnected so that no one will write to this connection anymore
+                    Status = ServiceConnectionStatus.Disconnected;
                 }
                 finally
                 {
@@ -195,10 +198,7 @@ internal abstract partial class ServiceConnectionBase : IServiceConnection
             finally
             {
                 // wait until all the connections are cleaned up to close the outgoing pipe
-                // mark the status as Disconnected so that no one will write to this connection anymore
                 // Don't allow write anymore when the connection is disconnected
-                Status = ServiceConnectionStatus.Disconnected;
-
                 await _writeLock.WaitAsync();
                 try
                 {
--- a/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
+++ b/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
@@ -175,12 +175,11 @@ internal abstract partial class ServiceConnectionBase : IServiceConnection
                         _ = UpdateAzureIdentityAsync(key, syncTimer);
                     }
                     await ProcessIncomingAsync(connection);
-
-                    // mark the status as Disconnected so that no one will write to this connection anymore
-                    Status = ServiceConnectionStatus.Disconnected;
                 }
                 finally
                 {
+                    // mark the status as Disconnected so that no one will write to this connection anymore
+                    Status = ServiceConnectionStatus.Disconnected;
                     syncTimer?.Stop();
 
                     // when ProcessIncoming completes, clean up the connection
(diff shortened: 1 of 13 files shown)


vwxyzh:
I think partition is better.

src/Microsoft.Azure.SignalR.Protocols/ConnectionMessage.cs
```diff
@@ -171,6 +171,8 @@ public ConnectionDataMessage(string connectionId, ReadOnlySequence<byte> payload
         /// Gets or sets the payload is partial or not.
         /// </summary>
         public bool IsPartial { get; set; }
+
+        public byte PartitionKey { get; }
```
vwxyzh: why not `public byte PartitionKey => GeneratePartitionKey(ConnectionId);`

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionBase.cs
```diff
@@ -175,6 +175,9 @@ public async Task StartAsync(string target = null)
                         _ = UpdateAzureIdentityAsync(key, syncTimer);
                     }
                     await ProcessIncomingAsync(connection);
+
+                    // mark the status as Disconnected so that no one will write to this connection anymore
+                    Status = ServiceConnectionStatus.Disconnected;
```
zackliu: Any reason drop by exception doesn't treat as Disconnected?

src/Microsoft.Azure.SignalR.Protocols/GroupMessage.cs
```diff
@@ -141,13 +152,14 @@ public UserLeaveGroupMessage(string userId, string groupName, ulong? tracingId =
             UserId = userId;
             GroupName = groupName;
             TracingId = tracingId;
+            PartitionKey = GeneratePartitionKey("g." + groupName);
```
zackliu: Why only this partition key contains a "g."

src/Microsoft.Azure.SignalR.Protocols/ConnectionMessage.cs
zackliu: To avoid duplicate calculation?

vicancy: because I don't want a calculation on every get

vicancy: use `=>` for less size

src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -24,11 +25,16 @@ internal abstract class ServiceConnectionContainerBase : IServiceConnectionConta
...
 
+        private static readonly TimeSpan MessageWriteRetryDelay = TimeSpan.FromMilliseconds(200);
+        private static readonly int MessageWriteMaxRetry = 3;
+
         // Give (interval * 3 + 1) delay when check value expire.
         private static readonly long DefaultServersPingTimeoutTicks = Stopwatch.Frequency * ((long)Constants.Periods.DefaultServersPingInterval.TotalSeconds * 3 + 1);
 
         private static readonly Tuple<string, long> DefaultServersTagContext = new Tuple<string, long>(string.Empty, 0);
 
+        private readonly ConcurrentDictionary<byte, WeakReference<IServiceConnection>> _partitionedCache = new();
```
vwxyzh: i think we can initialize the dictionary, since we know all of the keys.
and we can use `Dictionary<byte, StrongBox<WeakReference<IServiceConnection>>>`.

vwxyzh (approved):

PR #2037, opened 2024-09-06
Adding aspire apphost to samples

vicancy:


vicancy:
Closing in favour of #2240, which carries this work forward against the current tree.

Comparing the two: #2240 touches 14 of the 20 files here, with the same AppHost project, solution entry, workflow and `ChatSample` changes. The six that are not carried over are `ChatSample.Net60`, `ChatSample.Net70` and the two screenshots under `samples/images/` - all of which have since been removed from `dev`. That is also why this branch now shows as conflicted: it edits files that no longer exist. Nothing in #2240 references the removed screenshots, so no content is lost by closing this.

PR #2061, opened 2024-10-23
Remove MicrosoftEntraAccessKey.Rest.cs

terencefan:
- Remove unused code
- Merge other code to `MicrosoftEntraAccessKey.cs`
- Add test case for normal scenario


src/Microsoft.Azure.SignalR.Common/Auth/MicrosoftEntra/MicrosoftEntraAccessKey.cs
```diff
@@ -5,19 +5,22 @@
...
+using System.Net.Http.Headers;
 using System.Security.Claims;
 using System.Threading;
 using System.Threading.Tasks;
 
 using Azure.Core;
 
 using Microsoft.Azure.SignalR.Common;
-
+using Newtonsoft.Json;
```
vwxyzh: can we switch to `System.Text.Json`?

vwxyzh (approved):

PR #2066, opened 2024-10-30
Add IAccessKey for future refactor

terencefan:
 - Detail 1
 - Detail 2


src/Microsoft.Azure.SignalR.Common/Auth/AccessKey.cs
```diff
@@ -4,44 +4,39 @@
 using System;
 using System.Collections.Generic;
 using System.Security.Claims;
+using System.Security.Cryptography.X509Certificates;
```
vicancy: who uses this namespace?

terencefan: removed

vicancy (approved):

PR #2067, opened 2024-10-31
Add culture info manager

xingsy97:
Fix #2038

PR #2068, opened 2024-10-31
Rewrite culture handling logic

### LINKED ISSUE #2038: Blazor Server Side Custom CultureInfo getting overridden when using AzureSignalR
### Describe the bug
I notice when running the application it uses my custom culture for half a second, then updates and goes back to the default CultureInfo properties for whatever culture i'm using.
the custom cultures that I want to use work correctly and don't get overridden. Why does adding AzureSignalR override my custom cultures?

More detail information, see [Blazor Server Side Custom CultureInfo getting overridden when using AzureSignalR](https://learn.microsoft.com/en-us/answers/questions/1741737/blazor-server-side-custom-cultureinfo-getting-over).

### To Reproduce
Repo:
https://github.com/jbryanSavers/BlazorCultureCookieSignalRTest

Without using Azure-SignalR, the output:

image

Using Azure-SignalR, the result like this:

image

### Further technical details
- Your Azure SignalR SDK version：  Using Latest Version .

- Your Server ASPNETCORE version or Assembly version of `Microsoft.AspNetCore.SignalR` 
- Your SignalR Client SDK version

xingsy97:
Based on https://github.com/Azure/azure-signalr/pull/2067
Fix #2038


src/Microsoft.Azure.SignalR.AspNet/Middleware/NegotiateMiddleware.cs
```diff
@@ -160,7 +160,7 @@ private Task ProcessNegotiationRequest(IOwinContext owinContext, HostContext con
             // add OriginalPath and QueryString when the clients protocol is higher than 2.0, earlier ASP.NET SignalR clients does not support redirect URL with query parameters
             if (!string.IsNullOrEmpty(clientProtocol) && Version.TryParse(clientProtocol, out var version) && version >= ClientSupportQueryStringVersion)
             {
-                var clientRequestId = _connectionRequestIdProvider.GetRequestId();
+                var clientRequestId = _connectionRequestIdProvider.GetRequestId("");
```
vicancy: so for aspnet the id always start with `-`?

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
```diff
@@ -149,6 +154,12 @@ protected override Task OnClientConnectedAsync(OpenConnectionMessage message)
         connection.ServiceConnection = this;
 
         connection.Features.Set<IConnectionMigrationFeature>(null);
+
+        if (!_cultureInfoManager.TryApplyCulture(connection.RequestId))
```
vicancy:
```cs
if (_cultureInfoManager.TryGetCultureFeature(connection.RequestId, out var feature)){
connection.GetHttpContext().Features.Set<IRequestCultureFeature>(feature)
}

```

src/Microsoft.Azure.SignalR.Common/Auth/AuthUtility.cs
```diff
@@ -65,8 +65,8 @@ public static string GenerateAccessToken(byte[] keyBytes,
         return jwtToken.Length > MaxTokenLength ? throw new AzureSignalRAccessTokenTooLongException() : jwtToken;
     }
 
-    public static string GenerateRequestId()
+    public static string GenerateRequestId(string traceIdentifier)
     {
-        return Convert.ToBase64String(BitConverter.GetBytes(Stopwatch.GetTimestamp()));
+        return $"{traceIdentifier}-{Convert.ToBase64String(BitConverter.GetBytes(Stopwatch.GetTimestamp()))}";
```
vicancy: traceIdentifier is in such format `0HN7T92QVLV5T:00000009` https://github.com/aspnet/KestrelHttpServer/blob/master/src/Kestrel.Core/Internal/Infrastructure/CorrelationIdGenerator.cs#L19, and could be customized using middlewares. 
Do we want `:` in the requestId? might the generated requestId too long?

src/Microsoft.Azure.SignalR.Common/Interfaces/IClientConnectionManager.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.Collections.Generic;
+using System.Globalization;
```
vicancy: no need?

src/Microsoft.Azure.SignalR/HubHost/NegotiateHandler.cs
```diff
@@ -52,7 +53,8 @@ public NegotiateHandler(
 #if NET6_0_OR_GREATER
             EndpointDataSource endpointDataSource,
 #endif
-            ILogger<NegotiateHandler<THub>> logger)
+            ILogger<NegotiateHandler<THub>> logger,
```
vicancy: logger is usually the last parameter

src/Microsoft.Azure.SignalR/HubHost/NegotiateHandler.cs
```diff
@@ -89,12 +91,14 @@ public async Task<NegotiationResponse> Process(HttpContext context)
...
 
+            var clientRequestId = _connectionRequestIdProvider.GetRequestId(context.TraceIdentifier);
             var queryString = GetQueryString(
                 request.QueryString.HasValue ? request.QueryString.Value.Substring(1) : null,
-                cultureName,
-                uiCultureName
+                clientRequestId
             );
 
+            _cultureInfoManager.TryAddCultureFeature(clientRequestId, cultureFeature);
```
vicancy: is it possible to check if culture feature is unchanged? If it is unchanged, there is no need to add and apply

vicancy: is it that only blazor app is impacted by the culture?

src/Microsoft.Azure.SignalR.AspNet/Middleware/NegotiateMiddleware.cs
xingsy97: use a prefix "aspnet"

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnection.cs
xingsy97: `connection.GetHttpContext().Features.Set<IRequestCultureFeature>(feature)` doesn't take effect
I use `CultureInfo.CurrentCulture=XXX` and `CultureInfo.CureentUICulture=XXX`

src/Microsoft.Azure.SignalR.Common/Auth/AuthUtility.cs
xingsy97: now I use `return $"{traceIdentifier}:{Stopwatch.GetTimestamp().ToString("X")}";`
base64 encoding is not needed for this id will be UrlEncoded before inserted into query string
use hex to shorten the length

src/Microsoft.Azure.SignalR/HubHost/NegotiateHandler.cs
xingsy97: 1\ yes, we can check. However if we don't add the feature, a not found error log will be triggered when trying to apply the culture feature.
2\ added blazor check

JialinXin:
Add some UTs?

src/Microsoft.Azure.SignalR.Common/Auth/AuthUtility.cs
```diff
@@ -65,8 +65,10 @@ public static string GenerateAccessToken(byte[] keyBytes,
         return jwtToken.Length > MaxTokenLength ? throw new AzureSignalRAccessTokenTooLongException() : jwtToken;
     }
 
-    public static string GenerateRequestId()
+    public static string GenerateRequestId(string traceIdentifier)
     {
-        return Convert.ToBase64String(BitConverter.GetBytes(Stopwatch.GetTimestamp()));
+        // Before filled into query string, this id will be process by "WebUtility.UrlEncode(...)". So base64 encoding is not needed.
```
vicancy: let's replace `:` with `-` to make it more readable, also use `string.IsNullorEmpty(traceidentifier) ? {Stopwatch.GetTimestamp().ToString("X") : $"{traceIdentifier}:{Stopwatch.GetTimestamp().ToString("X")}"` instead of append `aspnet` prefix

src/Microsoft.Azure.SignalR.Common/DefaultCultureFeatureManager.cs
```diff
@@ -37,6 +37,12 @@ public bool TryRemoveCultureFeature(string requestId, out IRequestCultureFeature
         return false;
     }
 
+    public bool IsDefaultFeature(IRequestCultureFeature feature)
+    {
+        // this is the default feature value in blazor when no culture feature is configured by app server
+        return feature.RequestCulture.Culture == CultureInfo.DefaultThreadCurrentCulture && feature.RequestCulture.UICulture == CultureInfo.DefaultThreadCurrentUICulture && feature.Provider is AcceptLanguageHeaderRequestCultureProvider;
```
vicancy: is AcceptLanguageHeaderRequestCultureProvider does not mean it is default though

src/Microsoft.Azure.SignalR.Common/Auth/AuthUtility.cs
```diff
@@ -65,8 +65,12 @@ public static string GenerateAccessToken(byte[] keyBytes,
...
+        // Use hex to shorten the length.
+        return string.IsNullOrEmpty(traceIdentifier) 
+            ? Stopwatch.GetTimestamp().ToString("X")
+            : $"{traceIdentifier}-{Stopwatch.GetTimestamp().ToString("X")}";
```
vicancy: traceIdentifier could also contain `:`

src/Microsoft.Azure.SignalR.Common/Auth/AuthUtility.cs
```diff
@@ -65,8 +65,12 @@ public static string GenerateAccessToken(byte[] keyBytes,
...
     }
 
-    public static string GenerateRequestId()
+    public static string GenerateRequestId(string traceIdentifier)
     {
-        return Convert.ToBase64String(BitConverter.GetBytes(Stopwatch.GetTimestamp()));
+        // Before filled into query string, this id will be process by "WebUtility.UrlEncode(...)". So base64 encoding is not needed.
+        // Use hex to shorten the length.
+        return string.IsNullOrEmpty(traceIdentifier) 
+            ? Stopwatch.GetTimestamp().ToString("X")
```
vicancy: following the practice of traceIdentifier, we could also use some interlock.increment index appended to the timestamp to make it unique

vicancy (approved):

PR #2071, opened 2024-11-04
Throw different 401 error message

terencefan:


src/Microsoft.Azure.SignalR.Common/ServiceConnections/Internal/WebSocketsTransport.cs
```diff
@@ -12,403 +12,409 @@
 using Microsoft.Azure.SignalR.Common;
 using Microsoft.Extensions.Logging;
 
-namespace Microsoft.Azure.SignalR.Connections.Client.Internal
+namespace Microsoft.Azure.SignalR.Connections.Client.Internal;
```
vicancy: any logic change in this file? please separate the format change into another PR

vicancy (approved):

PR #2085, opened 2024-11-13
Update MicrosoftEntraAccessKey Available logic

terencefan:


src/Microsoft.Azure.SignalR.Common/Exceptions/AzureSignalRAccessTokenNotAuthorizedException.cs
```diff
@@ -36,17 +38,17 @@ public AzureSignalRAccessTokenNotAuthorizedException(string credentialName, Exce
...
     }
 
-    private static string GetInnerReason(Exception exception)
+    private static string GetInnerReason(Exception? exception)
     {
         return exception switch
         {
             AzureSignalRUnauthorizedException => AzureSignalRUnauthorizedException.ErrorMessageMicrosoftEntra,
-            _ => exception.Message,
+            _ => exception?.Message ?? "The access key has expired.",
```
vicancy: this might be correct for now, but making this a default assumption in this exception class sounds not an extensible solution

vicancy (approved):

PR #2089, opened 2024-11-20
Update messagepack to tag v2.5.192

vicancy:


vwxyzh (approved):

src/Microsoft.Azure.SignalR.Protocols/ServiceProtocol.cs
```diff
@@ -1440,27 +1441,36 @@ private static string ReadString(ref MessagePackReader reader, string field)
...
+    private static string? ReadString(ref MessagePackReader reader, string formatField, int param)
     {
         try
         {
             return reader.ReadString();
         }
         catch (Exception ex)
         {
-            throw new InvalidDataException($"Reading '{string.Format(formatField, param1, param2)}' as String failed.", ex);
+            throw new InvalidDataException($"Reading '{string.Format(formatField, param)}' as String failed.", ex);
```
JialinXin, line 1473: Is it nullable? throw here looks it doesn't accept null result.

JialinXin (approved):

PR #2096, opened 2024-11-26
Lower EndpointOffline to warning

### LINKED ISSUE #2093: Control log level of _endpointOffline log line
### Is your feature request related to a problem? Please describe.
When making rolling deployments (i.e. in pod rotations in Kubernetes), we are seeing the following error in our logs:

`Hub '{hub}' is now disconnected from '{endpoint}'. Please check log for detailed info.`

We're treating logs of Error level as scenarios, where something is _really_ wrong and something that developers needs to pay attention to immediately - and in this case, we don't have any control over the level of this log line because it's set up within [`ServiceConnectionContainerBase.cs`](https://github.com/Azure/azure-signalr/blob/30d6a43a1b72349922948a7c494ee7d8dc068e32/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs#L761) 

Ideally, we'd like to have it only as a warning since we know that it does happen when making a rolling deployment.

### Describe the solution you'd like
Either:
- Simply lowering the _endpointOffline to Warning or
- Use a named logger so the consumers are able to control/override the log level using Serilog

### Additional context
We've configured Azure SignalR to gracefully shutdown and migrate client connections, like so:

```
builder.Services.AddSignalR()
        .AddAzureSignalR(option =>
        {            
            option.GracefulShutdown.Mode = GracefulShutdownMode.MigrateClients;
            option.GracefulShutdown.Timeout = TimeSpan.FromSeconds(30);
        })
```

We're using Serilog to handle the logging for us.

vicancy:
Fix https://github.com/Azure/azure-signalr/issues/2093

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
+++ b/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
@@ -758,7 +758,7 @@ internal abstract class ServiceConnectionContainerBase : IServiceConnectionConta
             LoggerMessage.Define<string, string>(LogLevel.Information, new EventId(1, "EndpointOnline"), "Hub '{hub}' is now connected to '{endpoint}'.");
 
         private static readonly Action<ILogger, string, string, Exception> _endpointOffline =
-            LoggerMessage.Define<string, string>(LogLevel.Error, new EventId(2, "EndpointOffline"), "Hub '{hub}' is now disconnected from '{endpoint}'. Please check log for detailed info.");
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.SignalR.Common/ServiceConnections/Internal/WebSocketsTransport.cs
```diff
@@ -88,6 +87,8 @@ public WebSocketsTransport(WebSocketConnectionOptions connectionOptions,
                 _webSocket.Options.UseDefaultCredentials = connectionOptions.UseDefaultCredentials.Value;
             }
 
+
```
vwxyzh: revert the changes for blank lines

vwxyzh (approved):

zackliu (approved):

PR #2105, opened 2024-12-03
Remove authorization for healthcheck API

cqnguy23:
- Remove Authorization for healthcheck API


src/Microsoft.Azure.SignalR.Emulator/Common/SignalRServiceWebApiDefinition.cs
```diff
@@ -11,7 +11,6 @@
 
 namespace Microsoft.Azure.SignalR.Controllers.Common
 {
-    [Authorize]
     [Route("api/v1")]
```
vicancy: use using [AllowAnonymous] for health api enough?

vicancy (approved):

PR #2107, opened 2024-12-04
remove inactive keys from sychronizer

terencefan:


src/Microsoft.Azure.SignalR.Common/Auth/MicrosoftEntra/AccessKeySynchronizer.cs
```diff
@@ -8,35 +8,26 @@
...
 namespace Microsoft.Azure.SignalR;
 
 internal sealed class AccessKeySynchronizer : IAccessKeySynchronizer, IDisposable
 {
     private readonly ConcurrentDictionary<MicrosoftEntraAccessKey, bool> _keyMap = new(ReferenceEqualityComparer.Instance);
 
-    private readonly ILogger<AccessKeySynchronizer> _logger;
-
-    private readonly TimerAwaitable _timer = new TimerAwaitable(TimeSpan.Zero, TimeSpan.FromMinutes(1));
+    private readonly TimerAwaitable _timer = new TimerAwaitable(TimeSpan.FromMinutes(1), TimeSpan.FromMinutes(1));
```
terencefan, line 18: There is no need to start the timer task immediately as we load the access key on demand.

PR #2110, opened 2024-12-05
remove Endpoint in IAccessKey

terencefan:


src/Microsoft.Azure.SignalR.Common/Auth/MicrosoftEntra/MicrosoftEntraAccessKey.cs
```diff
@@ -78,22 +78,17 @@ private set
...
-    public Uri Endpoint { get; }
-
     internal Exception? LastException { get; private set; }
 
     internal string GetAccessKeyUrl { get; }
 
     internal TimeSpan GetAccessKeyRetryInterval { get; set; } = TimeSpan.FromSeconds(3);
 
-    public MicrosoftEntraAccessKey(Uri endpoint,
+    public MicrosoftEntraAccessKey(Uri serverEndpoint,
```
vicancy, line 87: as the naming convention we used in ServiceEndpoint, this should be the serviceEndpoint?

vicancy (approved):

PR #2114, opened 2024-12-06
Refresh AccessKey passively

terencefan:


src/Microsoft.Azure.SignalR.Common/Auth/MicrosoftEntra/MicrosoftEntraAccessKey.cs
```diff
@@ -119,55 +117,61 @@ public async Task<string> GenerateAccessTokenAsync(string audience,
                                                        AccessTokenAlgorithm algorithm,
                                                        CancellationToken ctoken = default)
     {
-        if (!_initializedTcs.Task.IsCompleted)
+        var updateTask = Task.CompletedTask;
+        if (NeedRefresh)
```
vicancy, line 121: NeedRefresh is false before Available is set, so it is possible that multiple `cts` are created, move the cts creation into the lock?

src/Microsoft.Azure.SignalR.Common/Auth/MicrosoftEntra/MicrosoftEntraAccessKey.cs
```diff
@@ -119,55 +117,61 @@ public async Task<string> GenerateAccessTokenAsync(string audience,
...
+        lock (_lock)
         {
-            return;
+            if (!_updateTaskSource.Task.IsCompleted)
+            {
+                source = _updateTaskSource;
+            }
+            else
+            {
+                _updateTaskSource = new TaskCompletionSource<bool>();
```
vicancy: `TaskCreationOptions.RunContinuationsAsynchronously`

src/Microsoft.Azure.SignalR.Common/Auth/MicrosoftEntra/MicrosoftEntraAccessKey.cs
```diff
@@ -119,55 +117,61 @@ public async Task<string> GenerateAccessTokenAsync(string audience,
...
+            _updateTaskSource.TrySetResult(true);
         }
+    }
 
-        if (Interlocked.CompareExchange(ref _updateState, UpdateTaskRunning, UpdateTaskIdle) != UpdateTaskIdle)
+    internal async Task<bool> UpdateAccessKeyAsync(CancellationToken ctoken = default)
+    {
+        TaskCompletionSource<bool> source;
+        lock (_lock)
         {
```
vicancy, line 157: double check NeedRefresh inside the lock

src/Microsoft.Azure.SignalR.Common/Auth/MicrosoftEntra/MicrosoftEntraAccessKey.cs
```diff
@@ -119,55 +117,61 @@ public async Task<string> GenerateAccessTokenAsync(string audience,
...
-            return;
-        }
-        else if (!Available && delta < GetAccessKeyIntervalWhenUnauthorized)
+        lock (_lock)
         {
-            return;
+            _keyBytes = Encoding.UTF8.GetBytes(keyStr);
+            _kid = kid;
+            Available = true;
+            _updateTaskSource.TrySetResult(true);
```
vicancy, line 149: only _updateTaskSource needs to be in the lock?

src/Microsoft.Azure.SignalR.Common/Auth/MicrosoftEntra/MicrosoftEntraAccessKey.cs
terencefan: NeedRefresh is always true, the init value of `_updatedAt` is `DateTime.MinValue`

vicancy (approved):

PR #2118, opened 2024-12-30
Add group member query service message protocol

Y-Sindo:
To implement group member query API in persistent mode, we need to add a mechanism to respond with binary payload from service.
* Move some util method in `ServiceProtocol` to a separate util class `MessagePackUtils`, so that they could be reused by other classes.
* Deprecate `string message` member in `AckMessage` and add `Payload` member. The `string message` member is never used. 
* Introduce `IMessagePackSerializable` interface, so that we could put the (de)serialization methods of model classes inside themselves.
* Refactor `AckHandler` to allow acking with binary payload


src/Microsoft.Azure.SignalR.Protocols/ServiceProtocol.cs
```diff
@@ -661,7 +661,16 @@ private static void WriteAckMessage(ref MessagePackWriter writer, AckMessage mes
         writer.Write(ServiceProtocolConstants.AckMessageType);
         writer.Write(message.AckId);
         writer.Write(message.Status);
-        writer.Write(message.Message);
+        if (message.Payload.HasValue)
```
vicancy, line 688: change to WriteArrayHeader(6) for this added property message.Payload instead?

src/Microsoft.Azure.SignalR.Protocols/ServiceProtocol.cs
```diff
@@ -1244,9 +1263,23 @@ private static AckMessage CreateAckMessage(ref MessagePackReader reader, int arr
     {
         var ackId = ReadInt32(ref reader, "ackId");
         var status = ReadInt32(ref reader, "status");
-        var message = ReadString(ref reader, "message");
+        ReadOnlySequence<byte>? payload = null;
+        string? message = null;
+        if (reader.NextMessagePackType == MessagePackType.Binary)
```
vicancy, line 1282: I would prefer using different array length instead

src/Microsoft.Azure.SignalR.Protocols/ServiceProtocolConstants.cs
```diff
@@ -45,4 +45,6 @@ public static class ServiceProtocolConstants
     public const int ServiceMappingMessageType = 37;
     public const int ConnectionReconnectMessageType = 38;
     public const int ConnectionFlowControlMessageType = 39;
+    public const int GroupMemberQueryMessageType = 40;
+    public const int GroupMemberQueryResponseMessageType = 41;
```
vicancy: do we still need this if we reuse ack message?

src/Microsoft.Azure.SignalR.Protocols/Models/GroupMemberQueryResponsePayload.cs
```diff
@@ -0,0 +1,57 @@
...
+        var members = new List<GroupMember>(memberCount);
+        for (var i = 0; i < memberCount; i++)
+        {
+            members.Add(reader.Deserialize<GroupMember>("groupMembers"));
+        }
+        result.Members = members;
+        result.ContinuationToken = reader.ReadString();
+    }
+
+    //    public override bool Equals(object obj)
```
vicancy: ?

src/Microsoft.Azure.SignalR.Management/HubContext/GroupManager.cs
```diff
@@ -14,5 +17,7 @@ public abstract class GroupManager : IGroupManager
         public abstract Task RemoveFromGroupAsync(string connectionId, string groupName, CancellationToken cancellationToken = default);
 
         public abstract Task RemoveFromAllGroupsAsync(string connectionId, CancellationToken cancellationToken = default);
+
+        public virtual IAsyncEnumerable<GroupMember> ListConnectionsInGroup(string groupName, int? max, CancellationToken token) => throw new NotImplementedException();
```
vicancy: Async suffix

src/Microsoft.Azure.SignalR.Common/Utilities/AckHandler.cs
```diff
@@ -62,29 +79,19 @@ public Task<AckStatus> CreateMultiAck(out int id, TimeSpan? ackTimeout = default
...
-                    case IAckInfo<AckStatus> ackInfo:
-                        if (ackInfo.Ack(status))
-                        {
-                            _acks.TryRemove(id, out _);
-                        }
-                        break;
-                    default:
-                        throw new InvalidCastException($"Expected: IAckInfo<{typeof(IAckInfo<AckStatus>).Name}>, actual type: {info.GetType().Name}");
-                }
+                _acks.TryRemove(id, out _);
```
vicancy, line 95: changes the original logic of type check?

src/Microsoft.Azure.SignalR.Protocols/Models/GroupMember.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using MessagePack;
+using static Microsoft.Azure.SignalR.Protocol.MessagePackUtils;
+
+namespace Microsoft.Azure.SignalR.Protocol;
```
vicancy: nit: set as the default namespace and no need to specify

src/Microsoft.Azure.SignalR.Protocols/Models/GroupMemberQueryResponsePayload.cs
```diff
@@ -0,0 +1,57 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
```
vicancy: I would suggest separating the protocol change into another PR

src/Microsoft.Azure.SignalR.Common/Utilities/AckHandler.cs
Y-Sindo: Yes, as we've move the `Ack(...)` method up from `IAckInfo<T>` to `IAckInfo`, there is no need for type check.

src/Microsoft.Azure.SignalR.Protocols/ServiceProtocol.cs
Y-Sindo: We cannot extend the array length, as in old protocol, we expect that the extension members start from the fifth...

src/Microsoft.Azure.SignalR.Protocols/Models/GroupMemberQueryResponsePayload.cs
Y-Sindo: Removed

src/Microsoft.Azure.SignalR.Common/Utilities/AckHandler.cs
```diff
@@ -170,39 +174,61 @@ private interface IAckInfo
...
                 TimeoutAt = DateTime.UtcNow + timeout;
             }
+            public abstract bool Ack(AckStatus status, ReadOnlySequence<byte>? payload = null);
+            public Task<T> Task => _tcs.Task;
+            public void Cancel() => _tcs.TrySetCanceled();
+        }
 
-            public bool Ack(AckStatus status = AckStatus.Ok) =>
-                _tcs.TrySetResult(status);
+        private class SingleAckWithStatusInfo : SingleAckInfo<AckStatus>
```
vicancy: nit: naming: SingleStatusAck?

src/Microsoft.Azure.SignalR.Common/Utilities/AckHandler.cs
```diff
@@ -170,39 +174,61 @@ private interface IAckInfo
...
 
-            public void Cancel() => _tcs.TrySetCanceled();
+            public override bool Ack(AckStatus status, ReadOnlySequence<byte>? payload = null) =>
+                _tcs.TrySetResult(status);
+        }
+
+        private sealed class SingleAckWithMessagePackPayloadInfo<T> : SingleAckInfo<IMessagePackSerializable> where T : IMessagePackSerializable, new()
+        {
+            public SingleAckWithMessagePackPayloadInfo(TimeSpan timeout) : base(timeout) { }
+            public override bool Ack(AckStatus status, ReadOnlySequence<byte>? payload = null)
```
vicancy, line 216: Maybe we could pass in the Payload deserializer in .ctor?

src/Microsoft.Azure.SignalR.Common/Utilities/AckHandler.cs
```diff
@@ -170,39 +174,61 @@ private interface IAckInfo
...
+                if (status == AckStatus.Timeout)
+                {
+                    return _tcs.TrySetException(new TimeoutException($"Waiting for a {typeof(T).Name} response timed out."));
+                }
+                if (payload == null)
+                {
+                    throw new ArgumentNullException(nameof(payload));
+                }
+                var reader = new MessagePackReader(payload.Value);
+                var result = reader.Deserialize<T>(string.Empty);
```
vicancy: what if this throws?

vicancy: why string.Empty?

src/Microsoft.Azure.SignalR.Common/Utilities/AckHandler.cs
```diff
@@ -35,15 +40,27 @@ public Task<AckStatus> CreateSingleAck(out int id, TimeSpan? ackTimeout = defaul
...
 
+        public Task<T> CreateSingleAck<T>(out int id, TimeSpan? ackTimeout = default, CancellationToken cancellationToken = default) where T : IMessagePackSerializable, new()
+        {
+            id = NextId();
+            if (_disposed)
+            {
+                return Task.FromResult(new T());
+            }
+            var info = (IAckInfo<IMessagePackSerializable>)_acks.GetOrAdd(id, _ => new SingleAckWithMessagePackPayloadInfo<T>(ackTimeout ?? _defaultAckTimeout));
+            cancellationToken.Register(info.Cancel);
```
vicancy, line 61: also validate `if (info is MultiAckWithStatusInfo)`?

Y-Sindo: The `id` is a newly generated number, so the return `info` must be the one we added in the method.

src/Microsoft.Azure.SignalR.Common/Utilities/AckHandler.cs
Y-Sindo: We could do this once we have other types of payload.

src/Microsoft.Azure.SignalR.Common/Utilities/AckHandler.cs
Y-Sindo: Updated

src/Microsoft.Azure.SignalR.Common/Utilities/AckHandler.cs
Y-Sindo: Updated

src/Microsoft.Azure.SignalR.Common/Utilities/AckHandler.cs
Y-Sindo: Now add a `ServiceModelProtocol` to hide the MessagePack concept.

src/Microsoft.Azure.SignalR.Protocols/Models/ServiceModelProtocol.cs
```diff
@@ -0,0 +1,58 @@
...
+
+using System;
+using System.Buffers;
+using MessagePack;
+
+namespace Microsoft.Azure.SignalR.Protocol;
+
+#nullable enable
+
+public class ServiceModelProtocol
```
vicancy: is internal enough?

Y-Sindo: Runtime needs this class to serialize models.

vicancy (approved):

PR #2119, opened 2025-01-02
Client invocation for management sdk

TD2106:
 - Add client invocation manager for management sdk.


TD2106:
@microsoft-github-policy-service agree company="Microsoft"

src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
```diff
@@ -214,5 +224,58 @@ protected override T AppendMessageTracingId<T>(T message)
...
+                throw new ArgumentException("Invalid result or invocationId", nameof(result));
+            }
+
+            // `TryCompletionResult` returns false when the corresponding invocation is not existing.
+            if (_clientInvocationManager.Caller.TryCompleteResult(connectionId, result))
+            {
+                // For caller server, the only purpose of sending ClientCompletionMessage is to inform service to cleanup the invocation, which means only InvocationId and ConnectionId are needed.
+                // To avoid serialization for useless payload, we keep payload as empty bytes.
+                var message = AppendMessageTracingId(new ClientCompletionMessage(result.InvocationId, connectionId, _callerId, "", new ReadOnlyMemory<byte>()));
+                await WriteAsync(message);
```
JialinXin: Is this still needed? In serverless case, service need to know this is a paired CompletionMessage then send back, meanwhile it should be able to cleanup local directly without information from the caller.

src/Microsoft.Azure.SignalR.Management/ClientInvocation/WeakCallerClientResultsManager.cs
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+        public Task<T> AddInvocation<T>(string hub, string connectionId, string invocationId, CancellationToken cancellationToken)
+        {
+            var tcs = new TaskCompletionSourceWithCancellation<T>(
+                cancellationToken,
+                () => TryCompleteResult(connectionId, CompletionMessage.WithError(invocationId, "Canceled")));
+
+            // When the caller server is also the client router, Azure SignalR service won't send a ServiceMappingMessage to server.
+            // To handle this condition, CallerClientResultsManager itself should record this mapping information rather than waiting for a ServiceMappingMessage sent by service. Only in this condition, this method is called with instanceId != null.
```
JialinXin: modify the comment to fit serverless scenario.

src/Microsoft.Azure.SignalR.Management/ClientInvocation/WeakCallerClientResultsManager.cs
```diff
@@ -0,0 +1,182 @@
...
+                return false;
+            }
+            else
+            {
+                // connection was disconnected or someone else completed the invocation
+                return false;
+            }
+        }
+
+        public bool TryCompleteResult(string connectionId, ClientCompletionMessage message)
```
JialinXin: I think Service can send back original `CompletionMessage` directly. Below methods are not really fit serverless and should throw `NotSupportedException`.

`TryCompleteResult(ConnectionId, ClientCompletionMessage)`
`TryCompleteResult(ConnectionId, ErrorCompletionMessage)`
`AddServiceMapping()`

src/Microsoft.Azure.SignalR.Management/ClientInvocation/WeakClientInvocationManager.cs
```diff
@@ -0,0 +1,35 @@
...
+using Microsoft.AspNetCore.SignalR;
+
+#nullable enable
+
+namespace Microsoft.Azure.SignalR.Management.ClientInvocation
+{
+    internal sealed class WeakClientInvocationManager : IClientInvocationManager
+    {
+        public ICallerClientResultsManager Caller { get; }
+        public IRoutedClientResultsManager Router { get; }
```
JialinXin: Throw directly without assign a dummy router manager?

src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
TD2106: Removed.

src/Microsoft.Azure.SignalR.Management/ClientInvocation/WeakCallerClientResultsManager.cs
TD2106: Removed this comment.

src/Microsoft.Azure.SignalR.Management/ClientInvocation/WeakCallerClientResultsManager.cs
TD2106: Removed these implementation

src/Microsoft.Azure.SignalR.Management/ClientInvocation/WeakClientInvocationManager.cs
TD2106: Removed the dummy router manager.

src/Microsoft.Azure.SignalR.Management/ClientInvocation/WeakCallerClientResultsManager.cs
```diff
@@ -0,0 +1,171 @@
...
+#if NET7_0_OR_GREATER
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Collections.Generic;
+using System.Collections.Concurrent;
+using Microsoft.Azure.SignalR.Protocol;
+using Microsoft.AspNetCore.SignalR.Protocol;
+using Microsoft.AspNetCore.SignalR;
+using System.Linq;
```
vwxyzh: order

src/Microsoft.Azure.SignalR.Management/ClientInvocation/WeakClientInvocationManager.cs
```diff
@@ -0,0 +1,37 @@
...
+using Microsoft.AspNetCore.SignalR;
+
+#nullable enable
+
+namespace Microsoft.Azure.SignalR.Management
+{
+    internal sealed class WeakClientInvocationManager : IClientInvocationManager
+    {
+        public ICallerClientResultsManager Caller { get; }
+        public IRoutedClientResultsManager? Router { get; }
```
vwxyzh, line 14: is it always `null`?

src/Microsoft.Azure.SignalR.Management/DependencyInjectionExtensions.cs
```diff
@@ -119,6 +120,14 @@ private static IServiceCollection AddSignalRServiceCore(this IServiceCollection
 
             services.AddRestClient();
             services.AddSingleton<NegotiateProcessor>();
+            services.AddSingleton<IServerNameProvider>(sp => new DefaultServerNameProvider());
```
vicancy: services.AddSingleton<IServerNameProvider, DefaultServerNameProvider>();

src/Microsoft.Azure.SignalR.Management/ClientInvocation/WeakClientInvocationManager.cs
TD2106: Yes it should be null in serverless case because there is no sticky stickiness, the routing is done 100% on the service end.

src/Microsoft.Azure.SignalR.Management/ClientInvocation/WeakCallerClientResultsManager.cs
TD2106: Good catch, updated.

src/Microsoft.Azure.SignalR.Management/ClientInvocation/WeakCallerClientResultsManager.cs
```diff
@@ -0,0 +1,196 @@
...
+    internal sealed class WeakCallerClientResultsManager : ICallerClientResultsManager, IInvocationBinder
+    {
+        private readonly ConcurrentDictionary<string, PendingInvocation> _pendingInvocations = new();
+        private readonly string _clientResultManagerId = Guid.NewGuid().ToString("N");
+        private long _lastInvocationId = 0;
+        private readonly AckHandler _ackHandler;
+        private readonly IServiceEndpointManager _endpointManager;
+        private readonly IEndpointRouter _endpointRouter;
+        private readonly IHubProtocolResolver _hubProtocolResolver;
+
```
vicancy: nit: remove extra line

src/Microsoft.Azure.SignalR.Management/ClientInvocation/WeakCallerClientResultsManager.cs
```diff
@@ -0,0 +1,196 @@
...
+            var tcs = new TaskCompletionSourceWithCancellation<T>(
+                cancellationToken,
+                () => TryCompleteResult(connectionId, CompletionMessage.WithError(invocationId, "Canceled")));
+
+            var serviceEndpoints = _endpointManager.GetEndpoints(hub);
+            var expectedAckCount = _endpointRouter.GetEndpointsForConnection(connectionId, serviceEndpoints).Count();
+
+            var multiAck = _ackHandler.CreateMultiAck(out var ackId);
+
+            _ackHandler.SetExpectedCount(ackId, expectedAckCount);
```
vicancy: this is not correct behavior, since `_endpointRouter.GetEndpointsForConnection` value might change from time to time. We need to fix this logic first.  @JialinXin

src/Microsoft.Azure.SignalR.Management/ClientInvocation/WeakCallerClientResultsManager.cs
```diff
@@ -0,0 +1,196 @@
...
+
+        public void CleanupInvocationsByInstance(string instanceId) { }
+
+        public void CleanupInvocationsByConnection(string connectionId)
+        {
+            foreach (var (invocationId, invocation) in _pendingInvocations)
+            {
+                if (invocation.ConnectionId == connectionId)
+                {
+                    var message = CompletionMessage.WithError(invocationId, $"Connection '{invocation.ConnectionId}' is disconnected.");
```
vicancy: is this method only triggered when the connection is disconnected?

src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
```diff
@@ -214,5 +224,64 @@ protected override T AppendMessageTracingId<T>(T message)
 
             return base.AppendMessageTracingId(message);
         }
+
+#if NET7_0_OR_GREATER
+        public override async Task<T> InvokeConnectionAsync<T>(string connectionId, string methodName, object?[] args, CancellationToken cancellationToken = default)
+        {
+            if (cancellationToken == default)
+            {
+                using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(3));
```
vicancy: why 3 seconds?

src/Microsoft.Azure.SignalR.Management/ClientInvocation/WeakCallerClientResultsManager.cs
TD2106: Yes correct, but since this is serverless mode it will probably not get triggered. Maybe I can just remove this implementation

src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
TD2106: This is due to it being in serverless mode. So if the pod connecting the client crashes, it won't get notify to the remove this pending invocation. So I use 3 seconds here to make sure that it will get removed, to avoid memory leaks.

src/Microsoft.Azure.SignalR.Management/DependencyInjectionExtensions.cs
TD2106: Updated

src/Microsoft.Azure.SignalR.Management/ClientInvocation/WeakCallerClientResultsManager.cs
TD2106: Updated

src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
```diff
@@ -214,5 +227,65 @@ protected override T AppendMessageTracingId<T>(T message)
 
             return base.AppendMessageTracingId(message);
         }
+
+#if NET7_0_OR_GREATER
+        public override async Task<T> InvokeConnectionAsync<T>(string connectionId, string methodName, object?[] args, CancellationToken cancellationToken = default)
```
vicancy: add test coverage

src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
```diff
@@ -13,13 +16,25 @@ namespace Microsoft.Azure.SignalR.Management
 {
     internal class WebSocketsHubLifetimeManager<THub> : ServiceLifetimeManagerBase<THub>, IServiceHubLifetimeManager<THub> where THub : Hub
     {
+#if NET7_0_OR_GREATER
+        private const int DefaultInvocationTimeoutSeconds = 3;
```
vicancy: why it's 3 seconds? what is the proxy mode behavior?

src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
```diff
@@ -214,5 +227,65 @@ protected override T AppendMessageTracingId<T>(T message)
...
             return base.AppendMessageTracingId(message);
         }
+
+#if NET7_0_OR_GREATER
+        public override async Task<T> InvokeConnectionAsync<T>(string connectionId, string methodName, object?[] args, CancellationToken cancellationToken = default)
+        {
+            // cancellationToken is required to be cancellable.
+            if (!cancellationToken.CanBeCanceled)
+            {
+                using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(DefaultInvocationTimeoutSeconds));
```
vicancy: could use CreateLinkedTokenSource if default timeout is a must have

src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
```diff
@@ -214,5 +227,65 @@ protected override T AppendMessageTracingId<T>(T message)
...
+                cancellationToken = cts.Token;
+            }
+
+            if (IsInvalidArgument(connectionId))
```
vicancy: move validation checks up as the first steps

src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
```diff
@@ -214,5 +227,65 @@ protected override T AppendMessageTracingId<T>(T message)
...
+
+            if (result == null || string.IsNullOrEmpty(result.InvocationId))
+            {
+                throw new ArgumentException("Invalid result or invocationId", nameof(result));
+            }
+
+            // `TryCompletionResult` returns false when the corresponding invocation is not existing.
+            if (!_clientInvocationManager.Caller.TryCompleteResult(connectionId, result))
+            {
+                Logger.LogInformation("Failed to set result for connection {connectionId} with invocationId {invocationId}", connectionId, result.InvocationId);
```
vicancy: what are we expecting the customer to do with this information log?

src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
TD2106: Proxy mode behavior doesn't have timeout limit. However as discussed before, in serverless mode the app server can not get notified when a connection drop, or a pod crash, so I think it's necessary to have some kind of timeout in place. The actual timeout can be adjusted. Maybe 5 seconds is good.

vicancy: maybe 100 seconds, as the same as the default httpclient timeout

src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
TD2106: Added test for this class

src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
TD2106: Updated to 100 seconds

src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
TD2106: Updated

src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
TD2106: Moved as first steps

src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
TD2106: I removed this implementation as it won't be used in serverless mode.

src/Microsoft.Azure.SignalR.Management/ClientInvocation/WeakClientInvocationManager.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        public WeakClientInvocationManager(IServiceEndpointManager serviceEndpointManager, IEndpointRouter endpointRouter, IHubProtocolResolver hubProtocolResolver)
```
vicancy, line 16: directly DI CallerClientResultsManager `WeakClientInvocationManager(CallerClientResultsManager caller)`?

src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
```diff
@@ -215,5 +228,52 @@ protected override T AppendMessageTracingId<T>(T message)
...
+
+            if (IsInvalidArgument(methodName))
+            {
+                throw new ArgumentNullException(nameof(methodName));
+            }
+
+            // cancellationToken is required to be cancellable.
+            var cancellationTokenInUse = cancellationToken;
+
+            if (!cancellationToken.CanBeCanceled)
```
vicancy: why not always merge? if the passed in cancellation token can be canceled, the timeout no longer take effect? Is that intentional?

src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
```diff
@@ -14,13 +17,25 @@ namespace Microsoft.Azure.SignalR.Management
 {
     internal class WebSocketsHubLifetimeManager<THub> : ServiceLifetimeManagerBase<THub>, IServiceHubLifetimeManager<THub> where THub : Hub
     {
+#if NET7_0_OR_GREATER
+        private const int DefaultInvocationTimeoutSeconds = 100;
```
vicancy: we could store the TimeSpan here: readonly static TimeSpan...

src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
```diff
@@ -215,5 +228,52 @@ protected override T AppendMessageTracingId<T>(T message)
...
+
+            if (!cancellationToken.CanBeCanceled)
+            {
+                using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(DefaultInvocationTimeoutSeconds));
+                cancellationTokenInUse = CancellationTokenSource.CreateLinkedTokenSource(cts.Token, cancellationToken).Token;
+            }
+
+            var invocationId = _clientInvocationManager.Caller.GenerateInvocationId(connectionId);
+            var message = AppendMessageTracingId(new ClientInvocationMessage(invocationId, connectionId, _callerId, SerializeAllProtocols(methodName, args, invocationId)));
+            await WriteAsync(message);
```
vicancy: question: what if this throws? is the exception caught from outside and return invocation error? the the error user friendly?

src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
```diff
@@ -215,5 +228,52 @@ protected override T AppendMessageTracingId<T>(T message)
...
+                _clientInvocationManager.Caller.RemoveInvocation(invocationId);
+                throw;
+            }
+        }
+
+        public override Task SetConnectionResultAsync(string connectionId, CompletionMessage result)
+        {
+            // This method won't get trigger because we will not be sending CompletionMessage back from serverless mode.
+            // this is to honor the interface
+            return Task.CompletedTask;
```
vicancy: throw new NotSupportedException()?

src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
```diff
@@ -14,13 +17,25 @@ namespace Microsoft.Azure.SignalR.Management
 {
     internal class WebSocketsHubLifetimeManager<THub> : ServiceLifetimeManagerBase<THub>, IServiceHubLifetimeManager<THub> where THub : Hub
     {
+#if NET7_0_OR_GREATER
+        private const int DefaultInvocationTimeoutSeconds = 100;
+#endif
         private readonly IOptions<ServiceManagerOptions> _serviceManagerOptions;
+        private readonly IClientInvocationManager _clientInvocationManager;
+        private readonly IServerNameProvider _serverNameProvider;
```
vicancy: no need to be a field?

vicancy (commented):
Overall it looks cool to me, some minor comments

src/Microsoft.Azure.SignalR.Management/ClientInvocation/WeakClientInvocationManager.cs
TD2106: This is to follow the current implementation of default mode. In default mode it also creates CallerClientResultsManager.

src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
TD2106: Updated to always merge.

src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
TD2106: Updated to storing Timespan here

src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
TD2106: Updated.

src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
TD2106: Removed.

src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
TD2106: https://github.com/dotnet/aspnetcore/blob/e516ef066f5c66fd385bd8bcf332fe3ab8eb73eb/src/SignalR/server/Core/src/Internal/Proxies.cs#L167
https://github.com/dotnet/aspnetcore/blob/main/src/SignalR/clients/csharp/Client.Core/src/HubConnectionExtensions.InvokeAsync.cs#L27

This will be triggered via theses classes in the main SignalR library. If it throw, the exception will bubble up directly to where the user is calling it. User should be handling it. This is the sample from default mode.
```
public async Task<string> GetMessage(string ID)
        {
            try
            {
                var res = await Clients.Client(ID).InvokeAsync<string>("GetMessage", default);
                return $"From {ID}: {res}";
            }
            catch (Exception ex)
            {
                return $"[Error] Failed invoke connection {ID}: {ex.Message}";
            }
        }
```
As for the user friendly messages, the error completion message will have user friendly messages.

vicancy (approved):

PR #2121, opened 2025-01-08
Add service message  types for presence API

Y-Sindo:


src/Microsoft.Azure.SignalR.Protocols/Models/GroupMember.cs
```diff
@@ -0,0 +1,31 @@
...
+
+using MessagePack;
+using static Microsoft.Azure.SignalR.Protocol.MessagePackUtils;
+
+namespace Microsoft.Azure.SignalR.Protocol;
+
+/// <summary>
+/// Represents a connection in a group.
+/// </summary>
+public record GroupMember : IMessagePackSerializable
```
vicancy, line 12: maybe ConnectionInfo? this class could be used elsewhere, not necessarily as group member

vicancy (approved):
LGTM, one little comment about the naming

src/Microsoft.Azure.SignalR.Protocols/Models/GroupMember.cs
Y-Sindo: Later we might want to add a new field `JoinedTime` to the class, and then `ConnectionInfo` is not appropriate.

vicancy (approved):

PR #2123, opened 2025-01-13
Update service protocol

Y-Sindo:


src/Microsoft.Azure.SignalR.Protocols/ServiceProtocol.cs
```diff
@@ -681,21 +681,16 @@ private static void WriteCheckUserExistenceWithAckMessage(ref MessagePackWriter
...
     private static void WriteAckMessage(ref MessagePackWriter writer, AckMessage message)
     {
-        writer.WriteArrayHeader(5);
+        writer.WriteArrayHeader(6);
         writer.Write(ServiceProtocolConstants.AckMessageType);
         writer.Write(message.AckId);
         writer.Write(message.Status);
+        writer.Write(message.Message);
+        message.WriteExtensionMembers(ref writer);
         if (message.Payload.HasValue)
```
vwxyzh, line 690: if `false`, we lost the last item in the array.

Y-Sindo: Updated

vwxyzh (approved):

PR #2125, opened 2025-01-15
Implement presence API for `IServiceConnectionContainer`

Y-Sindo:
To expose presence API to the caller in the `IServiceConnectionContainer` interface, we have two choices:
1. Expose a low-layer method like "WriteXXXMessage", which accepts a `GroupMemberQueryMessage` parameter and return `GroupMemberQueryResponse`. However, the `MultipleEndpointServiceConnectionContainer` also implements this interface and has to implement this method. In multi-endpoint case, we have to have some aggregation logic on the results from all the endpoints. Then the verb `Write` cannot describe what has been done in multiple endpoint cases. Therefore we don't expose a low-layer method.
2. Expose a high-layer method `ListConnectionsInGroup`, to wrap operations including writing messages, fetch next pages according to `continuationToken`.

### DIFF
--- a/src/Microsoft.Azure.SignalR.AspNet/ServerConnections/ServiceConnectionManager.cs
+++ b/src/Microsoft.Azure.SignalR.AspNet/ServerConnections/ServiceConnectionManager.cs
@@ -169,7 +169,7 @@ internal class ServiceConnectionManager : IServiceConnectionManager
         }
     }
 
-    public IAsyncEnumerable<GroupMember> ListConnectionsInGroupAsync(string groupName, int? top)
+    public IAsyncEnumerable<GroupMember> ListConnectionsInGroupAsync(string groupName, int? top = null, ulong? tracingId = null)
     {
         throw new NotSupportedException();
     }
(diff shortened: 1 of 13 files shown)


vwxyzh:
rough idea for n-way merge for multi-endpoint.
https://github.com/Azure/azure-signalr/pull/2126
```mermaid
sequenceDiagram
    participant c as Client
    participant s as SDK
    participant m as Multi Endpoint Service Connection Container
    participant s1 as Service Connection Container For Service 1
    participant s2 as Service Connection Container For Service 2

    c->>s: Enumerate group member
    s->>m: Enumerate group member
    loop
        opt Need more data from Service 1
            m->>s1: Enumerate group member with pagination
            s1->>m: Pagination result
        end
        opt Need more data from Service 2
            m->>s2: Enumerate group member with pagination
            s2->>m: Pagination result
        end
        loop
            m->>m: N-way merge
            m->>s: Yield return
            s->>c: Yield return
        end
    end
```

vwxyzh (approved):

vicancy (approved):

PR #2127, opened 2025-01-20
Update service protocol

Y-Sindo:
Rename the `Max` parameter to `Top` to make it clearer. 
Don't set a client-side default value for the `Top` parameter.

### DIFF
--- a/src/Microsoft.Azure.SignalR.Protocols/ServiceMessage.cs
+++ b/src/Microsoft.Azure.SignalR.Protocols/ServiceMessage.cs
@@ -590,7 +590,7 @@ namespace Microsoft.Azure.SignalR.Protocol
         public string GroupName { get; set; } = string.Empty;
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


vicancy (approved):

src/Microsoft.Azure.SignalR.Protocols/ServiceMessage.cs
```diff
@@ -590,9 +590,9 @@ public class GroupMemberQueryMessage : ExtensibleServiceMessage, IAckableMessage
         public string GroupName { get; set; } = string.Empty;
 
         /// <summary>
-        /// The max count of connections to return.
+        /// The max count of connections to return.
```
vwxyzh: remove the whitespace in the end.

Y-Sindo: Removed

PR #2154, opened 2025-02-18
Fix client invocation bug with multiple endpoints

xingsy97:
 - It's wrong for a client invocation to simply use the number of endpoints as its ack number. See https://github.com/Azure/azure-signalr/blob/v1.30.2/src/Microsoft.Azure.SignalR/ClientInvocation/CallerClientResultsManager.cs
 - Now we use the actual number of endpoints which will be sent with client invocation message.

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointMessageWriter.cs
+++ b/src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointMessageWriter.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft. All rights reserved.
+﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
@@ -9,6 +9,7 @@ using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
 
+using Microsoft.AspNetCore.SignalR.Protocol;
 using Microsoft.Azure.SignalR.Common;
 using Microsoft.Azure.SignalR.Protocol;
 using Microsoft.Extensions.Logging;
@@ -58,6 +59,13 @@ internal class MultiEndpointMessageWriter : IServiceMessageWriter, IPresenceMana
         {
             // Accroding to target endpoints in method `WriteMultiEndpointMessageAsync`
             _clientInvocationManager.Caller.SetAckNumber(invocationMessage.InvocationId, TargetEndpoints.Length);
+            if (TargetEndpoints.Length == 0)
+            {
+                _clientInvocationManager.Caller.TryCompleteResult(
+                    invocationMessage.ConnectionId,
+                    CompletionMessage.WithError(invocationMessage.InvocationId, "No available endpoint to send invocation message.")
+                );
+            }
         }
 
         return WriteMultiEndpointMessageAsync(serviceMessage, connection => connection.WriteAsync(serviceMessage));
--- a/src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointMessageWriter.cs
+++ b/src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointMessageWriter.cs
@@ -21,11 +21,13 @@ namespace Microsoft.Azure.SignalR;
 internal class MultiEndpointMessageWriter : IServiceMessageWriter, IPresenceManager
 {
     private readonly ILogger _logger;
+    private readonly IClientInvocationManager _clientInvocationManager;
 
     internal HubServiceEndpoint[] TargetEndpoints { get; }
 
-    public MultiEndpointMessageWriter(IReadOnlyCollection<ServiceEndpoint> targetEndpoints, ILoggerFactory loggerFactory)
+    public MultiEndpointMessageWriter(IReadOnlyCollection<ServiceEndpoint> targetEndpoints, IClientInvocationManager invocationManager, ILoggerFactory loggerFactory)
     {
+        _clientInvocationManager = invocationManager;
         _logger = loggerFactory.CreateLogger<MultiEndpointMessageWriter>();
         var normalized = new List<HubServiceEndpoint>();
         if (targetEndpoints != null)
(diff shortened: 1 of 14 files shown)


src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointMessageWriter.cs
```diff
@@ -52,6 +54,12 @@ public MultiEndpointMessageWriter(IReadOnlyCollection<ServiceEndpoint> targetEnd
 
     public Task WriteAsync(ServiceMessage serviceMessage)
     {
+        if (serviceMessage is ClientInvocationMessage invocationMessage)
+        {
+            // Accroding to target endpoints in method `WriteMultiEndpointMessageAsync`
+            _clientInvocationManager.Caller.SetAckNumber(invocationMessage.InvocationId, TargetEndpoints.Length);
```
TD2106, line 61: [ServiceLifetimeManager](https://github.com/Azure/azure-signalr/blob/5b7d1ed6426d1f8eea5888c0f8ea5511d8464452/src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManager.cs#L133)

In this implementation, I see that we are sending the messages first then add invocation later. And in SetAckNumber, the pending invocation cache is being check. Does the pending invocation contains the invocation id when it is checked at this time?

If not I propose to add a new dictionary to store the number of messages sent out. And when addInvocation is called we can use that dictionary to get the number of messages sent out.

src/Microsoft.Azure.SignalR/ClientInvocation/CallerClientResultsManager.cs
```diff
@@ -206,6 +206,14 @@ public void RemoveInvocation(string invocationId)
             _pendingInvocations.TryRemove(invocationId, out _);
         }
 
+        public void SetAckNumber(string invocationId, int ackNumber)
```
TD2106, line 204: [Old set ack](https://github.com/Azure/azure-signalr/blob/5b7d1ed6426d1f8eea5888c0f8ea5511d8464452/src/Microsoft.Azure.SignalR/ClientInvocation/CallerClientResultsManager.cs#L51)

Shouldn't this be removed?

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointMessageWriter.cs
vicancy, line 61: when `TargetEndpoints.Length` is 0, the result is OK?

xingsy97: class `AckHandler` could handle such condition correctly. Refer to its method `SetExpectedCount`

xingsy97: changed to adding invocation first, then send the message first

xingsy97: fix here https://github.com/Azure/azure-signalr/pull/2154/files#diff-1f07592019f7c2af9fcce1e14fae72fe2c02010db65cb61d9c822e4dda56d44cR62-R70

src/Microsoft.Azure.SignalR.Common/ServiceConnections/MultiEndpointMessageWriter.cs
```diff
@@ -52,6 +55,19 @@ public MultiEndpointMessageWriter(IReadOnlyCollection<ServiceEndpoint> targetEnd
...
```
vicancy, line 61: do we still need to SetAck if Length == 0?

src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManager.cs
```diff
@@ -129,8 +129,14 @@ public override async Task<T> InvokeConnectionAsync<T>(string connectionId, stri
 
             var invocationId = _clientInvocationManager.Caller.GenerateInvocationId(connectionId);
             var message = AppendMessageTracingId(new ClientInvocationMessage(invocationId, connectionId, _callerId, SerializeAllProtocols(methodName, args, invocationId)));
-            await WriteAsync(message);
+            // The ack number of invocation will be set inside `WriteAsync`. So adding invocation should be first.
             var task = _clientInvocationManager.Caller.AddInvocation<T>(_hub, connectionId, invocationId, cancellationToken);
+            await WriteAsync(message);
+            if (ServiceConnectionContainer is not MultiEndpointServiceConnectionContainer)
```
vicancy, line 135: sounds tricky, a ServiceLifttimeManager should not have knowledge of what the container type is, it looks like a strong assumption and is fragile. What if we later change to use other container?

src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManager.cs
```diff
@@ -129,8 +129,14 @@ public override async Task<T> InvokeConnectionAsync<T>(string connectionId, stri
...
+            {
+                // `WriteAsync` in test class `TestServiceConnectionHandler` does not set ack number. Set the number manually.
+                _clientInvocationManager.Caller.SetAckNumber(invocationId, 1);
```
vicancy, line 138: we don't want such logic spread into multiple layers, choose one place for such logic

test/Microsoft.Azure.SignalR.Tests/ClientInvocationManagerTests.cs
```diff
@@ -151,17 +160,18 @@ public async Task TestCompleteWithRouterServer(string protocol, bool isCompletio
...
 
     [Fact]
     public void TestCallerManagerCancellation()
     {
-        var clientInvocationManager = GetTestClientInvocationManager();
+        var clientInvocationManager = GetTestClientInvocationManager(1);
         var invocationId = clientInvocationManager.Caller.GenerateInvocationId(TestConnectionIds[0]);
         var cts = new CancellationTokenSource();
         var task = clientInvocationManager.Caller.AddInvocation<string>("TestHub", TestConnectionIds[0], invocationId, cts.Token);
+        clientInvocationManager.Caller.SetAckNumber(invocationId, 1);
```
vicancy, line 174: all the test set acknumber manually? how we test if this SetAckNumber logic works well with invocation?

vicancy:
I am not sure if I like this approach. How about having WriteAsync return the successful write count? @vwxyzh any thoughts?

PR #2173, opened 2025-02-28
Fix the protocol issue when null is in the array 

vicancy:
In old sdk wedon't do null check inside an array. This caused issue when new protocol package is used to parse the flaky message when null is in the list.

### DIFF
--- a/src/Microsoft.Azure.SignalR.Protocols/MessagePackUtils.cs
+++ b/src/Microsoft.Azure.SignalR.Protocols/MessagePackUtils.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
@@ -125,7 +125,7 @@ internal static class MessagePackUtils
 
     internal static string ReadStringNotNull(ref MessagePackReader reader, string field)
     {
-        string? result = null;
+        string? result;
         try
         {
             result = reader.ReadString();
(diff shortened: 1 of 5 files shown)


src/Microsoft.Azure.SignalR.Protocols/MessagePackUtils.cs
```diff
@@ -156,19 +156,23 @@ internal static string ReadStringNotNull(ref MessagePackReader reader, string fi
...
     }
 
-    internal static string[] ReadStringArray(ref MessagePackReader reader, string field)
+    internal static string[] ReadStringArrayExcludeNull(ref MessagePackReader reader, string field)
     {
         var arrayLength = ReadArrayLength(ref reader, field);
         if (arrayLength > 0)
         {
-            var array = new string[arrayLength];
+            var list = new List<string>();
```
vwxyzh: `new List<string>(arrayLength)`?

test/Microsoft.Azure.SignalR.Protocols.Tests/ServiceMessageEqualityComparer.cs
```diff
@@ -487,6 +487,45 @@ private static bool HeadersEqual(IDictionary<string, StringValues> x, IDictionar
...
+        {
+            var moveNext = enumerator.MoveNext();
+            while (moveNext && enumerator.Current == null)
+            {
+                moveNext = enumerator.MoveNext();
+            }
+            return moveNext;
+        }
+
+        private static bool SequenceEqualSkipNull(object left, object right)
```
vwxyzh, line 500: What about following:
```cs
private static bool SequenceEqualSkipNull<T>(IEnumerable<T> left, IEnumerable<T> right) =>
    left.Where(x => x != null).Zip(right.Where(y => y != null), (x, y) => Equals(x, y)).All(z => z);
```

src/Microsoft.Azure.SignalR.Protocols/MessagePackUtils.cs
vicancy: arrayLength is long while capacity is int

vicancy: int should be enough

Y-Sindo:
image
Would this one be a breaking change too?

vwxyzh (approved):

PR #2192, opened 2025-04-09
Add null check

### LINKED ISSUE #2190: ServiceConnectionManager<T> throws NullReferenceException on DisposeAsync() if SetServiceConnection(...) was never called
### Describe the bug

When using AddAzureSignalR() in an ASP.NET Core app, if another service fails during startup (e.g., due to invalid configuration and such), the app may throw before reaching app.Run() — and before SignalR is fully initialized.

In this case, ServiceConnectionManager<T> never calls SetServiceConnection(...).

However, during host shutdown, SignalR still attempts to dispose the uninitialized ServiceConnectionManager<T>, resulting in an unhandled NullReferenceException.

### To Reproduce

1. Configure SignalR via:

`builder.Services.AddSignalR().AddAzureSignalR(options =>
{
    options.ConnectionString = config["Azure:SignalR:ConnectionString"];
});
`

2. Register a dependent service with invalid config:

`services.AddSingleton<MyService>(provider =>
{
    var baseUrl = config["MyService:BaseUrl"];
    if (string.IsNullOrEmpty(baseUrl))
        throw new ArgumentException("BaseUrl cannot be empty");
    return new MyService(baseUrl);
});
`
3. The app fails before app.Run() due to ArgumentException.

4. Host shutdown triggers, DisposeAsync() is called on all services.

5. SignalR’s internal ServiceConnectionManager<T>.StopAsync() fails due to _serviceConnection being null.

System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.Azure.SignalR.ServiceConnectionManager`1.StopAsync()
   at Microsoft.Azure.SignalR.ServiceConnectionManager`1.Dispose()
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.ServiceProviderEngineScope.DisposeAsync()
   

### Exceptions (if any)

* DisposeAsync() should guard against uninitialized internal state.

* If SetServiceConnection(...) was never called, StopAsync() should be skipped or handled gracefully.

* Alternatively, provide a public flag or lifecycle hook to safely detect incomplete initialization.

### Further technical details
- Azure SignalR SDK version latest
- Your Server ASPNETCORE version or Assembly version of `Microsoft.AspNetCore.SignalR`  : 8.0

vicancy:
Fix https://github.com/Azure/azure-signalr/issues/2190

### DIFF
--- a/src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnectionManager.cs
+++ b/src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnectionManager.cs
@@ -24,7 +24,7 @@ internal class ServiceConnectionManager<THub> : IDisposable, IServiceConnectionM
 
     public Task StartAsync()
     {
-        return _serviceConnection?.StartAsync() ?? throw new ArgumentNullException(nameof(_serviceConnection));
+        return _serviceConnection?.StartAsync() ?? throw new InvalidOperationException();
     }
 
     public Task StopAsync()
--- a/src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnectionManager.cs
+++ b/src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnectionManager.cs
@@ -10,11 +10,12 @@ using Microsoft.AspNetCore.SignalR;
 using Microsoft.Azure.SignalR.Common;
 using Microsoft.Azure.SignalR.Protocol;
 
+#nullable enable
 namespace Microsoft.Azure.SignalR;
 
 internal class ServiceConnectionManager<THub> : IDisposable, IServiceConnectionManager<THub> where THub : Hub
 {
-    private IServiceConnectionContainer _serviceConnection;
+    private IServiceConnectionContainer? _serviceConnection;
 
     public void SetServiceConnection(IServiceConnectionContainer serviceConnection)
     {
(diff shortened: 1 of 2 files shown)


src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnectionManager.cs
```diff
@@ -23,22 +24,22 @@ public void SetServiceConnection(IServiceConnectionContainer serviceConnection)
 
     public Task StartAsync()
     {
-        return _serviceConnection.StartAsync();
+        return _serviceConnection?.StartAsync() ?? throw new ArgumentNullException(nameof(_serviceConnection));
```
vwxyzh: it is not an argument.

alejandropadillav1: Thanks for pointing that out! I really appreciate you taking the time to review this. If you have any suggestions on how to better handle this case, I’d love to get your input so we can move forward with a proper fix.

alejandropadillav1 (approved):

src/Microsoft.Azure.SignalR/ServerConnections/ServiceConnectionManager.cs
vicancy: updated to invalidoperationexception

vwxyzh (approved):

Issue #2194, opened 2025-04-20
HubEmulator Groups.RemoveFromAllGroupsAsync generating null pointer error


rikschuiling:
### Describe the bug
When a connection is in zero groups and the Groups.RemoveFromAllGroupsAsync method gets called there will be thrown a null pointer exception. See the exception below.

### To Reproduce
Simply call the method after a connection is made.

### Exceptions (if any)
fail: Microsoft.AspNetCore.Server.Kestrel[13]
      Connection id "0HNC087N6J696", Request id "0HNC087N6J696:00000001": An unhandled exception was thrown by the application.
      System.NullReferenceException: Object reference not set to an instance of an object.
         at Microsoft.Azure.SignalR.Emulator.HubEmulator.ManyToManyMap`2.RemoveLeft(TLeft left) in D:\a\_work\1\s\src\Microsoft.Azure.SignalR.Emulator\HubEmulator\ManyToManyMap.cs:line 91
         at Microsoft.Azure.SignalR.Emulator.HubEmulator.GroupManager.RemoveConnectionFromAllGroups(String connectionId) in D:\a\_work\1\s\src\Microsoft.Azure.SignalR.Emulator\HubEmulator\GroupManager.cs:line 39
         at Microsoft.Azure.SignalR.Emulator.Controllers.SignalRServiceEmulatorWebApi.RemoveConnectionFromAllGroups(String hub, String connectionId, String application) in D:\a\_work\1\s\src\Microsoft.Azure.SignalR.Emulator\Controllers\SignalRServiceEmulatorWebApi.cs:line 230
         at lambda_method3(Closure, Object, Object[])
         at Microsoft.AspNetCore.Mvc.Infrastructure.ActionMethodExecutor.SyncActionResultExecutor.Execute(ActionContext actionContext, IActionResultTypeMapper mapper, ObjectMethodExecutor executor, Object controller, Object[] arguments)
         at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.InvokeActionMethodAsync()
         at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.Next(State& next, Scope& scope, Object& state, Boolean& isCompleted)
         at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.InvokeNextActionFilterAsync()
      --- End of stack trace from previous location ---
[...]

PR #2196, opened 2025-04-24
Fix emulator bug #2194

vicancy:


src/Microsoft.Azure.SignalR.Emulator/HubEmulator/ManyToManyMap.cs
```diff
@@ -1,6 +1,3 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.
-
```
JialinXin, line 3: why remove license info?

JialinXin (approved):

PR #2201, opened 2025-06-02
Check for null hub name

### LINKED ISSUE #2200: Misleading error when hubName is null
### Describe the bug
If you are application is misconfigured, and a null hub name is provided then the following exception is thrown.

```
System.InvalidOperationException: 'A suitable constructor for type 'Microsoft.Azure.SignalR.Management.RestHealthCheckService' could not be located. Ensure the type is concrete and all parameters of a public constructor are either registered as services or passed as arguments. Also ensure no extraneous arguments are provided.'
``` 

I found the error to be a misleading.

### To Reproduce

Attempt to create a hub context where the hub name is null

```
 using var serviceManager = new ServiceManagerBuilder()

     .WithConfiguration(_configuration)
     //or .WithOptions(o=>o.ConnectionString = _configuration["Azure:SignalR:ConnectionString"]
     .WithLoggerFactory(_loggerFactory)
     .BuildServiceManager();
 var hubContext = await serviceManager.CreateHubContextAsync(null, cancellationToken);
```

### Exceptions (if any)

System.InvalidOperationException: 'A suitable constructor for type 'Microsoft.Azure.SignalR.Management.RestHealthCheckService' could not be located. Ensure the type is concrete and all parameters of a public constructor are either registered as services or passed as arguments. Also ensure no extraneous arguments are provided.'

ghost:
Adds a guard to check for a null hub name and throws appropriate error.

Fixes #2200

### DIFF
--- a/src/Microsoft.Azure.SignalR.Management/DependencyInjectionExtensions.cs
+++ b/src/Microsoft.Azure.SignalR.Management/DependencyInjectionExtensions.cs
@@ -52,6 +52,15 @@ internal static class DependencyInjectionExtensions
     public static IServiceCollection AddHub<THub>(this IServiceCollection services, string hubName)
         where THub : Hub
     {
+#if NET6_0_OR_GREATER
+        ArgumentNullException.ThrowIfNull(hubName);
+#else
+        if (hubName is null)
+        {
+            throw new ArgumentNullException(nameof(hubName));
+        }
+#endif
+
         //for persistent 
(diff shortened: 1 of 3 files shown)


ghost:
@microsoft-github-policy-service agree

vicancy (approved):

ghost:
@vicancy I think the build may have failed due to a transient error. Is there a way to re-trigger it? Thanks.

PR #2205, opened 2025-07-17
Implement `ListConnectionsInGroup` of WebSocket tranpsort

Y-Sindo:


src/Microsoft.Azure.SignalR.AspNet/ServerConnections/ServiceConnectionManager.cs
```diff
@@ -141,6 +143,11 @@ public Task StopGetServersPing()
         return Task.WhenAll(GetConnections().Select(s => s.StopGetServersPing()));
     }
 
+    public IAsyncEnumerable<Page<GroupMember>> ListConnectionsInGroupAsync(string groupName, int? top = null, int? maxPageSize = null, string continuationToken = null, ulong? tracingId = null, CancellationToken token = default)
```
vwxyzh, line 146: should we remove the `continuationToken`?

vwxyzh: or return a `Task<Page<GroupMember>` with `continuationToken`

src/Microsoft.Azure.SignalR.Common/Pagination/GroupMemberQueryResultPage.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    private readonly IReadOnlyList<GroupMember> _value;
+    private readonly string? _continuationToken;
+
+    public GroupMemberQueryResultPage(IReadOnlyList<GroupMember> value, string? continuationToken)
+    {
+        _value = value;
+        _continuationToken = continuationToken;
+    }
+    public override IReadOnlyList<GroupMember> Values => _value;
```
vwxyzh, line 26: add a blank line.

src/Microsoft.Azure.SignalR.AspNet/ServerConnections/ServiceConnectionManager.cs
Y-Sindo: > should we remove the continuationToken?

The `AsyncPagenable<T>` provides a method to pass `continuationToken`, so I expose it.

```cs
public abstract IAsyncEnumerable<Page<T>> AsPages(string? continuationToken = null, int? pageSizeHint = null);
```

vwxyzh (dismissed):

vwxyzh (approved):

PR #2209, opened 2025-07-21
Remove outdated job line_endings check

Y-Sindo:


.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text=auto
```
vicancy, line 2: use more strict `text eol=lf`?

vicancy (approved):

.gitattributes
Y-Sindo: `auto` is more friendly to Windows users.

PR #2210, opened 2025-07-23
Refactor: use `HttpMessageHandler` to add access token

Y-Sindo:
Motivation: In `ListConnectionsInGroup` API, the next page is referenced by an URI, and we need to generate access token for that URI. However, currently the access token generation is tightly coupled with `RestApiProvider`, which generates access token for fixed endpoint.

This PR moves the logic of handling access token out of `RestApiProvider` and `RestClient`. Instead, add a `HttpMessageHandler` to set `Authorization` header.


src/Microsoft.Azure.SignalR.Management/AccessTokenHttpMessageHandler.cs
```diff
@@ -0,0 +1,38 @@
...
+
+#nullable enable
+
+/// <summary>
+/// Appends access token for each HTTP request
+/// </summary>
+internal class AccessTokenHttpMessageHandler : DelegatingHandler
+{
+    // Internal for testing purposes
+    internal static readonly string ServerName = $"{Environment.MachineName}_{Guid.NewGuid():N}";
```
vicancy: reuse ServerNameProvider?

src/Microsoft.Azure.SignalR.Management/AccessTokenHttpMessageHandler.cs
```diff
@@ -0,0 +1,38 @@
...
+
+    private readonly RestApiAccessTokenGenerator _accessTokenGenerator;
+
+    public AccessTokenHttpMessageHandler(IServiceEndpointManager serviceEndpointManager)
+    {
+        var endpoint = serviceEndpointManager.Endpoints.Keys.First();
```
vicancy: why only First

src/Microsoft.Azure.SignalR.Management/DependencyInjectionExtensions.cs
```diff
@@ -229,8 +231,10 @@ private static IServiceCollection AddRestClient(this IServiceCollection services
                 ConfigureMessageTracingId(sp, client);
             })
             .ConfigurePrimaryHttpMessageHandler(ConfigureProxy)
+            // The first HttpMessageHandler executes first.
             .AddHttpMessageHandler(sp => ActivatorUtilities.CreateInstance<RetryHttpMessageHandler>(sp, (HttpStatusCode code) => IsTransientErrorForNonMessageApi(code)))
-            .AddHttpMessageHandler(sp => ActivatorUtilities.CreateInstance<TimeoutHttpMessageHandler>(sp));
+            .AddHttpMessageHandler(sp => ActivatorUtilities.CreateInstance<TimeoutHttpMessageHandler>(sp))
```
vicancy:
```
.AddHttpMessageHandler<TimeoutHttpMessageHandler>()
.AddHttpMessageHandler<AccessTokenHttpMessageHandler>()
```

src/Microsoft.Azure.SignalR.Management/AccessTokenHttpMessageHandler.cs
```diff
@@ -0,0 +1,38 @@
...
+        _accessTokenGenerator = new RestApiAccessTokenGenerator(endpoint.AccessKey, ServerName);
+    }
+
+    protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+    {
+        var audience = request.RequestUri!.GetLeftPart(UriPartial.Path);
```
vicancy: reuse the previous logic using endpoint.AudienceBaseUrl?

src/Microsoft.Azure.SignalR.Management/AccessTokenHttpMessageHandler.cs
Y-Sindo: Transient mode supports only one `ServiceEndpoint`.

src/Microsoft.Azure.SignalR.Management/AccessTokenHttpMessageHandler.cs
Y-Sindo: UPdate

src/Microsoft.Azure.SignalR.Management/AccessTokenHttpMessageHandler.cs
vicancy: what if later multiple endpoints are supported, this will be a hidden hole

vicancy: what if transient mode add multiple service endpoints, will we throw?

vicancy: interesting, I do feel we should support multiple service endpoint for transient mode, this sounds like a bug it is not supported and it is not thrown

test/Microsoft.Azure.SignalR.Management.Tests/AccessTokenHttpMessageHandlerTests.cs
```diff
@@ -0,0 +1,67 @@
...
+namespace Microsoft.Azure.SignalR.Management.Tests
+{
+    public class AccessTokenHttpMessageHandlerTests
+    {
+        [Fact]
+        public async Task SendAsync_ShouldSetAuthorizationHeader()
+        {
+            var key = FakeEndpointUtils.FakeAccessKey;
+            var mockServiceEndpointManager = new Mock<IServiceEndpointManager>();
+            var mockEndpoint = new ServiceEndpoint($"Endpoint=https://test.service.signalr.net;AccessKey={key};Version=1.0;");
```
vicancy, line 29: could we aslo test different types of connectionString, like type=aad, having ClientEndpoint/ServerEndpoint values

vicancy (approved):
My main concern is the multi-endpoint support for transient mode, others look good to me.

test/Microsoft.Azure.SignalR.Management.Tests/AccessTokenHttpMessageHandlerTests.cs
Y-Sindo: How to generate access token is the job of access token generator, out of the scope of this test.

src/Microsoft.Azure.SignalR.Management/AccessTokenHttpMessageHandler.cs
Y-Sindo: We previously discussed and agreed to prioritize multi-endpoint support for persistent mode first. This decision was based on the fact that transient mode lacks endpoint matrices for routing, and persistent mode tends to be more efficient, particularly in high-traffic multi-endpoint scenarios. As there are no customer requests for multi-endpoint of transient mode, and we have geo-replica features available, which also serves as a solution for resilience and high availability, the support of transient mode is not prioritized.

In terms of potential issues, any unauthorized exceptions would be detected during basic testing, so there shouldn't be a hidden hole. Additionally, we do throw errors if users attempt to configure multiple endpoints in transient mode.

PR #2215, opened 2025-08-08
Fix issue caused by AckHandler disposing

Y-Sindo:
* Don't dispose of the global singleton of `AckHandler`. We use a global singleton of `AckHandler`, which should never be disposed of. 
* Throw `InvalidOperation` when accessing a disposed `AckHandler` instead of returning a positive result.
* Create a separate `AckHandler` in `MultiEndpointServiceConnectionContainerTests`. The tests are based on the condition that each `ServiceConnectionContainerBase` has its own `AckHandler`, and they don't work when we have a global `AckHandler`. The false positive result fixed by the above change made those tests pass falsely.

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
+++ b/src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
@@ -457,8 +457,12 @@ internal abstract class ServiceConnectionContainerBase : IServiceConnectionConta
     {
         if (disposing)
         {
+            // We shouldn't dispose the global one, which may be used by other instances.
+            if (_ackHandler != AckHandler.Singleton)
+            {
             _ackHandler.Dispose();
         }
+    }
     }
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.Azure.SignalR.Common/ServiceConnections/ServiceConnectionContainerBase.cs
```diff
@@ -457,9 +457,13 @@ protected virtual void Dispose(bool disposing)
     {
         if (disposing)
         {
+            // We shouldn't dispose the global one, which may be used by other instances.
+            if (_ackHandler != AckHandler.Singleton)
+            {
             _ackHandler.Dispose();
         }
     }
```
xingsy97, line 465: indent

Y-Sindo: Emmm Somehow it looks normal on IDE.

xingsy97 (approved):

PR #2216, opened 2025-08-08
Remove .net 6 from target frameworks

Y-Sindo:


src/Microsoft.Azure.SignalR.Common/Microsoft.Azure.SignalR.Common.csproj
```diff
@@ -6,7 +6,7 @@
   </PropertyGroup>
   
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net6.0;net8.0;net462;</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net8.0;net462;</TargetFrameworks>
```
vicancy, line 9: release a version first, and increase a minor version for removing net6

vicancy (dismissed):

src/Microsoft.Azure.SignalR.Common/Microsoft.Azure.SignalR.Common.csproj
Y-Sindo: We just released a version.

xingsy97 (approved):

PR #2217, opened 2025-08-11
Improve presence API

Y-Sindo:
* Expose a dedicated model class `SignalRGroupConnection` instead of `GroupMember` from SignalR service protocol.

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/Pagination/GroupMemberQueryResultPage.cs
+++ b/src/Microsoft.Azure.SignalR.Common/Pagination/GroupMemberQueryResultPage.cs
@@ -60,12 +60,12 @@ internal class GroupMemberQueryResultPageConverter : JsonConverter<GroupMemberQu
             string propertyName = reader.GetString()!;
             reader.Read();
 
-            switch (propertyName.ToLowerInvariant())
+            switch (propertyName)
             {
-                case "value":
+                case string s when s.Equals("value", StringComparison.OrdinalIgnoreCase):
                     values = JsonSerializer.Deserialize<List<SignalRGroupConnection>>(ref reader, options);
                     break;
-                case "nextlink":
+                case string s when s.Equals("nextlink", StringComparison.OrdinalIgnoreCase):
                     continuationToken = JsonSerializer.Deserialize<string>(ref reader, options);
                     break;
                 default:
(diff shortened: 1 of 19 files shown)


src/Microsoft.Azure.SignalR.Common/Pagination/GroupMemberQueryResultPage.cs
```diff
@@ -34,3 +32,58 @@ public override Response GetRawResponse()
         throw new NotSupportedException();
     }
 }
+
+internal class GroupMemberQueryResultPageConverter : JsonConverter<GroupMemberQueryResultPage>
```
vicancy, line 36: why not simply use `[JsonProperty("value")]`/`[JsonProperty("nextlink")]`

src/Microsoft.Azure.SignalR.Management/HubContext/GroupManager.cs
```diff
@@ -50,6 +49,6 @@ public abstract class GroupManager : IGroupManager
         /// <param name="top">The maximum number of connections to return.</param>
         /// <param name="cancellationToken">The cancellation token.</param>
         /// <returns>An asynchronous enumerable of group members.</returns>
-        internal virtual AsyncPageable<GroupMember> ListConnectionsInGroup(string groupName, int? top = null, CancellationToken cancellationToken = default) => throw new NotImplementedException();
+        internal virtual AsyncPageable<SignalRGroupConnection> ListInGroup(string groupName, int? top = null, CancellationToken cancellationToken = default) => throw new NotImplementedException();
```
vicancy: I would prefer `GetGroupConnections`

src/Microsoft.Azure.SignalR.Common/Pagination/GroupMemberQueryResultPage.cs
Y-Sindo: If we use `JsonProperty("value")`, we need to load the JSON object into memory first, but in this way, we don't need to load the JSON object, just process it streamingly.

vicancy (dismissed):

src/Microsoft.Azure.SignalR.Common/Pagination/GroupMemberQueryResultPage.cs
```diff
@@ -34,3 +32,58 @@ public override Response GetRawResponse()
...
+
+            if (reader.TokenType != JsonTokenType.PropertyName)
+            {
+                throw new JsonException("Expected PropertyName token");
+            }
+
+            string propertyName = reader.GetString()!;
+            reader.Read();
+
+            switch (propertyName.ToLowerInvariant())
```
vicancy: avoid allocation?
```cs
switch (propertyName)
{
    case string s when s.Equals("value", StringComparison.OrdinalIgnoreCase):

vicancy (approved):

PR #2221, opened 2025-09-09
Add presence API REST implementation

Y-Sindo:


src/Microsoft.Azure.SignalR.Common/Utilities/RestClient.cs
```diff
@@ -67,6 +67,15 @@ public Task SendWithRetryAsync(
         return SendAsyncCore(Constants.HttpClientNames.Resilient, api, httpMethod, null, null, AsAsync(handleExpectedResponse), cancellationToken);
```
vicancy: update to call the newly added method?

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
```diff
@@ -357,6 +359,76 @@ private static bool FilterExpectedResponse(HttpResponseMessage response, string
...
+                api.Query ??= new Dictionary<string, StringValues>();
+                api.Query["maxPageSize"] = pageSizeHint.Value.ToString(CultureInfo.InvariantCulture);
+            }
+            if (!string.IsNullOrEmpty(continuationToken))
+            {
+                api.Query ??= new Dictionary<string, StringValues>();
+                api.Query["continuationToken"] = continuationToken;
+            }
+            do
+            {
```
vicancy, line 396: we could also check if token.IsCancellationRequest then throw

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
```diff
@@ -357,6 +359,76 @@ private static bool FilterExpectedResponse(HttpResponseMessage response, string
...
+
+        async Task<Page<SignalRGroupConnection>> FetchSinglePage(RestApiEndpoint api, CancellationToken cancellationToken = default)
+        {
+            var page = default(Page<SignalRGroupConnection>);
+
+            await _restClient.SendWithRetryAsync(api, HttpMethod.Get, async response =>
+            {
+                if (!response.IsSuccessStatusCode)
+                {
+                    return false;
```
vicancy, line 425: what happens when return false?

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
```diff
@@ -357,6 +359,76 @@ private static bool FilterExpectedResponse(HttpResponseMessage response, string
...
+                api.Query["maxPageSize"] = pageSizeHint.Value.ToString(CultureInfo.InvariantCulture);
+            }
+            if (!string.IsNullOrEmpty(continuationToken))
+            {
+                api.Query ??= new Dictionary<string, StringValues>();
+                api.Query["continuationToken"] = continuationToken;
+            }
+            do
+            {
+                var page = await FetchSinglePage(api, token);
```
vicancy, line 397: what happens when FetchSinglePage fails, e.g. `!response.IsSuccessStatusCode`?

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
Y-Sindo: It will go into common error handling process, such as 400 invalid request, 401 not authorized.

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
Y-Sindo: Exception will be thrown from `RestClient`.

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
Y-Sindo: `HttpClient.SendAsync` has this check already. No need to check by ourselves.

>[ Consider not checking the token at all if your work is very quick, or you propagate it to the methods you call. That said, calling CancellationToken.ThrowIfCancellationRequested() is pretty lightweight so don’t think too hard about this one unless you see it on perf traces.](https://devblogs.microsoft.com/premier-developer/recommended-patterns-for-cancellationtoken/)

vicancy (approved):

PR #2244, opened 2026-01-21
Add TryParseMessage for JsonObjectSerializerHubProtocol

ZSWY666:
Add TryParseMessage for JsonObjectSerializerHubProtocol

### DIFF
--- a/src/Microsoft.Azure.SignalR.Management/Serialization/JsonObjectSerializerHubProtocol.cs
+++ b/src/Microsoft.Azure.SignalR.Management/Serialization/JsonObjectSerializerHubProtocol.cs
@@ -90,7 +90,6 @@ namespace Microsoft.Azure.SignalR.Management
 #if NET7_0_OR_GREATER
         public bool TryParseMessage(ref ReadOnlySequence<byte> input, IInvocationBinder binder, [NotNullWhen(true)] out HubMessage? message)
         {
-
             if (!TextMessageParser.TryParseMessage(ref input, out var payload))
             {
                 message = null!;
@@ -193,17 +192,10 @@ namespace Microsoft.Azure.SignalR.Management
 
                         message = BindCompletionMessage(invocationId, error, result, hasResult);
                         break;
-                    case HubProtocolConstants.InvocationMessageType:
-                    case HubProtocolConstants.StreamInvocationMessageType:
-                    case HubProtocolConstants.StreamItemMessageType:
-                    case HubProtocolConstants.CancelInvocationMessageType:
-                    case HubProtocolConstants.PingMessageType:
-                    case HubProtocolConstants.CloseMessageType:
-                        throw new NotSupportedException($"Not supported message type: {type}.");
                     case null:
                         throw new InvalidDataException($"Missing required property '{TypePropertyName}'.");
                     default:
-                        return null!;
+                        throw new NotSupportedException($"Not supported message type: {type}.");
                 }
                 return message;
             }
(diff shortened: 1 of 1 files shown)


src/Microsoft.Azure.SignalR.Management/Serialization/JsonObjectSerializerHubProtocol.cs
```diff
@@ -86,11 +87,162 @@ public bool IsVersionSupported(int version)
...
+                    case HubProtocolConstants.StreamInvocationMessageType:
+                    case HubProtocolConstants.StreamItemMessageType:
+                    case HubProtocolConstants.CancelInvocationMessageType:
+                    case HubProtocolConstants.PingMessageType:
+                    case HubProtocolConstants.CloseMessageType:
+                        throw new NotSupportedException($"Not supported message type: {type}.");
+                    case null:
+                        throw new InvalidDataException($"Missing required property '{TypePropertyName}'.");
+                    default:
+                        return null!;
```
Y-Sindo: We should try to avoid return "null" value for invalid cases. Move the case of `null` to the second position, then combine all the other cases into `default`, and  throw new NotSupportedException($"Not supported message type: {type}.");

src/Microsoft.Azure.SignalR.Management/Serialization/JsonObjectSerializerHubProtocol.cs
```diff
@@ -86,11 +87,162 @@ public bool IsVersionSupported(int version)
             return version == Version;
         }
 
+#if NET7_0_OR_GREATER
+        public bool TryParseMessage(ref ReadOnlySequence<byte> input, IInvocationBinder binder, [NotNullWhen(true)] out HubMessage? message)
+        {
+
```
vicancy: nit: extra line

src/Microsoft.Azure.SignalR.Management/Serialization/JsonObjectSerializerHubProtocol.cs
vicancy: by default throw NotSupported?

src/Microsoft.Azure.SignalR.Management/Serialization/JsonObjectSerializerHubProtocol.cs
```diff
@@ -86,11 +87,162 @@ public bool IsVersionSupported(int version)
...
+#else
         public bool TryParseMessage(ref ReadOnlySequence<byte> input, IInvocationBinder binder, out HubMessage message)
         {
             //We don't need reading message with this protocol.
             throw new NotSupportedException();
         }
+#endif
+
+#if NET7_0_OR_GREATER
+        private HubMessage ParseMessage(ReadOnlySequence<byte> input, IInvocationBinder binder)
```
vicancy, line 112: could we follow similar logic as https://github.com/dotnet/aspnetcore/blob/74afc71a482ecb97b335fbc39840c48c26683484/src/SignalR/common/Protocols.Json/src/Protocol/JsonHubProtocol.cs#L115

vicancy (commented):
some little comments

src/Microsoft.Azure.SignalR.Management/Serialization/JsonObjectSerializerHubProtocol.cs
ZSWY666: Yes we are following the same logic here.

src/Microsoft.Azure.SignalR.Management/Serialization/JsonObjectSerializerHubProtocol.cs
ZSWY666: refined

src/Microsoft.Azure.SignalR.Management/Serialization/JsonObjectSerializerHubProtocol.cs
ZSWY666: removed

Y-Sindo (approved):

src/Microsoft.Azure.SignalR.Management/Serialization/JsonObjectSerializerHubProtocol.cs
vicancy: I mean the same parsing logic for example using `Utf8JsonReader`

PR #2245, opened 2026-01-21
add protocolResolver to manager

ZSWY666:
add protocolResolver to manager

### DIFF
--- a/src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
+++ b/src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
@@ -367,8 +367,8 @@ internal class RestHubLifetimeManager<THub> : HubLifetimeManager<THub>, IService
         // Use protocol.Name to check for supported protocols
         switch (protocol.Name)
         {
-            case "json":
-            case "messagepack":
+            case Constants.Protocol.Json:
+            case Constants.Protocol.MessagePack:
                 return true;
             default:
                 return false;
--- a/src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
+++ b/src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
@@ -14,6 +14,9 @@ using System.Threading.Tasks;
 using Azure;
 
 using Microsoft.AspNetCore.SignalR;
+#if NET7_0_OR_GREATER
+using Microsoft.AspNetCore.SignalR.Protocol;
+#endif
 using Microsoft.Extensions.Primitives;
 
 using static Microsoft.Azure.SignalR.Constants;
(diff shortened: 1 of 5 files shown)


src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
```diff
@@ -353,6 +358,25 @@ public async Task SendStreamCompletionAsync(string connectionId, string streamId
         await _restClient.SendWithRetryAsync(api, HttpMethod.Post, handleExpectedResponse: null, cancellationToken: cancellationToken);
     }
 
+#if NET7_0_OR_GREATER
+
+#pragma warning disable IDE0051 // Will be used in the future updates
+    private static bool IsInvocationSupported(IHubProtocol protocol)
```
vicancy, line 364: there is a quite similar logic https://github.com/Azure/azure-signalr/blob/0d390b28999b5e0f7f164fc00cc4f3e6d817ad08/src/Microsoft.Azure.SignalR.Management/ServiceManagerBuilder.cs#L136

extract to a shared static method?

src/Microsoft.Azure.SignalR.Management/WebsocketsHubLifetimeManager.cs
```diff
@@ -269,6 +272,11 @@ public override async Task<T> InvokeConnectionAsync<T>(string connectionId, stri
             throw new ArgumentNullException(nameof(methodName));
         }
 
+        if (!ProtocolResolver.AllProtocols.All(IsInvocationSupported))
```
vicancy: is this the only place when ProtocolResolver is used? if yes, why we need this parameter inside base class?

ZSWY666: It will be used in the client invocation in transient mode, which will be added in the later PR

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
ZSWY666: talked offline, will keep the code as they are handling diff scope

src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManagerBase.cs
```diff
@@ -22,12 +22,14 @@ internal abstract class ServiceLifetimeManagerBase<THub> : HubLifetimeManager<TH
     protected ILogger Logger { get; set; }
 
     private readonly DefaultHubMessageSerializer _messageSerializer;
+    private readonly IHubProtocolResolver _protocolResolver;
```
vicancy:
```cs
protected IHubProtocolResolver HubProtocolResolver { get; init; }
// in .ctor:
HubProtocolResolver = protocolResolver;
...

```

vicancy (dismissed):

src/Microsoft.Azure.SignalR/HubHost/ServiceLifetimeManagerBase.cs
ZSWY666: refined

vicancy (approved):

PR #2246, opened 2026-01-21
enable add accepts while building requests

ZSWY666:
enable add accepts while building requests

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/Utilities/RestClient.cs
+++ b/src/Microsoft.Azure.SignalR.Common/Utilities/RestClient.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Net;
 using System.Net.Http;
+using System.Net.Http.Headers;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


src/Microsoft.Azure.SignalR.Common/Utilities/RestClient.cs
```diff
@@ -73,18 +73,30 @@ public Task SendWithRetryAsync(
...
+        return SendAsyncCore(Constants.HttpClientNames.MessageResilient, api, httpMethod, new InvocationMessage(methodName, args), null, handleExpectedResponse, null, cancellationToken);
+    }
+
+    public Task SendMessageWithRetryAsync(
+        RestApiEndpoint api,
+        HttpMethod httpMethod,
+        string methodName,
+        object?[] args,
+        Func<HttpResponseMessage, Task<bool>>? handleExpectedResponse = null,
+        string? accepts = null,
```
vicancy: directly use `MediaTypeWithQualityHeaderValue` strong type here?

ZSWY666: refined

vicancy (approved):

PR #2249, opened 2026-01-27
Add Rest support for client invocation in management SDK

ZSWY666:
 - Detail 1
 - Detail 2

### DIFF
--- a/src/Microsoft.Azure.SignalR.Common/Utilities/Rest/JsonPayloadContentBuilder.cs
+++ b/src/Microsoft.Azure.SignalR.Common/Utilities/Rest/JsonPayloadContentBuilder.cs
@@ -24,6 +24,4 @@ internal class JsonPayloadContentBuilder : IPayloadContentBuilder
     {
         return payload == null ? null : new JsonPayloadMessageContent(payload, _jsonObjectSerializer, typeHint);
     }
-
-    public ObjectSerializer? ObjectSerializer => _jsonObjectSerializer;
 }
--- a/src/Microsoft.Azure.SignalR.Common/Utilities/Rest/JsonPayloadContentBuilder.cs
+++ b/src/Microsoft.Azure.SignalR.Common/Utilities/Rest/JsonPayloadContentBuilder.cs
@@ -24,4 +24,6 @@ internal class JsonPayloadContentBuilder : IPayloadContentBuilder
     {
         return payload == null ? null : new JsonPayloadMessageContent(payload, _jsonObjectSerializer, typeHint);
     }
+
+    public ObjectSerializer? ObjectSerializer => _jsonObjectSerializer;
 }
--- a/src/Microsoft.Azure.SignalR.Management/Microsoft.Azure.SignalR.Management.csproj
+++ b/src/Microsoft.Azure.SignalR.Management/Microsoft.Azure.SignalR.Management.csproj
@@ -44,6 +44,7 @@
 
   <ItemGroup Condition=" '$(TargetFrameworkIdentifier)' != '.NETStandard' ">
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
+    <PackageReference Include="Microsoft.AspNetCore.SignalR.Protocols.MessagePack" Version="8.0.11" />
   </ItemGroup>
 
 </Project>
--- a/src/Microsoft.Azure.SignalR.Management/Microsoft.Azure.SignalR.Management.csproj
+++ b/src/Microsoft.Azure.SignalR.Management/Microsoft.Azure.SignalR.Management.csproj
@@ -44,7 +44,6 @@
 
   <ItemGroup Condition=" '$(TargetFrameworkIdentifier)' != '.NETStandard' ">
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
-    <PackageReference Include="Microsoft.AspNetCore.SignalR.Protocols.MessagePack" Version="8.0.11" />
   </ItemGroup>
 
 </Project>
--- a/src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
+++ b/src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
@@ -16,6 +16,9 @@ using System.Threading.Tasks;
 using Azure;
 
 using Microsoft.AspNetCore.SignalR;
+#if NET7_0_OR_GREATER
+using Microsoft.AspNetCore.SignalR.Protocol;
+#endif
 using Microsoft.Extensions.Primitives;
 
 using static Microsoft.Azure.SignalR.Constants;
@@ -33,13 +36,15 @@ internal class RestHubLifetimeManager<THub> : HubLifetimeManager<THub>, IService
     private readonly RestApiProvider _restApiProvider;
     private readonly string _hubName;
     private readonly string _appName;
+    private readonly IHubProtocolResolver _protocolResolver;
 
-    public RestHubLifetimeManager(string hubName, ServiceEndpoint endpoint, string appName, RestClient restClient)
+    public RestHubLifetimeManager(string hubName, ServiceEndpoint endpoint, string appName, RestClient restClient, IHubProtocolResolver protocolResolver)
     {
         _restApiProvider = new RestApiProvider(endpoint);
         _appName = appName;
         _hubName = hubName;
         _restClient = restClient;
+        _protocolResolver = protocolResolver;
     }
 
     public override async Task AddToGroupAsync(string connectionId, string groupName, CancellationToken cancellationToken = default)
@@ -478,12 +483,26 @@ internal class RestHubLifetimeManager<THub> : HubLifetimeManager<THub>, IService
         throw new NotImplementedException();
     }
 
+    private static bool IsInvocationSupported(IHubProtocol protocol)
+    {
+        // Use protocol.Name to check for supported protocols
+        switch (protocol.Name)
+        {
+            case Constants.Protocol.Json:
+            case Constants.Protocol.MessagePack:
+                return true;
+            default:
+                return false;
+        }
+    }
+
+#endif
 
     private static bool FilterExpectedResponse(HttpResponseMessage response, string expectedErrorCode) =>
         response.IsSuccessStatusCode
         || (response.StatusCode == HttpStatusCode.NotFound && response.Headers.TryGetValues(Headers.MicrosoftErrorCode, out var errorCodes) && errorCodes.First().Equals(expectedErrorCode, StringComparison.OrdinalIgnoreCase));
 
-    public AsyncPageable<SignalRGroupConnection> ListConnectionsInGroup(string groupName, int? top = null, CancellationToken token = default)
+    public AsyncPageable<SignalRGroupMember> ListConnectionsInGroup(string groupName, int? top = null, CancellationToken token = default)
     {
         if (string.IsNullOrWhiteSpace(groupName))
         {
@@ -497,7 +516,7 @@ internal class RestHubLifetimeManager<THub> : HubLifetimeManager<THub>, IService
 
         return new PageableGroupMember(FetchPages, token);
 
-        async IAsyncEnumerable<Page<SignalRGroupConnection>> FetchPages(string? continuationToken, int? pageSizeHint)
+        async IAsyncEnumerable<Page<SignalRGroupMember>> FetchPages(string? continuationToken, int? pageSizeHint)
         {
             // Calculate the api for the first page
             var api = _restApiProvider.GetListConnectionsInGroupEndpoint(_appName, _hubName, groupName);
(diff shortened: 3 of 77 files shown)


src/Microsoft.Azure.SignalR.Management/Microsoft.Azure.SignalR.Management.csproj
```diff
@@ -44,6 +44,7 @@
 
   <ItemGroup Condition=" '$(TargetFrameworkIdentifier)' != '.NETStandard' ">
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
+    <PackageReference Include="Microsoft.AspNetCore.SignalR.Protocols.MessagePack" Version="8.0.11" />
```
Y-Sindo: Why do we need this?

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
```diff
@@ -359,10 +363,129 @@ public async Task SendStreamCompletionAsync(string connectionId, string streamId
...
+                    var protocol = _protocolResolver.AllProtocols
+                        .FirstOrDefault(p => string.Equals(p.Name, protocolName, StringComparison.OrdinalIgnoreCase));
+
+                    if (protocol == null)
+                    {
+                        if (string.Equals(protocolName, "messagepack", StringComparison.OrdinalIgnoreCase) &&
+                            _protocolResolver.AllProtocols.Count == 1 &&
+                            _protocolResolver.AllProtocols[0] is JsonObjectSerializerHubProtocol jsonObjectSerializerHubProtocol)
+                        {
+                            // The hub protocol is the default one. Service will convert it to MessagePack and keep backward compatibility as users may depend on this feature for MessagePack client support.
```
Y-Sindo: We shouldn't new a MessagePack hub protocol by ourselves.

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
```diff
@@ -359,10 +363,129 @@ public async Task SendStreamCompletionAsync(string connectionId, string streamId
...
+                        else
+                        {
+                            throw new NotSupportedException($"The protocol '{protocolName}' is not supported.");
+                        }
+                    }
+
+                    // 3. Read raw completion payload from response body
+
+                    byte[] buffer;
+                    await using var stream = await response.Content.ReadAsStreamAsync(cancellationToken)
```
Y-Sindo: Can we read from this stream directly instead of via a memory stream?

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
```diff
@@ -359,10 +363,129 @@ public async Task SendStreamCompletionAsync(string connectionId, string streamId
...
+                        throw new HubException("Response is missing protocol header.");
+                    }
+                    var protocolName = protocolHeaderValues.FirstOrDefault();
+                    if (string.IsNullOrEmpty(protocolName))
+                    {
+                        throw new HubException("Response protocol header is empty.");
+                    }
+
+                    // 2. Pick the hub protocol that matches X-Protocol
+                    var protocol = _protocolResolver.AllProtocols
```
Y-Sindo: Use `IHubProtocolResolver.GetProtocol` instead.

src/Microsoft.Azure.SignalR.Common/Utilities/Rest/JsonPayloadContentBuilder.cs
```diff
@@ -24,4 +24,6 @@ public JsonPayloadContentBuilder(ObjectSerializer jsonObjectSerializer)
     {
         return payload == null ? null : new JsonPayloadMessageContent(payload, _jsonObjectSerializer, typeHint);
     }
+
+    public ObjectSerializer? ObjectSerializer => _jsonObjectSerializer;
```
Y-Sindo: Useless?

test/Microsoft.Azure.SignalR.E2ETests/Management/ServiceHubContextE2EFacts.cs
```diff
@@ -1179,4 +1392,200 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
...
+    }
+
+    public enum TestEnum
+    {
+        None,
+        MethodInvoked,
+        aaamytest
+    }
+
+    public enum ClientInvocationTestIntentions
```
Y-Sindo: I think we have added too many layers to the tests for different cases. We have to look through many layers to understand what's the differences between each case. It's hard to check what's the input and expected output for each case. We'd better split them into different `[Fact]` or `[Theory]`, highlight the test setup and result verification in each fact or thoery.

src/Microsoft.Azure.SignalR.Management/Microsoft.Azure.SignalR.Management.csproj
ZSWY666: We will help convert the json invocation message to messagepack if the protocol is not set by user in runtime, therefore, we will need to initialize a MessagePackHubProtocol in SDK to deserialize the relative response
image

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
ZSWY666: same as another comment, maybe we can talk offline

test/Microsoft.Azure.SignalR.E2ETests/Management/ServiceHubContextE2EFacts.cs
ZSWY666: split to 7 tests

src/Microsoft.Azure.SignalR.Common/Utilities/Rest/JsonPayloadContentBuilder.cs
ZSWY666: removed

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
ZSWY666: refined

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
ZSWY666: refined

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
Y-Sindo: If the customer doesn't add a MessagePackHub protocol, then we are safe to say that the customer doesn't need MessagePack hub protocol. We can just throw exception if we cannot handle it.

ZSWY666: removed

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
```diff
@@ -359,10 +362,113 @@ public async Task SendStreamCompletionAsync(string connectionId, string streamId
...
+                    {
+                        throw new HubException("Response protocol header is empty.");
+                    }
+
+                    // 2. Pick the hub protocol that matches X-Protocol
+                    var protocol = _protocolResolver.GetProtocol(protocolName, supportedProtocols: null);
+
+                    if (protocol == null)
+                    {
+                        throw new NotSupportedException($"The protocol '{protocolName}' is not supported.");
```
Y-Sindo: NotSupportedException is used when we don't support the protocol at all. But here the situation is the users don't provide a proper protocol. Use `InvalidOperationException` and suggest users adding missing protocol with `ServiceManagerBuilder.AddHubProtocol()` or `ServiceManagerBuilder.WithHubProtocol`

test/Microsoft.Azure.SignalR.E2ETests/Management/ServiceHubContextE2EFacts.cs
```diff
@@ -971,6 +977,348 @@ public async Task ListConnectionsInGroupTest(ServiceTransportType serviceTranspo
...
+        var serviceManager = serviceManagerBuilder.WithLoggerFactory(loggerFactory).BuildServiceManager();
+        using var hubContext = await serviceManager.CreateHubContextAsync(HubName, default);
+
+        // Act & Assert: Test JSON protocol
+        await TestClientInvocationAsync(hubContext, "json", hasCustomisedSerializer: false);
+
+        // Act & Assert: Test MessagePack protocol (only in Transient mode where service auto-converts)
+        if (serviceTransportType == Management.ServiceTransportType.Transient)
+        {
+            await TestClientInvocationAsync(hubContext, "messagepack", hasCustomisedSerializer: false);
```
Y-Sindo: Although these are split into multiple test methods, most of the actual behavior is still hidden behind shared “core” helper `TestClientInvocationAsync` and controlled by passing different constants. This makes each test harder to read in isolation, and when a failure happens it’s not obvious which behavior/regression we’re validating. Could we refactor so each test is more explicit about the arrange/act/assert for its own scenario (even if it means a bit more duplication), and keep shared helpers limited to setup/teardown or truly common plumbing?

ZSWY666: refined, not sure if it is ok to introduce such tremendous dupe code though

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
ZSWY666: refined

test/Microsoft.Azure.SignalR.E2ETests/Management/ServiceHubContextE2EFacts.cs
```diff
@@ -971,6 +977,623 @@ public async Task ListConnectionsInGroupTest(ServiceTransportType serviceTranspo
...
+        {
+            // Act: Invoke method that returns a string
+            var stringResult = await hubContext.Clients.Client(clientConnection.ConnectionId)
+                .InvokeAsync<string>("InvokeString", "", default).OrTimeout();
+
+            // Assert: Verify string result
+            Assert.Equal("Method Invoked", stringResult);
+
+            // Act: Invoke method that returns an object
+            var objectResult = await hubContext.Clients.Client(clientConnection.ConnectionId)
```
Y-Sindo: I think you can return a type which contains string, enum, null so that one invocation is enough.

test/Microsoft.Azure.SignalR.E2ETests/Management/ServiceHubContextE2EFacts.cs
```diff
@@ -971,6 +977,623 @@ public async Task ListConnectionsInGroupTest(ServiceTransportType serviceTranspo
...
+                .InvokeAsync<testObject>("InvokeObject", "", default).OrTimeout();
+
+            // Assert: Verify object result with standard enum serialization
+            Assert.Equal("Method Invoked", objectResult.Name);
+            Assert.Equal("MethodInvoked", objectResult.EnumValue.ToString());
+
+            // Act: Invoke method that returns null
+            var nullResult = await hubContext.Clients.Client(clientConnection.ConnectionId)
```
Y-Sindo: And the input parameter should also be a complex object, to test the serialization of the input method.

test/Microsoft.Azure.SignalR.E2ETests/Management/ServiceHubContextE2EFacts.cs
Y-Sindo: And do we need to test the condition that the method doesn't have return, that is, the return is `void`?

test/Microsoft.Azure.SignalR.E2ETests/Management/ServiceHubContextE2EFacts.cs
```diff
@@ -971,6 +977,623 @@ public async Task ListConnectionsInGroupTest(ServiceTransportType serviceTranspo
...
+        await connection.StartAsync();
+        RegisterClientInvocationHandlers(connection);
+        return connection;
+    }
+
+    private static async Task<HubConnection> CreateMessagePackClientConnectionAsync(string endpoint, string accessToken)
+    {
+        var connection = new HubConnectionBuilder()
+            .WithUrl(endpoint, option => option.AccessTokenProvider = () => Task.FromResult(accessToken))
+            .WithAutomaticReconnect()
```
Y-Sindo, line 1418: This is not necessary.

test/Microsoft.Azure.SignalR.E2ETests/Management/ServiceHubContextE2EFacts.cs
ZSWY666: refined,.

We do not support that method doesn't have return in CSharp as they will have to idendify a return type while invoking client methods. Maybe we can talk about this offline

test/Microsoft.Azure.SignalR.E2ETests/Management/ServiceHubContextE2EFacts.cs
ZSWY666: refined

test/Microsoft.Azure.SignalR.E2ETests/Management/ServiceHubContextE2EFacts.cs
ZSWY666: removed

Y-Sindo (approved):

Y-Sindo:
This PR only support REST mode, right? Could we update the PR title to clarify it?

ZSWY666:
> This PR only support REST mode, right? Could we update the PR title to clarify it?

modified

PR #2267, opened 2026-06-02
refine response handle process to avoid OOM

ZSWY666:
 - Detail 1
 - Detail 2


src/Microsoft.Azure.SignalR.Common/Auth/MicrosoftEntra/MicrosoftEntraAccessKey.cs
vwxyzh, line 254: `httpClient.MaxResponseContentBufferSize = 1024 * 1024;`

PR #2283, opened 2026-07-09
Support Authentication Refresh for Azure SignalR (Serverless mode)

### LINKED ISSUE #2281: Feature: Azure SignalR Authentication Refresh for Serverless Mode
## Background

ASP.NET Core added `POST {hubUrl}/refresh?id={connectionToken}`, re-authenticates the request, and updates the connection's `ClaimsPrincipal` and expiration without reconnecting.

- https://github.com/Azure/azure-signalr/issues/2282 describes the shared client contract, service protocol, semantics, and error mapping.
- https://github.com/Azure/azure-signalr/issues/2278 covers the ASP.NET Core app-server using Azure SignalR service.
- This document covers **Serverless-mode**.

In Serverless mode, there is no hub server. The customer hosts a negotiation endpoint by using the **Azure SignalR Management SDK**. That host is the authentication boundary in front of ASRS. An *Azure Function App* is the most common host. The code samples use an Azure Functions host as the canonical example, but the `ServiceHubContext` calls are identical everywhere.

The client-facing `/refresh` endpoint is a sibling of the `/negotiate` endpoint. Negotiate now calls the new `ServiceHubContext.NegotiateWithTokenLifetimeAsync(...)`, which returns `tokenLifetimeSeconds` alongside the connection info so a refresh-aware client knows when to schedule its refresh (mirroring Default mode). The `/refresh` endpoint propagates the refresh request to ASRS through the Management SDK. Unlike Default mode, where the app-server orchestration is internal and there is no new `Microsoft.Azure.SignalR` public API, serverless needs **new public Management SDK APIs** (`ServiceHubContext.NegotiateWithTokenLifetimeAsync` for negotiate; `ServiceHubContext.RefreshConnectionAuthenticationAsync` + optional `ServiceHubContext.GetConnectionClaimsAsync` for refresh). The refresh APIs are transport-agnostic: they behave identically for `ServiceTransportType.Transient` (REST data plane) and `ServiceTransportType.Persistent` (service-protocol tunnel).

### 1. Application-facing surface (Management SDK host)

The refresh endpoint host validates the client's **app** token (EasyAuth / JWT bearer / cookie / custom — whatever it already uses for `/negotiate`), then calls the Management SDK. The HTTP endpoint, route, and auth policy stay **customer-owned**, exactly like the negotiate endpoint. The sample below is an Azure Functions host; the identical `ServiceHubContext` calls work in an ASP.NET Core minimal API or any other Management SDK host.

```csharp
public sealed class RefreshFunctions(ServiceHubContext hubContext, IAppTokenValidator tokens)
{
    // Mirrors negotiate: the client POSTs its refreshed app token; id = connection token.
    [Function("refresh")]
    public async Task<IActionResult> Refresh(
        [HttpTrigger(AuthorizationLevel.Anonymous, "post", Route = "refresh")] HttpRequest req)
    {
        var connectionToken = req.Query["id"].ToString();
        if (string.IsNullOrEmpty(connectionToken))
            return new BadRequestObjectResult(new { error = "missing_connection_token" });

        if (!tokens.TryValidate(req, out var principal))          // app-token auth
            return new UnauthorizedResult();                       // 401 invalid_token

        // Optional accept/reject gate.
        // Only fetch PreviousUser when the policy actually needs it.
        if (Policy.NeedsPreviousUser)
        {
            var previous = await hubContext.GetConnectionClaimsAsync(connectionToken); // read-only
            if (!Policy.Allow(previous.Claims, principal))
                return new ObjectResult(new { error = "permission_change_rejected" }) { StatusCode = 403 };
        }

        try
        {
            var result = await hubContext.RefreshConnectionAuthenticationAsync(
                connectionToken,
                new RefreshConnectionAuthenticationOptions
                {
                    AuthenticationExpiresOn = principal.GetExpiration(), // new auth deadline
                    Claims = principal.ProjectClaims(),                   // application claims to apply
                    TokenLifetime = configuredServiceTokenLifetime,       // refreshed service-token maximum
                });
            return new OkObjectResult(new
            {
                accessToken = result.AccessToken,                    // refreshed SERVICE token
                tokenLifetimeSeconds = result.TokenLifetimeSeconds,  // when to refresh next
            });
        }
        catch (AzureSignalRException ex)                             // typed failure -> HTTP mapping
        {
            return ex.ToRefreshErrorResult();                        // 403 / 404 / 400 / 500 (see Error Mapping)
        }
    }
}
```

The **.NET client is the stock SignalR client** (`WithAuthenticationRefresh`, `AccessTokenProvider`) pointed at the negotiation server's base URL. It needs no serverless-specific configuration and is identical to Default mode.

### 2. New Management SDK public API

Three new public APIs on `ServiceHubContext`: `NegotiateWithTokenLifetimeAsync` (negotiate advertising the token lifetime), plus `RefreshConnectionAuthenticationAsync` and `GetConnectionClaimsAsync` (keyed by the **connection token**).

#### Negotiate advertising `tokenLifetimeSeconds`

`NegotiateWithTokenLifetimeAsync` extends negotiate to also return the effective client access-token lifetime, so a refresh-aware client can schedule its refresh — the serverless equivalent of the Default-mode negotiate response carrying `tokenLifetimeSeconds`. The host supplies the maximum service-token lifetime through `TokenLifetime` and the application-authentication deadline through `AuthenticationExpiresOn`.

```diff
namespace Microsoft.Azure.SignalR.Management;

+public partial class NegotiationOptions
+{
+    // Absolute application-authentication deadline. Null preserves the legacy CloseOnAuthenticationExpiration fallback; MaxValue explicitly means no deadline.
+    public System.DateTimeOffset? AuthenticationExpiresOn { get; set; }
+}

 public abstract partial class ServiceHubContext : IHubContext<Hub>, IServiceHubContext, IDisposable
 {
+    // Negotiates and additionally returns the client access token lifetime
+    // used to mint the service token, so a refresh-aware client can schedule its refresh.
+    public virtual Task<NegotiationResult> NegotiateWithTokenLifetimeAsync(
+        NegotiationOptions negotiationOptions = null,
+        System.Threading.CancellationToken cancellationToken = default);
 }

+// The result of a NegotiateWithTokenLifetimeAsync call: the client connection info plus the token lifetime.
+public sealed partial class NegotiationResult
+{
+    internal NegotiationResult() { }
+    public string Url { get; }                 // endpoint URL for the client to connect to ASRS
+    public string AccessToken { get; }         // initial service access token
+    public int TokenLifetimeSeconds { get; }   // when the client should schedule its first refresh
+}
```

A negotiate host echoes the lifetime so the stock SignalR client schedules its refresh, mirroring Default mode:

```csharp
var negotiate = await hubContext.NegotiateWithTokenLifetimeAsync(
    new NegotiationOptions
    {
        UserId = userId,
        Claims = claims,
        TokenLifetime = configuredServiceTokenLifetime,
        AuthenticationExpiresOn = applicationAuthenticationExpiresOn,
        CloseOnAuthenticationExpiration = true,
    });
return new OkObjectResult(new
{
    url = negotiate.Url,
    accessToken = negotiate.AccessToken,
    tokenLifetimeSeconds = negotiate.TokenLifetimeSeconds,   // client schedules its refresh against this
});
```

In Serverless mode, both `NegotiateAsync` and `NegotiateWithTokenLifetimeAsync` use `min(TokenLifetime, AuthenticationExpiresOn - now)` for the actual service-token JWT lifetime whenever a real authentication deadline exists. `NegotiateWithTokenLifetimeAsync` additionally returns that effective lifetime as `tokenLifetimeSeconds`; the caller uses that value only when it exposes a refresh endpoint to its clients. The SDK preserves `AuthenticationExpiresOn` itself as the connection's close-on-authentication-expiration deadline. For backward compatibility, when `AuthenticationExpiresOn` is null and `CloseOnAuthenticationExpiration` is true, the deadline defaults to `now + TokenLifetime`. `DateTimeOffset.MaxValue` explicitly means no authentication deadline. A host that does not expose refresh continues to use `NegotiateAsync`, which applies the same JWT lifetime cap but does not return `tokenLifetimeSeconds`.

`MaximumAuthenticationExpiration` is an ASP.NET Core SignalR `HttpConnectionDispatcherOptions` setting used by Default mode. It is not added to `NegotiationOptions`: Serverless hosts provide their effective absolute deadline explicitly through `AuthenticationExpiresOn`.

#### Refresh and get claims

`RefreshConnectionAuthenticationAsync` and `GetConnectionClaimsAsync` keyed by the **connection token**.

```diff
namespace Microsoft.Azure.SignalR.Management;

 public abstract partial class ServiceHubContext : IHubContext<Hub>, IServiceHubContext, IDisposable
 {
+    // Applies a refresh to a live connection and returns the refreshed SERVICE access token
+    // plus the next tokenLifetimeSeconds. Works for both Transient and Persistent transports.
+    public virtual Task<RefreshConnectionAuthenticationResult> RefreshConnectionAuthenticationAsync(
+        string connectionToken,
+        RefreshConnectionAuthenticationOptions options = null,
+        System.Threading.CancellationToken cancellationToken = default);

+    // Optional read-only fetch of the connection's current claims.
+    public virtual Task<ConnectionClaims> GetConnectionClaimsAsync(
+        string connectionToken,
+        System.Threading.CancellationToken cancellationToken = default);
 }

+public sealed class RefreshConnectionAuthenticationOptions
+{
+    // New application-authentication deadline. Null preserves the existing deadline.
+    public System.DateTimeOffset? AuthenticationExpiresOn { get; set; }
+    // Projected application claims. Null preserves the existing claims.
+    public System.Collections.Generic.IEnumerable<System.Security.Claims.Claim> Claims { get; set; }
+    // Maximum lifetime for the newly minted service token. Defaults to one hour.
+    public System.TimeSpan TokenLifetime { get; set; } = System.TimeSpan.FromHours(1);
+}

+// The result of a successful RefreshConnectionAuthenticationAsync call.
+public sealed partial class RefreshConnectionAuthenticationResult
+{
+    internal RefreshConnectionAuthenticationResult() { }
+    // Refreshed service access token to return to the client.
+    public string AccessToken { get; }
+    // Seconds until the client should refresh again. Limited by the effective authentication deadline.
+    public int TokenLifetimeSeconds { get; }
+}

+// The result of a GetConnectionClaimsAsync call.
+public sealed partial class ConnectionClaims
+{
+    internal ConnectionClaims() { }
+    // The connection's current claim set, usable as PreviousUser for an accept/reject policy.
+    public System.Collections.Generic.IReadOnlyList<System.Security.Claims.Claim> Claims { get; }
+}
```

- `connectionToken` — the `id` from the customer-facing endpoint (the **connection token**, negotiate v1; v0 is rejected). ASRS resolves it to the connection id internally.
- `options.AuthenticationExpiresOn` — the new authentication expiration deadline, derived from the validated app token and normalized to UTC by the SDK. When null, the existing authentication expiration is preserved.
- `options.Claims` — the projected application claim set; applied by ASRS **after** the same-user check passes. When null, the existing claims are preserved.
- `options.TokenLifetime` — the maximum lifetime for the newly minted service token. It defaults to one hour and must be positive. Because Serverless Functions are stateless, the refresh caller supplies this value on every refresh; the earlier negotiate options are not retained.

##### `RefreshConnectionAuthenticationOptions` contract

`RefreshConnectionAuthenticationOptions` is a new public sealed Management SDK type in both the `netstandard2.0` and `net8.0` API surfaces. Its parameterless constructor produces a service-token-renewal operation: `AuthenticationExpiresOn` and `Claims` are null, while `TokenLifetime` is one hour. Passing `options: null` is equivalent to constructing this default instance.

| Property | Default | Validation and behavior |
| --- | --- | --- |
| `AuthenticationExpiresOn` | `null` | Null preserves the live connection's current authentication deadline. A supplied value is converted to UTC and must be later than `DateTimeOffset.UtcNow`; otherwise the SDK throws `ArgumentOutOfRangeException` before contacting ASRS. `DateTimeOffset.MaxValue` is accepted as the internal preserve-deadline sentinel. |
| `Claims` | `null` | Null preserves the current application claims. A supplied sequence is projected onto the live connection only after runtime same-user validation; reserved `asrs.s.*` state remains service-owned. |
| `TokenLifetime` | One hour | Must be greater than zero or the SDK throws `ArgumentOutOfRangeException` before contacting ASRS. This is a maximum for the newly minted service token, not a new connection deadline. |

The SDK always mints from the runtime-returned post-refresh claim set. In particular, claims-only and renewal-only calls use the returned effective `asrs.s.aeo` value to cap both the JWT lifetime and `TokenLifetimeSeconds`; they do not assume that an omitted request deadline means no deadline. An `Ok` response with a null or empty post-refresh claim set is rejected with `AzureSignalRException`, because minting without the preserved service-owned claims would produce an invalid token.

The API supports four refresh combinations:

| Call | Effect |
| --- | --- |
| `RefreshConnectionAuthenticationAsync(connectionToken, new() { AuthenticationExpiresOn = expiration, Claims = claims, TokenLifetime = lifetime })` | **Expiration + claims:** updates both and returns a service token capped by `lifetime`. |
| `RefreshConnectionAuthenticationAsync(connectionToken, new() { AuthenticationExpiresOn = expiration, TokenLifetime = lifetime })` | **Expiration-only refresh:** advances the authentication deadline and preserves the existing claims. |
| `RefreshConnectionAuthenticationAsync(connectionToken, new() { Claims = claims, TokenLifetime = lifetime })` | **Claims-only refresh:** applies the projected claims and preserves the existing authentication deadline. |
| `RefreshConnectionAuthenticationAsync(connectionToken, new() { TokenLifetime = lifetime })` | **Service-token renewal only:** preserves both the existing claims and authentication deadline, and returns a newly minted service access token. |

Passing null options is equivalent to default options: preserve claims and deadline and renew the service token with the one-hour default maximum. For claims-only and service-token-renewal-only calls, the Management SDK maps the omitted expiration to `DateTimeOffset.MaxValue` before dispatch. `DateTimeOffset.MaxValue` is the internal wire sentinel for "preserve the existing expiration": ASRS does not advance or clear the connection's close-on-authentication-expiration deadline.

`GetConnectionClaimsAsync` mirrors the Default-mode read path: the host whose policy needs `PreviousUser` calls it before `RefreshConnectionAuthenticationAsync`; an expiration-only policy skips it entirely.

### 3. New REST data-plane APIs

If you're using a language other than .NET, you can also manually invoke the REST APIs

```http
POST /api/hubs/{hub}/connections/:refreshAuth?api-version=2026-07-01   (write; AAD -> clientConnection/Write)
Authorization: Bearer {service-credential}
Content-Type: application/json
{ "connectionToken": "<token>", "expireTime": "2026-07-01T12:00:00Z", "claims"?: [ { "type": "...", "value": "..." } ] }

Responses:
| Status | Body | Meaning |
| --- | --- | --- |
| `200 OK` | `{ "claims": [ { "type": ..., "value": ... } ] }` | Refresh applied. Returns the connection's post-refresh claim set; the Management SDK mints the refreshed access token from it and computes `tokenLifetimeSeconds` locally. |
| `400 Bad Request` | `ErrorDetail` | Validation failure (missing `connectionToken` or `expireTime`, invalid `connectionToken`, malformed body). Set `expireTime` to `DateTimeOffset.MaxValue` to preserve the live connection's current authentication deadline. |
| `404 Not Found` | `ErrorDetail` | `ConnectionNotFound` — unknown/disconnected/negotiate v0 connection. |
| `403 Forbidden` | `ErrorDetail` | `ConnectionUserMismatch` — connection claims resolve to a different user. |
| `5xx` (default) | `ErrorDetail` | Internal server error. |


POST /api/hubs/{hub}/connections/claims?api-version=2026-07-01     (read; AAD -> clientConnection/Read)
Authorization: Bearer {service-credential}
Content-Type: application/json
{ "connectionToken": "<token>" }

Responses:
| Status | Body | Meaning |
| --- | --- | --- |
| `200 OK` | `{ "claims": [ { "type": ..., "value": ... } ] }` | Returns the current claim set. |
| `400 Bad Request` | `ErrorDetail` | Missing or invalid `connectionToken`, malformed request. |
| `404 Not Found` | `ErrorDetail` | `ConnectionNotFound` — unknown/disconnected/negotiate v0 connection. |
```

### 4. Internal mechanics

- **Transport-agnostic dispatch.** `RefreshConnectionAuthenticationAsync` / `GetConnectionClaimsAsync` are added to the internal `IServiceHubLifetimeManager`; `RestHubLifetimeManager` (Transient) implements the REST calls above, `WebsocketsHubLifetimeManager` (Persistent) sends the token-keyed `RefreshAuthMessage` (type 41) / `GetConnectionClaimsMessage` (type 42) over the existing service connection — the same service protocols the Default-mode SDK uses. Customers see one API; the manager picks the path from the configured transport.
- **Persistent multi-endpoint ownership.** When the owning endpoint is unknown, `MultiEndpointServiceConnectionContainer` fans claims lookup or refresh out to all online endpoints. A single non-`NotFound`/non-`Timeout` response identifies the owner; multiple resolving endpoints fail with `InternalServerError`. A successful claims lookup carries the owner internally so the immediately following refresh can pin to it. If that endpoint goes offline, refresh falls back to fan-out. Management SDK multi-endpoint operation requires `ServiceTransportType.Persistent`; Transient multi-endpoint is not implemented.
- **Local minting.** After ASRS confirms the refresh and returns the post-refresh claim set, `ServiceHubContextImpl` signs it with the **owning endpoint's** provider (`ServiceEndpointProvider.GenerateClientAccessTokenAsync`); it never re-derives system routing inputs from the request. The runtime-returned effective `asrs.s.aeo` deadline takes precedence over the request deadline, and the actual refreshed JWT lifetime matches `tokenLifetimeSeconds`: the shorter of `RefreshConnectionAuthenticationOptions.TokenLifetime` and remaining authentication lifetime. If no authentication deadline exists, the requested token lifetime is used.
- **No routed push.** Serverless has no owning app server, so there is no `UpdateConnectionClaimsMessage` (type 43). Instead the runtime stores the refreshed claims on `connection.User` and they propagate **lazily** on the next client → upstream payload.

## Behavior and Semantics (Serverless-mode specifics)

- The shared client `refresh` contract, identity/claims rules, token scheduling, and concurrency/multi-endpoint sharding are in https://github.com/Azure/azure-signalr/issues/2282.

Serverless mode adds:

### The application gate (Serverless `OnAuthenticationRefresh` equivalent)

- Runs in the negotiation server (the only host with the client's request context and app token), before ASRS mutates anything. Reject → `403 permission_change_rejected`, nothing changes.
- `PreviousUser` is fetched read-only via `GetConnectionClaimsAsync`; `NewUser` is the validated app principal. Fetch it only when the policy needs it, an expiration-only policy skips the read and calls `RefreshConnectionAuthenticationAsync` directly.

### Claims propagation (lazy upstream)

Serverless has no owning app server, so there is no routed push. Refresh triggers **no** proactive upstream call: the runtime stores the refreshed claims on `connection.User`, and the **next** client → upstream payload (a `messages`/invocation event, plus `connected`/`disconnected`) is built from that stored set, carrying the refreshed `X-ASRS-User-Id` / `X-ASRS-User-Claims` / CloudEvents user fields. This aligns with when upstream actually authorizes a connection (only when it sends a message), so an idle/receive-only client never needs to propagate. ASRS emits no proactive `refreshed` upstream event.

### Azure Functions SignalR Service Bindings

Once the Management SDK ships `RefreshConnectionAuthenticationAsync`, adds (a) `tokenLifetimeSeconds` to the `SignalRConnectionInfo` negotiate binding and (b) a `[SignalRRefresh]` input binding wrapping `RefreshConnectionAuthenticationAsync` and `GetConnectionClaimsAsync`, so JS/Python/Java/PowerShell Function Apps get serverless refresh declaratively. Gated on Management SDK GA.

## Flow

```mermaid
sequenceDiagram
    autonumber
    participant Client
    participant Func as Mgmt SDK host
    participant ASRS as Azure SignalR Service
    participant Upstream as Customer Upstream

    Client->>Func: POST /negotiate
    Func->>Func: NegotiateWithTokenLifetimeAsync -> { url, accessToken, tokenLifetimeSeconds }
    Func-->>Client: { url, accessToken, tokenLifetimeSeconds }
    Client->>ASRS: WebSocket / SSE / LongPolling connect with initial service access token
    ASRS->>ASRS: Store AuthenticationExpiresOn, user claims

    Note over Client,ASRS: tokenLifetimeSeconds about to elapse (client schedules refresh)

    Client->>Func: POST /refresh?hub={hub}&id={connectionToken}<br/>Authorization: Bearer {new-app-token}
    Func->>Func: Validate new app token
    opt Gate needs PreviousUser
      Func->>ASRS: GetConnectionClaimsAsync(connectionToken)<br/>Transient 'POST /connections/claims' / Persistent GetConnectionClaimsMessage (read-only)
      ASRS-->>Func: { claims } or failure
      Func->>Func: Apply accept/reject policy
    end
    Func->>Func: Derive ExpireTime + project claims
    Func->>ASRS: RefreshConnectionAuthenticationAsync(connectionToken, options)

    alt Transient (REST data plane)
      Func->>ASRS: POST /connections/:refreshAuth { connectionToken, expireTime, claims }
    else Persistent (service-protocol tunnel)
      Func->>ASRS: RefreshAuthMessage(ConnectionToken, Claims, ExpireTime, AckId)
    end

    ASRS->>ASRS: Resolve token to id, reject v0, forward to owning pod
    ASRS->>ASRS: Verify same user, update claims + AuthenticationExpiresOn

    alt Refresh applied
      ASRS-->>Func: 200 { claims } (Transient) or Ack(Ok, post-refresh Claims) (Persistent)
      Func->>Func: Mint accessToken from returned Claims for the owning endpoint
      Func-->>Client: 200 OK { accessToken, tokenLifetimeSeconds }
      Note over ASRS,Upstream: Refresh triggers no upstream call, claims ride the next client message
      ASRS->>Upstream: next client message carries refreshed<br/>X-ASRS-User-Id / X-ASRS-User-Claims / CloudEvents user fields
    else Refresh failed
      ASRS-->>Func: 4xx/5xx ErrorDetail or Ack(failure, reason)
      Func-->>Client: 401 / 403 / 404 / 500 with typed error
    end

    Note over Client,ASRS: Existing connection stays open
```

MoChilia:
## Summary
Implements the Serverless-mode (Management SDK) side of #2281. See the issue for the full background, API design, and behavior/semantics.

## What this PR does

Lets a negotiation server (an Azure Function App or any Management SDK host) refresh an Azure SignalR client connection's auth expiration and application claims without reconnecting, mirroring Default mode. Unlike Default mode — which reuses the app's existing `HttpConnectionDispatcherOptions` and adds no public API — serverless has no app server, so this PR introduces **new public Management SDK API** (`ServiceHubContext.RefreshAuthAsync` +`ServiceHubContext.GetConnectionClaimsAsync`) that the host calls. The API is transport-agnostic: it behaves identically for `ServiceTransportType.Transient` (REST data plane) and `ServiceTransportType.Persistent` (service-protocol tunnel).

## Changes

The API is public and multi-targets `netstandard2.0` / `net8.0` (no net11 guard needed). The Persistent path reuses the service protocol messages `RefreshAuthMessage` (41) / `GetConnectionClaimsMessage` (42) already released in v1.33.1.

- **New public API** — `ServiceHubContext.RefreshAuthAsync(connectionToken, expireTime, claims?)` and `GetConnectionClaimsAsync(connectionToken)`, plus result types `RefreshAuthResult { AccessToken, TokenLifetimeSeconds }` and `ConnectionClaimsResult { Claims }`. Keyed by the connection token; `claims` null/empty ⇒ expiration-only refresh.
- **`IServiceHubLifetimeManager`** — adds `RefreshAuthAsync` / `GetConnectionClaimsAsync`.
- **Transient (`RestHubLifetimeManager`)** — calls the new REST data-plane endpoints: `POST /connections/:refreshAuth` (write) and `QUERY /connections/claims` (read); both return `{ claims }` on success. Adds `RestApiProvider.RefreshApiVersion` + `GetRefreshConnectionAuthEndpoint` / `GetConnectionClaimsEndpoint`, and a `RestClient` overload that threads an `HttpContent` body.
- **Persistent (`WebsocketsHubLifetimeManager`)** — sends the token-keyed `RefreshAuthMessage` / `GetConnectionClaimsMessage` over the existing service connection.
- **Token minting (`ServiceHubContextImpl`)** — mints the refreshed service access token locally from the runtime-returned post-refresh claims for the connection's owning endpoint; `TokenLifetimeSeconds = min(AccessTokenLifetime, expireTime − now)`.

## Tests

- `RestHubLifetimeManagerFacts` — REST `:refreshAuth` / `claims` URL, body, and status mapping; argument validation.
- `WebsocketsHubLifetimeManagerFacts` — Persistent delegates the message to the connection container; argument validation.
- `ServiceHubContextImplFacts` — minting happy path from mocked endpoint/provider; `ThrowOnNonSuccess` status mapping; `Ok` + null-claims and no-endpoint throw.

### DIFF
--- a/src/Microsoft.Azure.SignalR.Management/RestApiProvider.cs
+++ b/src/Microsoft.Azure.SignalR.Management/RestApiProvider.cs
@@ -14,6 +14,8 @@ internal class RestApiProvider
     internal const string Version = "2024-12-01";
     public const string HealthApiPath = $"api/health?api-version={Version}";
 
+    internal const string RefreshApiVersion = "2026-07-01";
+
     private readonly string _serverEndpoint;
 
     public RestApiProvider(ServiceEndpoint endpoint)
@@ -110,12 +112,22 @@ internal class RestApiProvider
         return GenerateRestApiEndpoint(appName, hubName, $"/connections/{Uri.EscapeDataString(connectionId)}/:invoke");
     }
 
-    private RestApiEndpoint GenerateRestApiEndpoint(string appName, string hubName, string pathAfterHub, IDictionary<string, StringValues> queries = null)
+    public RestApiEndpoint GetRefreshConnectionAuthEndpoint(string appName, string hubName)
+    {
+        return GenerateRestApiEndpoint(appName, hubName, "/connections/:refreshAuth", apiVersion: RefreshApiVersion);
+    }
+
+    public RestApiEndpoint GetConnectionClaimsEndpoint(string appName, string hubName)
+    {
+        return GenerateRestApiEndpoint(appName, hubName, "/connections/:getClaims", apiVersion: RefreshApiVersion);
+    }
+
+    private RestApiEndpoint GenerateRestApiEndpoint(string appName, string hubName, string pathAfterHub, IDictionary<string, StringValues> queries = null, string apiVersion = Version)
     {
         var requestPrefixWithHub = $"{_serverEndpoint}api/hubs/{Uri.EscapeDataString(hubName.ToLowerInvariant())}";
         pathAfterHub = string.IsNullOrEmpty(appName)
-            ? $"{pathAfterHub}?api-version={Version}"
-            : $"{pathAfterHub}?application={Uri.EscapeDataString(appName.ToLowerInvariant())}&api-version={Version}";
+            ? $"{pathAfterHub}?api-version={apiVersion}"
+            : $"{pathAfterHub}?application={Uri.EscapeDataString(appName.ToLowerInvariant())}&api-version={apiVersion}";
         return new RestApiEndpoint($"{requestPrefixWithHub}{pathAfterHub}") { Query = queries };
     }
 }
--- a/src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
+++ b/src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
@@ -35,6 +35,9 @@ internal class RestHubLifetimeManager<THub> : HubLifetimeManager<THub>, IService
     private const string NullOrEmptyStringErrorMessage = "Argument cannot be null or empty.";
     private const string TtlOutOfRangeErrorMessage = "Ttl cannot be less than 0.";
 
+    // HttpMethod.Query only exists on .NET 10+, so use a shared instance for the netstandard2.0/net8.0 targets.
+    private static readonly HttpMethod QueryMethod = new("QUERY");
+
     private readonly RestClient _restClient;
     private readonly RestApiProvider _restApiProvider;
     private readonly string _hubName;
@@ -349,7 +352,7 @@ internal class RestHubLifetimeManager<THub> : HubLifetimeManager<THub>, IService
         var status = AckStatus.InternalServerError;
         IReadOnlyList<Claim>? resultClaims = null;
         using var content = CreateJsonContent(requestBody);
-        await _restClient.SendWithRetryAsync(api, HttpMethod.Post, content, async response =>
+        await _restClient.SendWithRetryAsync(api, QueryMethod, content, async response =>
         {
             switch (response.StatusCode)
             {
--- a/src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
+++ b/src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
@@ -8,7 +8,9 @@ using System.Linq;
 using System.Net;
 using System.Net.Http;
 
+using System.Security.Claims;
 using System.Text.Json;
+using System.Text.Json.Serialization;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -292,6 +294,145 @@ internal class RestHubLifetimeManager<THub> : HubLifetimeManager<THub>, IService
             return FilterExpectedResponse(response, ErrorCodes.WarningConnectionNotExisted);
         }, cancellationToken: cancellationToken);
         return exists;
+    }
+
+    public async Task<RefreshConnectionAuthResult> RefreshAuthAsync(string connectionToken, DateTimeOffset expireTime, IEnumerable<Claim>? claims, CancellationToken cancellationToken)
+    {
+        if (string.IsNullOrEmpty(connectionToken))
+        {
+            throw new ArgumentException(NullOrEmptyStringErrorMessage, nameof(connectionToken));
+        }
+
+        var api = _restApiProvider.GetRefreshConnectionAuthEndpoint(_appName, _hubName);
+        var requestBody = new RefreshConnectionAuthRequestBody
+        {
+            ConnectionToken = connectionToken,
+            ExpireTime = expireTime,
+            Claims = ToClaimDtos(claims),
+        };
+
+        var status = AckStatus.InternalServerError;
+        IReadOnlyList<Claim>? resultClaims = null;
+        using var content = CreateJsonContent(requestBody);
+        await _restClient.SendWithRetryAsync(api, HttpMethod.Post, content, async response =>
+        {
+            switch (response.StatusCode)
+            {
+                case HttpStatusCode.OK:
+                    status = AckStatus.Ok;
+                    resultClaims = await ReadClaimsAsync(response);
+                    return true;
+                case HttpStatusCode.Forbidden:
+                    status = AckStatus.Forbidden;
+                    return true;
+                case HttpStatusCode.NotFound:
+                    status = AckStatus.NotFound;
+                    return true;
+                default:
+                    return false;
+            }
+        }, cancellationToken: cancellationToken);
+
+        return new RefreshConnectionAuthResult(status, resultClaims);
+    }
+
+    public async Task<GetConnectionClaimsResult> GetConnectionClaimsAsync(string connectionToken, CancellationToken cancellationToken)
+    {
+        if (string.IsNullOrEmpty(connectionToken))
+        {
+            throw new ArgumentException(NullOrEmptyStringErrorMessage, nameof(connectionToken));
+        }
+
+        var api = _restApiProvider.GetConnectionClaimsEndpoint(_appName, _hubName);
+        var requestBody = new GetConnectionClaimsRequestBody { ConnectionToken = connectionToken };
+
+        var status = AckStatus.InternalServerError;
+        IReadOnlyList<Claim>? resultClaims = null;
+        using var content = CreateJsonContent(requestBody);
+        await _restClient.SendWithRetryAsync(api, HttpMethod.Post, content, async response =>
+        {
+            switch (response.StatusCode)
+            {
+                case HttpStatusCode.OK:
+                    status = AckStatus.Ok;
+                    resultClaims = await ReadClaimsAsync(response);
+                    return true;
+                case HttpStatusCode.NotFound:
+                    status = AckStatus.NotFound;
+                    return true;
+                default:
+                    return false;
+            }
+        }, cancellationToken: cancellationToken);
+
+        return new GetConnectionClaimsResult(status, resultClaims);
+    }
+
+    private static HttpContent CreateJsonContent<T>(T body) =>
+        new StringContent(JsonSerializer.Serialize(body), System.Text.Encoding.UTF8, "application/json");
+
+    private static async Task<IReadOnlyList<Claim>> ReadClaimsAsync(HttpResponseMessage response)
+    {
+        using var stream = await response.Content.ReadAsStreamAsync();
+        var body = await JsonSerializer.DeserializeAsync<ClaimsResponseBody>(stream);
+        var dtos = body?.Claims;
+        if (dtos == null || dtos.Length == 0)
+        {
+            return Array.Empty<Claim>();
+        }
+        var result = new Claim[dtos.Length];
+        for (var i = 0; i < dtos.Length; i++)
+        {
+            result[i] = new Claim(dtos[i].Type, dtos[i].Value);
+        }
+        return result;
+    }
+
+    private static ClaimDto[]? ToClaimDtos(IEnumerable<Claim>? claims)
+    {
+        if (claims == null)
+        {
+            return null;
+        }
+        var list = new List<ClaimDto>();
+        foreach (var claim in claims)
+        {
+            list.Add(new ClaimDto { Type = claim.Type, Value = claim.Value });
+        }
+        return list.Count == 0 ? null : list.ToArray();
+    }
+
+    private sealed class RefreshConnectionAuthRequestBody
+    {
+        [JsonPropertyName("connectionToken")]
+        public string ConnectionToken { get; set; } = string.Empty;
+
+        [JsonPropertyName("expireTime")]
+        public DateTimeOffset ExpireTime { get; set; }
+
+        [JsonPropertyName("claims")]
+        public ClaimDto[]? Claims { get; set; }
+    }
+
+    private sealed class GetConnectionClaimsRequestBody
+    {
+        [JsonPropertyName("connectionToken")]
+        public string ConnectionToken { get; set; } = string.Empty;
+    }
+
+    private sealed class ClaimsResponseBody
+    {
+        [JsonPropertyName("claims")]
+        public ClaimDto[]? Claims { get; set; }
+    }
+
+    private sealed class ClaimDto
+    {
+        [JsonPropertyName("type")]
+        public string Type { get; set; } = string.Empty;
+
+        [JsonPropertyName("value")]
+        public string Value { get; set; } = string.Empty;
     }
 
     public async Task<bool> UserExistsAsync(string userId, CancellationToken cancellationToken = default)
--- a/src/Microsoft.Azure.SignalR.Management/ServiceHubContext.cs
+++ b/src/Microsoft.Azure.SignalR.Management/ServiceHubContext.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel;
+using System.Security.Claims;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -37,6 +38,23 @@ namespace Microsoft.Azure.SignalR.Management
         /// </summary>
         /// <returns>A negotiation response object that contains an endpoint url and an access token for the client to connect to the Azure SignalR instance. </returns>
         public virtual ValueTask<NegotiationResponse> NegotiateAsync(NegotiationOptions negotiationOptions = null, CancellationToken cancellationToken = default) => throw new NotImplementedException();
+
+        /// <summary>
+        /// Refreshes the authentication expiration and the application claims of a live client connection without reconnecting,
+        /// then returns a refreshed service access token for the client.
+        /// </summary>
+        /// <param name="connectionToken">The connection token (the <c>id</c> from the client-facing refresh endpoint).</param>
+        /// <param name="expireTime">The new authentication expiration deadline (UTC).</param>
+        /// <param name="claims">The projected application claim set; applied after the same-user check passes. When null, the refresh is expiration-only.</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        public virtual Task<RefreshAuthResult> RefreshAuthAsync(string connectionToken, DateTimeOffset expireTime, IEnumerable<Claim> claims = null, CancellationToken cancellationToken = default) => throw new NotImplementedException();
+
+        /// <summary>
+        /// Reads the current claim set of a live client connection. 
+        /// </summary>
+        /// <param name="connectionToken">The connection token (the <c>id</c> from the client-facing refresh endpoint).</param>
+        /// <param name="cancellationToken">The cancellation token.</param>
+        public virtual Task<ConnectionClaimsResult> GetConnectionClaimsAsync(string connectionToken, CancellationToken cancellationToken = default) => throw new NotImplementedException();
 
         public virtual Task DisposeAsync() => Task.CompletedTask;
 
--- a/src/Microsoft.Azure.SignalR.Management/ServiceHubContextImpl.cs
+++ b/src/Microsoft.Azure.SignalR.Management/ServiceHubContextImpl.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Security.Claims;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -22,6 +23,7 @@ namespace Microsoft.Azure.SignalR.Management
     {
         private readonly string _hubName;
         private readonly IHubContext<Hub> _hubContext;
+        private readonly IServiceHubLifetimeManager _lifetimeManager;
         private readonly NegotiateProcessor _negotiateProcessor;
         private readonly IServiceEndpointManager _endpointManager;
 
@@ -49,11 +51,74 @@ namespace Microsoft.Azure.SignalR.Management
             ServiceProvider = serviceProvider;
             _negotiateProcessor = negotiateProcessor;
             _endpointManager = endpointManager;
+            _lifetimeManager = lifetimeManager;
         }
 
         public override ValueTask<NegotiationResponse> NegotiateAsync(NegotiationOptions options, CancellationToken cancellationToken)
         {
             return new ValueTask<NegotiationResponse>(_negotiateProcessor.NegotiateAsync(_hubName, options, cancellationToken));
+        }
+
+        public override async Task<RefreshAuthResult> RefreshAuthAsync(string connectionToken, DateTimeOffset expireTime, IEnumerable<Claim> claims = null, CancellationToken cancellationToken = default)
+        {
+            if (string.IsNullOrEmpty(connectionToken))
+            {
+                throw new ArgumentException("Argument cannot be null or empty.", nameof(connectionToken));
+            }
+
+            var result = await _lifetimeManager.RefreshAuthAsync(connectionToken, expireTime, claims, cancellationToken);
+            ThrowOnNonSuccess(result.Status);
+
+            if (result.Claims == null)
+            {
+                throw new AzureSignalRException("The service did not return the post-refresh claim set required to mint the refreshed access token.");
+            }
+
+            var accessTokenLifetime = Constants.Periods.DefaultAccessTokenLifetime;
+            ServiceEndpoint owningEndpoint = result.OwningEndpoint;
+            if (owningEndpoint == null)
+            {
+                var endpoints = _endpointManager.GetEndpoints(_hubName);
+                if (endpoints.Count == 0)
+                {
+                    throw new AzureSignalRException("No service endpoint is available to mint the refreshed access token.");
+                }
+                owningEndpoint = endpoints[0];
+            }
+            var provider = _endpointManager.GetEndpointProvider(owningEndpoint);
+            var accessToken = await provider.GenerateClientAccessTokenAsync(_hubName, result.Claims, accessTokenLifetime);
+            var remainingSeconds = (expireTime - DateTimeOffset.UtcNow).TotalSeconds;
+            var tokenLifetimeSeconds = (int)Math.Max(0, Math.Min(accessTokenLifetime.TotalSeconds, remainingSeconds));
+            return new RefreshAuthResult(accessToken, tokenLifetimeSeconds);
+        }
+
+        public override async Task<ConnectionClaimsResult> GetConnectionClaimsAsync(string connectionToken, CancellationToken cancellationToken = default)
+        {
+            if (string.IsNullOrEmpty(connectionToken))
+            {
+                throw new ArgumentException("Argument cannot be null or empty.", nameof(connectionToken));
+            }
+
+            var result = await _lifetimeManager.GetConnectionClaimsAsync(connectionToken, cancellationToken);
+            ThrowOnNonSuccess(result.Status);
+            return new ConnectionClaimsResult(result.Claims ?? Array.Empty<Claim>());
+        }
+
+        private static void ThrowOnNonSuccess(AckStatus status)
+        {
+            switch (status)
+            {
+                case AckStatus.Ok:
+                    return;
+                case AckStatus.Forbidden:
+                    throw new AzureSignalRException("The refreshed token resolves to a different user (ConnectionUserMismatch).");
+                case AckStatus.NotFound:
+                    throw new AzureSignalRException("The target connection was not found, is disconnected, or uses an unsupported negotiate version.");
+                case AckStatus.Timeout:
+                    throw new AzureSignalRException("The operation timed out before the service acknowledged it.");
+                default:
+                    throw new AzureSignalRException("An unexpected error occurred while processing the request.");
+            }
         }
 
         public override IEnumerable<ServiceEndpoint> GetServiceEndpoints() => _endpointManager.GetEndpoints(_hubName);
--- a/src/Microsoft.Azure.SignalR.Management/ServiceHubContextImpl.cs
+++ b/src/Microsoft.Azure.SignalR.Management/ServiceHubContextImpl.cs
@@ -10,6 +10,7 @@ using System.Threading.Tasks;
 
 using Microsoft.AspNetCore.Http.Connections;
 using Microsoft.AspNetCore.SignalR;
+using Microsoft.Azure.SignalR.Common;
 using Microsoft.Azure.SignalR.Protocol;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
--- a/src/Microsoft.Azure.SignalR.Management/ServiceHubContextImpl.cs
+++ b/src/Microsoft.Azure.SignalR.Management/ServiceHubContextImpl.cs
@@ -70,7 +70,7 @@ namespace Microsoft.Azure.SignalR.Management
             var result = await _lifetimeManager.RefreshAuthAsync(connectionToken, expireTime, claims, cancellationToken);
             ThrowOnNonSuccess(result.Status);
 
-            if (result.Claims == null)
+            if (result.Claims == null || result.Claims.Count == 0)
             {
                 throw new AzureSignalRException("The service did not return the post-refresh claim set required to mint the refreshed access token.");
             }
--- a/api/Microsoft.Azure.SignalR.Management.net8.0.cs
+++ b/api/Microsoft.Azure.SignalR.Management.net8.0.cs
@@ -8,6 +8,11 @@ namespace Microsoft.Azure.SignalR.Management
         public abstract System.Threading.Tasks.Task<bool> GroupExistsAsync(string groupName, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken));
         public abstract System.Threading.Tasks.Task<bool> UserExistsAsync(string userId, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken));
     }
+    public sealed partial class ConnectionClaimsResult
+    {
+        internal ConnectionClaimsResult() { }
+        public System.Collections.Generic.IReadOnlyList<System.Security.Claims.Claim> Claims { [System.Runtime.CompilerServices.CompilerGeneratedAttribute] get { throw null; } }
+    }
     public abstract partial class GroupManager : Microsoft.AspNetCore.SignalR.IGroupManager
     {
         protected GroupManager() { }
@@ -69,6 +74,12 @@ namespace Microsoft.Azure.SignalR.Management
         public NewtonsoftServiceHubProtocolOptions() { }
         public Newtonsoft.Json.JsonSerializerSettings PayloadSerializerSettings { [System.Runtime.CompilerServices.CompilerGeneratedAttribute] get { throw null; } [System.Runtime.CompilerServices.CompilerGeneratedAttribute] set { } }
     }
+    public sealed partial class RefreshAuthResult
+    {
+        internal RefreshAuthResult() { }
+        public string AccessToken { [System.Runtime.CompilerServices.CompilerGeneratedAttribute] get { throw null; } }
+        public int TokenLifetimeSeconds { [System.Runtime.CompilerServices.CompilerGeneratedAttribute] get { throw null; } }
+    }
     public abstract partial class ServiceHubContext : Microsoft.AspNetCore.SignalR.IHubContext<Microsoft.AspNetCore.SignalR.Hub>, Microsoft.Azure.SignalR.Management.IServiceHubContext, System.IDisposable
     {
         protected ServiceHubContext() { }
@@ -81,9 +92,11 @@ namespace Microsoft.Azure.SignalR.Management
         public virtual Microsoft.Azure.SignalR.Management.UserGroupManager UserGroups { get { throw null; } }
         public virtual void Dispose() { }
         public virtual System.Threading.Tasks.Task DisposeAsync() { throw null; }
+        public virtual System.Threading.Tasks.Task<Microsoft.Azure.SignalR.Management.ConnectionClaimsResult> GetConnectionClaimsAsync(string connectionToken, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
         [System.ComponentModel.EditorBrowsableAttribute(System.ComponentModel.EditorBrowsableState.Never)]
         public virtual System.Collections.Generic.IEnumerable<Microsoft.Azure.SignalR.ServiceEndpoint> GetServiceEndpoints() { throw null; }
         public virtual System.Threading.Tasks.ValueTask<Microsoft.AspNetCore.Http.Connections.NegotiationResponse> NegotiateAsync(Microsoft.Azure.SignalR.Management.NegotiationOptions negotiationOptions = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
+        public virtual System.Threading.Tasks.Task<Microsoft.Azure.SignalR.Management.RefreshAuthResult> RefreshAuthAsync(string connectionToken, System.DateTimeOffset expireTime, System.Collections.Generic.IEnumerable<System.Security.Claims.Claim> claims = null, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken)) { throw null; }
         [System.ComponentModel.EditorBrowsableAttribute(System.ComponentModel.EditorBrowsableState.Never)]
         public virtual Microsoft.Azure.SignalR.Management.ServiceHubContext WithEndpoints(System.Collections.Generic.IEnumerable<Microsoft.Azure.SignalR.ServiceEndpoint> endpoints) { throw null; }
     }
--- a/api/Microsoft.Azure.SignalR.Management.netstandard2.0.cs
+++ b/api/Microsoft.Azure.SignalR.Management.netstandard2.0.cs
@@ -8,6 +8,11 @@ namespace Microsoft.Azure.SignalR.Management
         public abstract System.Threading.Tasks.Task<bool> GroupExistsAsync(string groupName, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken));
         public abstract System.Threading.Tasks.Task<bool> UserExistsAsync(string userId, System.Threading.CancellationToken cancellationToken = default(System.Threading.CancellationToken));
     }
+    public sealed partial class ConnectionClaimsResult
+    {
+        internal ConnectionClaimsResult() { }
+        public System.Collections.Generic.IReadOnlyList<System.Security.Claims.Claim> Claims { [System.Runtime.CompilerServices.CompilerGeneratedAttribute] get { throw null; } }
+    }
     public abstract partial class GroupManager : Microsoft.AspNetCore.SignalR.IGroupManager
     {
         protected GroupManager() { }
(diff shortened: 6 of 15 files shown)


src/Microsoft.Azure.SignalR.Management/RestApiProvider.cs
```diff
@@ -110,12 +112,22 @@ public RestApiEndpoint GetClientInvocationEndpoint(string appName, string hubNam
...
 
-    private RestApiEndpoint GenerateRestApiEndpoint(string appName, string hubName, string pathAfterHub, IDictionary<string, StringValues> queries = null)
+    public RestApiEndpoint GetRefreshConnectionAuthEndpoint(string appName, string hubName)
+    {
+        return GenerateRestApiEndpoint(appName, hubName, "/connections/:refreshAuth", apiVersion: RefreshApiVersion);
+    }
+
+    public RestApiEndpoint GetConnectionClaimsEndpoint(string appName, string hubName)
+    {
+        return GenerateRestApiEndpoint(appName, hubName, "/connections/:getClaims", apiVersion: RefreshApiVersion);
```
vwxyzh: what about "query"?
see: https://www.rfc-editor.org/rfc/rfc10008.html#section-2

MoChilia: Fabulous idea! Agree

src/Microsoft.Azure.SignalR.Management/RestApiProvider.cs
```diff
@@ -14,6 +14,8 @@ internal class RestApiProvider
     internal const string Version = "2024-12-01";
     public const string HealthApiPath = $"api/health?api-version={Version}";
 
+    internal const string RefreshApiVersion = "2026-07-01";
```
Y-Sindo: Why does the refresh API have a separate version? If you add a new controller for a new version, it should extend the previous controller and inherit all the existing APIs.

src/Microsoft.Azure.SignalR.Management/HubContext/NegotiationResult.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        /// <summary>
+        /// The access token for a client to connect to Azure SignalR Service.
+        /// </summary>
+        public string AccessToken { get; }
+
+        /// <summary>
+        /// Seconds until the client should refresh its authentication.
+        /// </summary>
+        public int TokenLifetimeSeconds { get; }
```
Y-Sindo, line 31: SignalR finally uses `TimeSpan TokenLifetime` https://github.com/dotnet/aspnetcore/blob/main/src/SignalR/common/Http.Connections.Common/src/NegotiationResponse.cs#L62

src/Microsoft.Azure.SignalR.Management/Negotiation/NegotiateProcessor.cs
```diff
@@ -68,5 +68,21 @@ public async Task<NegotiationResponse> NegotiateAsync(string hubName, Negotiatio
...
+        public async Task<NegotiationResult> NegotiateWithTokenLifetimeAsync(string hubName, NegotiationOptions negotiationOptions, CancellationToken cancellationToken = default)
+        {
+            negotiationOptions ??= NegotiationOptions.Default;
+            var response = await NegotiateAsync(hubName, negotiationOptions, cancellationToken);
+            return new NegotiationResult(response.Url, response.AccessToken, ComputeTokenLifetimeSeconds(negotiationOptions));
+        }
+
+        private static int ComputeTokenLifetimeSeconds(NegotiationOptions negotiationOptions)
+        {
+            var seconds = Math.Max(0, negotiationOptions.TokenLifetime.TotalSeconds);
```
Y-Sindo: If `negotiationOptions.TokenLifetime.TotalSeconds)` < 0 , then it's an invalid input and the generated token won't work. How about checking it at the beginning?

src/Microsoft.Azure.SignalR.Management/ServiceHubContext.cs
```diff
@@ -38,6 +39,28 @@ public abstract class ServiceHubContext : IServiceHubContext, IDisposable
         /// <returns>A negotiation response object that contains an endpoint url and an access token for the client to connect to the Azure SignalR instance. </returns>
         public virtual ValueTask<NegotiationResponse> NegotiateAsync(NegotiationOptions negotiationOptions = null, CancellationToken cancellationToken = default) => throw new NotImplementedException();
```
Y-Sindo, line 40: Deprecate this method so that users don't need to choose between two `NegotiateAsync` methods?

src/Microsoft.Azure.SignalR.Management/RestHubLifetimeManager.cs
```diff
@@ -294,6 +299,145 @@ await _restClient.SendWithRetryAsync(api, HttpMethod.Head, handleExpectedRespons
...
+        public string ConnectionToken { get; set; } = string.Empty;
+    }
+
+    private sealed class ClaimsResponseBody
+    {
+        [JsonPropertyName("claims")]
+        public ClaimDto[]? Claims { get; set; }
+    }
+
+    private sealed class ClaimDto
```
Y-Sindo: Could we move these REST models to a separate directory like `RestModels`?

src/Microsoft.Azure.SignalR.Management/ServiceHubContext`T.cs
```diff
@@ -44,8 +44,13 @@ public abstract class ServiceHubContext<T> : IHubContext<Hub<T>, T>, IDisposable
         /// <returns>A negotiation response object that contains an endpoint url and an access token for the client to connect to the Azure SignalR instance. </returns>
         public abstract ValueTask<NegotiationResponse> NegotiateAsync(NegotiationOptions negotiationOptions = null, CancellationToken cancellationToken = default);
```
Y-Sindo, line 48: Also deprecate this one

src/Microsoft.Azure.SignalR.Management/ServiceHubContext`T.cs
```diff
@@ -44,8 +44,13 @@ public abstract class ServiceHubContext<T> : IHubContext<Hub<T>, T>, IDisposable
...
 
+        /// <summary>
+        /// Performs a negotiation that also returns the client access token's lifetime seconds.
+        /// </summary>
+        public virtual Task<NegotiationResult> NegotiateWithTokenLifetimeAsync(NegotiationOptions negotiationOptions = null, CancellationToken cancellationToken = default) => throw new NotImplementedException();
+
         public abstract ValueTask DisposeAsync();
```
Y-Sindo, line 71: Also add another two new methods？

src/Microsoft.Azure.SignalR.Management/ServiceHubContextImpl.cs
```diff
@@ -49,13 +52,81 @@ public ServiceHubContextImpl(string hubName, IHubContext<Hub> hubContext, IServi
...
+            var result = await _lifetimeManager.RefreshConnectionAuthenticationAsync(connectionToken, expireTime, claims, cancellationToken);
+            ThrowOnNonSuccess(result.Status);
+
+            if (result.Claims == null || result.Claims.Count == 0)
+            {
+                throw new AzureSignalRException("The service did not return the post-refresh claim set required to mint the refreshed access token.");
+            }
+
+            var accessTokenLifetime = Constants.Periods.DefaultAccessTokenLifetime;
+            ServiceEndpoint owningEndpoint = result.OwningEndpoint;
```
Y-Sindo: In multi-endpoint scenario, we need to make sure the generated token is for the same endpoint, right? There is no reconnection during the refresh.

Currently transient mode doesn't support multi-endpoint. Persistent mode supports it.

src/Microsoft.Azure.SignalR.Management/ServiceHubContextImpl.cs
```diff
@@ -49,13 +52,81 @@ public ServiceHubContextImpl(string hubName, IHubContext<Hub> hubContext, IServi
...
+        }
+
+        public override async Task<RefreshConnectionAuthenticationResult> RefreshConnectionAuthenticationAsync(string connectionToken, DateTimeOffset expireTime, IEnumerable<Claim> claims = null, CancellationToken cancellationToken = default)
+        {
+            if (string.IsNullOrEmpty(connectionToken))
+            {
+                throw new ArgumentException("Argument cannot be null or empty.", nameof(connectionToken));
+            }
+
+            var result = await _lifetimeManager.RefreshConnectionAuthenticationAsync(connectionToken, expireTime, claims, cancellationToken);
```
Y-Sindo: Just realize `RefreshConnectionAuthResult` and `RefreshConnectionAuthenticationResult` are different types. It seems confusing... Could we give them better name?

Y-Sindo: For example, this class can be "RefreshedToken"?

src/Microsoft.Azure.SignalR.Management/ServiceHubContextImpl.cs
```diff
@@ -49,13 +52,81 @@ public ServiceHubContextImpl(string hubName, IHubContext<Hub> hubContext, IServi
...
+                owningEndpoint = endpoints[0];
+            }
+            var provider = _endpointManager.GetEndpointProvider(owningEndpoint);
+            var accessToken = await provider.GenerateClientAccessTokenAsync(_hubName, result.Claims, accessTokenLifetime);
+            var remainingSeconds = (expireTime - DateTimeOffset.UtcNow).TotalSeconds;
+            var tokenLifetimeSeconds = (int)Math.Max(0, Math.Min(accessTokenLifetime.TotalSeconds, remainingSeconds));
+            return new RefreshConnectionAuthenticationResult(accessToken, tokenLifetimeSeconds);
+        }
+
+        public override async Task<ConnectionClaimsResult> GetConnectionClaimsAsync(string connectionToken, CancellationToken cancellationToken = default)
```
Y-Sindo: Could we just name it "ConnectionClaims" Or `GetConnectionClaimsResult`?

src/Microsoft.Azure.SignalR.Management/ServiceHubContext.cs
Y-Sindo: Apply `[EditorBrowsable(EditorBrowsableState.Never)]`

src/Microsoft.Azure.SignalR.Management/RestApiProvider.cs
MoChilia: No need for a separate version, just because the new version hasn’t been released yet, use one here in case of regression, I will remove later

src/Microsoft.Azure.SignalR.Management/HubContext/NegotiationResult.cs
MoChilia: It's TokenLifetime in seconds, the new signalr client parses it as an Int32 https://github.com/dotnet/aspnetcore/blob/63084a990e68fbb0cea7567e33af97e07484360f/src/SignalR/clients/csharp/Http.Connections.Client/src/HttpConnection.cs#L797-L804

src/Microsoft.Azure.SignalR.Management/Negotiation/NegotiateProcessor.cs
MoChilia: Good point. I'll add that validation directly in Negotiation

src/Microsoft.Azure.SignalR.Management/ServiceHubContextImpl.cs
MoChilia: Absolutely right

src/Microsoft.Azure.SignalR.Management/ServiceHubContextImpl.cs
MoChilia: Good point. I’m renaming the internal API that sends `RefreshAuthMessage` to `RefreshAuthAsync`, with `RefreshAuthResult` as its result type. The public Management SDK API remains `RefreshConnectionAuthenticationAsync` with `RefreshConnectionAuthenticationResult`, which returns not only the service response but also a newly minted service access token. I think this naming distinguishes the internal service-protocol operation from the public SDK operation while keeping each method/result pair consistent.

src/Microsoft.Azure.SignalR.Management/Negotiation/NegotiateProcessor.cs
Y-Sindo: Lint: As the total lifetime is ensured to be positive, this line is unnecessary.

src/Microsoft.Azure.SignalR.Management/ConnectionAuthenticationHelper.cs
```diff
@@ -0,0 +1,92 @@
...
+            if (endpoints.Count == 0)
+            {
+                throw new AzureSignalRException("No service endpoint is available to mint the refreshed access token.");
+            }
+            owningEndpoint = endpoints[0];
+        }
+        var provider = endpointManager.GetEndpointProvider(owningEndpoint);
+        var accessToken = await provider.GenerateClientAccessTokenAsync(hubName, result.Claims, accessTokenLifetime);
+        var remainingSeconds = (expireTime - DateTimeOffset.UtcNow).TotalSeconds;
+        var tokenLifetimeSeconds = (int)Math.Max(0, Math.Min(accessTokenLifetime.TotalSeconds, remainingSeconds));
```
Y-Sindo: This can also be simplified as `accessTokenLifetime` is positive.

src/Microsoft.Azure.SignalR.Management/RestApiProvider.cs
```diff
@@ -114,12 +112,12 @@ public RestApiEndpoint GetClientInvocationEndpoint(string appName, string hubNam
...
+        return GenerateRestApiEndpoint(appName, hubName, "/connections/:refreshAuth");
     }
 
     public RestApiEndpoint GetConnectionClaimsEndpoint(string appName, string hubName)
     {
-        return GenerateRestApiEndpoint(appName, hubName, "/connections/claims", apiVersion: RefreshApiVersion);
+        return GenerateRestApiEndpoint(appName, hubName, "/connections/claims");
     }
 
     private RestApiEndpoint GenerateRestApiEndpoint(string appName, string hubName, string pathAfterHub, IDictionary<string, StringValues> queries = null, string apiVersion = Version)
```
Y-Sindo: Remove this `apiVersion` parameter ?

test/Microsoft.Azure.SignalR.E2ETests/Management/RefreshAuthE2ETests.cs
```diff
@@ -0,0 +1,286 @@
...
+using Microsoft.AspNetCore.SignalR;
+using Microsoft.AspNetCore.SignalR.Client;
+using Microsoft.AspNetCore.Testing.xunit;
+using Microsoft.Azure.SignalR.Common;
+using Microsoft.Azure.SignalR.Tests;
+using Microsoft.Azure.SignalR.Tests.Common;
+
+using Xunit;
+
+#pragma warning disable CS0618 // IServiceManager exposes direct client endpoint/token generation used by these E2E tests.
```
Y-Sindo: Why not using `HubContext.NegotiateAsync()`? It's better to test directly again the API we recommend users to use.

src/Microsoft.Azure.SignalR.Management/ConnectionAuthenticationHelper.cs
MoChilia: `remainingSeconds` can still be negative

Y-Sindo: If remainingSeconds is negative, the entire expression evaluates to 0. If tokenLifetime ends up being 0, doesn't that indicate an invalid scenario? Under what circumstances would this happen? Also, how is expireTime determined? Is it an arbitrary value provided by the user?

MoChilia: This calculation happens after the runtime has successfully applied the refresh and updated the authentication deadline. Throwing at this point could mislead the customer into believing that no state was changed, even though the runtime has already mutated the connection state.

Returning 0 does not mean that the service access token has a zero lifetime. The token is still minted with the configured `AccessTokenLifetime`. `TokenLifetimeSeconds` is only a client-side refresh scheduling hint. A value of 0 means the authentication deadline has already elapsed, so the client should not schedule another future refresh.

This is a rare race condition that can occur when the new authentication deadline is very short. Then the runtime may apply the refresh successfully, but by the time the response is prepared, `remainingSeconds` has reached zero or become negative. Clamping the value to zero is a safer representation of that race than throwing after the state change has already succeeded.

MoChilia: `expireTime` is determined by the application that authenticates the refresh request. It should be the expiration time of the newly validated application credential.

src/Microsoft.Azure.SignalR.Management/ConnectionAuthenticationHelper.cs
```diff
@@ -50,8 +52,8 @@ public static async Task<RefreshConnectionAuthenticationResult> RefreshConnectio
             ?? throw new AzureSignalRException("Unexpected refresh result: The owning endpoint was not provided.");
         var provider = endpointManager.GetEndpointProvider(owningEndpoint);
         var accessToken = await provider.GenerateClientAccessTokenAsync(hubName, result.Claims, accessTokenLifetime);
```
Y-Sindo: Respect the expiretime?

src/Microsoft.Azure.SignalR.Management/ConnectionAuthenticationHelper.cs
```diff
@@ -32,12 +32,14 @@ public static async Task<RefreshConnectionAuthenticationResult> RefreshConnectio
             throw new ArgumentException("Argument cannot be null or empty.", nameof(connectionToken));
         }
 
-        if (expireTime <= DateTimeOffset.UtcNow)
+        var utcExpireTime = expireTime?.ToUniversalTime() ?? DateTimeOffset.MaxValue;
```
Y-Sindo, line 37: So if we see `DateTimeOffset.MaxValue`, we will treat it as null? Why don't we just pass a null into lifetime manager?

src/Microsoft.Azure.SignalR.Management/HubContext/RefreshConnectionAuthenticationOptions.cs
```diff
@@ -0,0 +1,31 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Security.Claims;
+
+using Microsoft.Azure.SignalR.Common;
+
+namespace Microsoft.Azure.SignalR.Management;
+
+/// <summary>
```
Y-Sindo, line 14: Enable nullable for new files

src/Microsoft.Azure.SignalR.Management/ServiceHubContext`T.cs
```diff
@@ -44,8 +47,30 @@ public abstract class ServiceHubContext<T> : IHubContext<Hub<T>, T>, IDisposable
         /// <returns>A negotiation response object that contains an endpoint url and an access token for the client to connect to the Azure SignalR instance. </returns>
         public abstract ValueTask<NegotiationResponse> NegotiateAsync(NegotiationOptions negotiationOptions = null, CancellationToken cancellationToken = default);
 
+        /// <summary>
+        /// Performs a negotiation that also returns the client access token's lifetime seconds. Requires <see cref="NegotiationOptions.EnableAuthenticationRefresh"/>.
+        /// </summary>
+        /// <exception cref="ArgumentException"><see cref="NegotiationOptions.EnableAuthenticationRefresh"/> is not enabled.</exception>
```
Y-Sindo: Enabling auth refresh shouldn't be necessary. The token always has a lifetime. And we want to replace the original `NegotiationAsync` method with this method, right?

src/Microsoft.Azure.SignalR.Management/ConnectionAuthenticationHelper.cs
```diff
@@ -0,0 +1,129 @@
...
+        string connectionToken,
+        RefreshConnectionAuthenticationOptions options,
+        CancellationToken cancellationToken)
+    {
+        if (string.IsNullOrEmpty(connectionToken))
+        {
+            throw new ArgumentException("Argument cannot be null or empty.", nameof(connectionToken));
+        }
+
+        options ??= new RefreshConnectionAuthenticationOptions();
```
Y-Sindo: Use a static readonly  instance instead. No need to create a new options each time.

src/Microsoft.Azure.SignalR.Common/Utilities/ClaimsUtility.cs
```diff
@@ -162,6 +172,29 @@ public static IEnumerable<Claim> BuildJwtClaims(
...
+            {
+                if (TokenGeneratedClaimTypes.Contains(claim.Type))
+                {
+                    continue;
+                }
+
+                var canonicalType = ClaimTypeMapping.OutboundClaimTypeMap.TryGetValue(claim.Type, out var outboundType)
+                    ? outboundType
+                    : claim.Type;
+                var key = string.Concat(canonicalType, "\0", claim.Value, "\0", claim.ValueType);
```
Y-Sindo, line 189: Why do we deduplicate claims?

src/Microsoft.Azure.SignalR.Management/ServiceHubContext.cs
```diff
@@ -38,6 +39,28 @@ public abstract class ServiceHubContext : IServiceHubContext, IDisposable
...
 
+        /// <summary>
+        /// Performs a negotiation that also returns the client access token's lifetime seconds. Requires <see cref="NegotiationOptions.EnableAuthenticationRefresh"/>.
+        /// </summary>
+        /// <exception cref="ArgumentException"><see cref="NegotiationOptions.EnableAuthenticationRefresh"/> is not enabled.</exception>
+        public virtual Task<NegotiationResult> NegotiateWithTokenLifetimeAsync(NegotiationOptions negotiationOptions = null, CancellationToken cancellationToken = default) => throw new NotImplementedException();
```
Y-Sindo, line 45: Add some comments regarding how token lifetime is computed?

src/Microsoft.Azure.SignalR.Management/ConnectionAuthenticationHelper.cs
MoChilia: It's now calculated with `accessTokenLifetime = min(configuredTokenLifetime, AuthenticationExpiresOn-now)`
https://github.com/Azure/azure-signalr/blob/68689147895721afc5b1eb5712af9c2257263743/src/Microsoft.Azure.SignalR.Management/ConnectionAuthenticationHelper.cs#L72-L96

src/Microsoft.Azure.SignalR.Management/ConnectionAuthenticationHelper.cs
MoChilia: It is regarded as having no explicit token lifetime. The `ExpireTime` for `RefreshAuthMessage` is meant to be required. No expire time refresh is not in scope previously.
https://github.com/Azure/azure-signalr/blob/2085f747bd387d6437ffe6140ab4793a8d800b8d/src/Microsoft.Azure.SignalR.Protocols/ServiceMessage.cs#L345

src/Microsoft.Azure.SignalR.Management/ServiceHubContext`T.cs
MoChilia: Not sure about this. If a user doesn’t want to enable auth refresh, there doesn’t seem to be any reason to use `NegotiateWithTokenLifetimeAsync`.

src/Microsoft.Azure.SignalR.Common/Utilities/ClaimsUtility.cs
MoChilia: Updated the code. Deduplication should occur only across different source types that map to the same JWT type.
https://github.com/Azure/azure-signalr/pull/2283/changes/68689147895721afc5b1eb5712af9c2257263743#diff-e3d705f2bd91398f2b7a348f796f72c2614c2d5f8c53c36300881bd4b7bc1d64

src/Microsoft.Azure.SignalR.Management/ServiceHubContext.cs
MoChilia: Comments updated in https://github.com/Azure/azure-signalr/pull/2283/changes/828b73913346ff850c8312f11a7d81aa9c20676a

src/Microsoft.Azure.SignalR.Management/ConnectionAuthenticationHelper.cs
Y-Sindo: So only persistent mode requires `ExpireTime`. We should do this trick in WebSocketLifetimeManager, so that we don't need to convert this max value to null somewhere else again.

src/Microsoft.Azure.SignalR.Management/ServiceHubContext`T.cs
Y-Sindo: I think it's for simplicity and reduce users' choice to make. If we make `NegotiationWithTokenLifetimeAsync` a superset of the `NegotiationAsync` method, then we can always recommend users the former one.

src/Microsoft.Azure.SignalR.Management/ConnectionAuthenticationHelper.cs
MoChilia: It’s required in both persistent mode and transient mode. The rest API also requires `ExpireTime` for consistency, following signalr’s definition of the new authentication expiration time: when the authentication ticket has no expiration, it is `DateTimeOffset.MaxValue`.
https://github.com/dotnet/aspnetcore/blob/63084a990e68fbb0cea7567e33af97e07484360f/src/SignalR/common/Http.Connections/src/AuthenticationRefreshContext.cs#L38-L42

Y-Sindo:
image
Now `IAccessKey` is not available in .net 11, don't forget to add net11 as tfm in common package

src/Microsoft.Azure.SignalR.Management/ConnectionAuthenticationHelper.cs
```diff
@@ -0,0 +1,129 @@
...
+        if (accessTokenLifetime == TimeSpan.Zero)
+        {
+            throw new AzureSignalRException("The effective authentication expiration elapsed before the refreshed access token could be generated.");
+        }
+        var owningEndpoint = result.OwningEndpoint
+            ?? throw new AzureSignalRException("Unexpected refresh result: The owning endpoint was not provided.");
+        var provider = endpointManager.GetEndpointProvider(owningEndpoint);
+        var accessToken = await provider.GenerateClientAccessTokenAsync(
+            hubName,
+            ClaimsUtility.PrepareClaimsForToken(result.Claims),
```
Y-Sindo, line 66: If duplication is generated during token generation, why do we deduplicate before token generation?

